PR #2, opened 2017-11-14
Create LICENSE

raphaelning:
Add a top-level license file detectable by GitHub. It is created
using GitHub's standard BSD 3-clause license template.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,29 @@
+BSD 3-Clause License
+
+Copyright (c) 2017, Intel Corporation
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+* Redistributions of source code must retain the above copyright notice, this
+  list of conditions and the following disclaimer.
+
+* Redistributions in binary form must reproduce the above copyright notice,
+  this list of conditions and the following disclaimer in the documentation
+  and/or other materials provided with the distribution.
+
+* Neither the name of the copyright holder nor the names of its
+  contributors may be used to endorse or promote products derived from
+  this software without specific prior written permission.
+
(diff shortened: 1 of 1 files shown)


raphaelning:
It turns out one of the source files (core/page_walker.c) has a different license (Apache 2.0) due to historical reasons, so we have to find another LICENSE file template. My colleague is working on that.

stweil:
Is it possible to add the BSD 3-clause license to `core/page_walker.c` (and optionally remove the Apache 2.0 license)? If all license holders (Intel Corporation, others?) agree, that would be a simple solution.

raphaelning:
Makes sense. I believe the only license holder is Intel, since I was told the code was taken from another Intel software project. I'll check with my colleague to see if we can just replace the license for this file.

Speaking of licenses, include/hax_interface.h and include/vcpu_state.h are basically the same code as target/i386/hax-interface.h in the QEMU tree, but the QEMU file has a GPLv2 license. I'm not sure if this inconsistency is an issue. Should we add the BSD license to the QEMU file, and perhaps remove the GPLv2 license?

stweil:
Adding the BSD license looks logical for me.

I suggest to send a git patch for `target/i386/hax-interface.h` which adds the BSD license to the _qemu-devel_ mailing list. Then we can discuss there whether it is better to have two licenses for that file or whether it is fine to have only the BSD license.

mattl:
Glad to see this is happening, I was going to file an issue about this :)

raphaelning:
Thanks, I'll make the patch and send it to _qemu-devel_.

As for `core/page_walker.c`, the advice I got from my colleagues is to relicense it to 3-clause BSD. There is an internal approval process we need to go through, but we think it can be done by tomorrow. Then I'll come back and merge this change.

raphaelning:
`core/page_walker.c` has been relicensed to 3-clause BSD. Merging this patch now.

PR #24, opened 2018-01-16
Add NX Support.

raphaelning:
Signed-off-by: Haitao Shan <hshan@google.com>


raphaelning:
Test: Boot a Linux guest, `cat /proc/cpuinfo` from within the guest shows _nx_ in feature list, `dmesg` contains either

`NX (Execute Disable) protection: active` (if guest is x86_64), or

`Notice: NX (Execute Disable) protection cannot be enabled: non-PAE kernel!` (if guest is x86).

PR #37, opened 2018-03-26
HAXM RAM PROTECTION Implementation

Taogle2018:
Protection of ram access using EPT table.


Taogle2018:
Set up a new pull request for only EPT part. I would think the other part needs more discussion and possible re-implementation, as well as tests.

raphaelning:
@AlexAltea That's right, thanks for the reminder!

Issue #12, opened 2017-11-27
vcpu_panic: Unexpected MMIO instruction (opcode 0xF7)


LukeUsher:
The platform I am trying to virtualize with HAXM uses the TEST opcode to check the value of an MMIO register. 

HAXM's fast mmio implementation doesn't handle this and the vcpu implementation panics.

![image](https://user-images.githubusercontent.com/740003/33246272-819d5fc4-d30a-11e7-92f6-eb37755270d4.png)

raphaelning:
Thanks for the report. Indeed HAXM's x86 instruction emulation logic is incomplete. I think it needs some refactoring, which will make it easier for us to add support for more instructions.

AlexAltea:
@LukeUsher PR #42 should have fixed this issue. Can confirm it?

raphaelning:
Yes, thanks for the reminder! The only `TEST` opcodes that may access memory are `0xF6`, `0xF7`, `0x84` and `0x85`, all of which should be supported by now.

PR #42, opened 2018-04-06
Refactored x86 instruction emulator

AlexAltea:
**NOTE: This is not yet ready to be merged. It's work-in-progress, but given how how deep these changes will be, I'd like early feedback from other developers.**

*Last update: 2018-04-20.*

---

I started rewriting the instruction emulator, as the current one is oversimplified to support a small subset of common instructions, but can't easily be extended to support others. This will fix issue #12 as well as many issues emulation-related in my guests.

Please let me know your thoughts/questions on the matter, and report any design issue, so that I can modify it in time. I'll keep updating this post as I progress with this task.

## Design
The new emulator I'm working on is heavily inspired by KVM's x86 emulator [[1]](https://github.com/torvalds/linux/blob/master/arch/x86/kvm/emulate.c), following these design choices (description is oversimplified but gives should give a rough idea):

- Decoupled instruction *decoder* and instruction *emulator*. The *decoder* parses the instruction, storing the corresponding emulator handler and properties of destination/source operands in a context-like structure. The *emulator* uses this context to dispatch `HAX_EXIT_FAST_MMIO` to handle destination/source memory-operands and will execute the handler.

- Emulator handlers will come in two flavors:
    1. Most instructions, e.g. ALU-related, will be functions using inlined assembly (see remark *1*) with just the emulated instruction where all operands have been replaced by fixed register-operands. These instructions are followed of additional code to save EFLAGS/RFLAGS.
    2. Some instructions, e.g. branching-related, will be handled by custom functions specifically written for that particular instruction.

- As expected, this emulator will be decoupled from `vcpu.c` into its own file `emulate.c`.

A high-level overview+comparison of the proposed changes can be found below (apologies for the bad handwriting but that was faster to sketch everything):

- HAXM (old): haxm-old
- HAXM (new): haxm-new

## Remarks

1. MSVC doesn't support inlined assembly on x64, and (correct me if I'm mistaken) Xcode/LLVM doesn't support MASM syntax that we could use as alternative. This leaves us with two choices:
    1. Duplicate the code for macOS and Windows driver-specific code, and let the linker solve the issue. This will be around 400 lines of code of otherwise unnecessary platform-specific code.
    2. Write the assembly emulator handlers in (say) NASM, and configure Windows/macOS projects to use this external tool. This will add an extra dependency, but the code will be cleaner.

2. As far as I've analyzed the situation, none of these changes will require changes to the API that might break clients like QEMU.


AlexAltea:
> As to where to put the new assembly code, I'd prefer the second solution which avoids code duplication. [...] What do you think?

Completely agreed. I will do that then.

> Your Orbital project looks very interesting. Is HAXM relevant to you for both Windows and macOS?

Thank you. For now, my project is Windows/Linux-only for technical reasons unrelated to HAXM. Nevertheless, I'll make sure my emulator redesign doesn't break macOS (or anything else).

core/include/emulate.h
```diff
@@ -0,0 +1,44 @@
...
+#define HAX_CORE_EMULATE_H_
+
+#include "../include/hax.h"
+#include "emulate_ops.h"
+
+/* Access completed successfully */
+#define EM_CONTINUE        0
+
+struct em_operand_t {
+    unsigned int bytes;
```
raphaelning: nit: Any reason not to use a fixed width integer type like `uint32_t`? I actually don't know what `bytes` refers to here.

core/include/emulate.h
```diff
@@ -0,0 +1,44 @@
...
+    enum {
+        OP_NONE,
+        OP_REG,
+        OP_MEM,
+        OP_IMM,
+    } type;
+    struct segmented_addr_t {
+        HAX_VADDR_T addr;
```
raphaelning: Is `HAX_VADDR_T` defined anywhere? Do you mean `vaddr_t`?

core/include/emulate.h
```diff
@@ -0,0 +1,44 @@
...
+        unsigned segment;
```
raphaelning: nit: Consider using a fixed width type?

core/include/emulate.h
```diff
@@ -0,0 +1,44 @@
...
+
+struct em_context_t {
+    struct em_instruction_t *insn;
+    struct em_operand_t dst;
+    struct em_operand_t src1;
+    struct em_operand_t src2;
+    uint32_t eflags;
+};
+
+int decode_insn(struct em_context_t *ctxt);
```
raphaelning: Is this the function to be called from outside the emulator module, e.g. `vcpu.c`? If so, how to pass in the raw bytes of the instruction?

core/haxlib.vcxproj
```diff
@@ -134,8 +134,7 @@
   </PropertyGroup>
   <!-- Needed by any VcxProj -->
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.props" />
-  <ImportGroup Label="ExtensionSettings">
-  </ImportGroup>
+  <ImportGroup Label="ExtensionSettings" />
```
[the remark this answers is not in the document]
AlexAltea: Visual Studio 2017 (v15.6.4). It was auto-generated, but presumably has no effect, I can revert this part to avoid polluting the history.

core/haxlib.vcxproj
```diff
@@ -180,31 +179,49 @@
     <ClCompile>
       <SDLCheck>true</SDLCheck>
     </ClCompile>
+    <MASM>
+      <UseSafeExceptionHandlers>true</UseSafeExceptionHandlers>
```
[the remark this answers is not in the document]
AlexAltea: I added this manually since it's required under the current settings in order to build *.asm* files. Same setting had also been applied by you guys in `IntelHaxm.vcxproj`.

Anyway, we still need to check whether we can use MASM for the macOS builds and if not, then switch to something else instead (NASM?), which will remove all of these settings in VS.

core/haxlib.vcxproj
```diff
@@ -239,6 +256,8 @@
   <ItemGroup>
     <!-- We only add items (e.g. form ClSourceFiles) that do not already exist (e.g in the ClCompile list), this avoids duplication -->
     <ClCompile Include="@(ClSourceFiles)" Exclude="@(ClCompile)" />
+    <ClCompile Include="emulate.c" />
```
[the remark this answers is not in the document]
AlexAltea: Interesting, I'll fix this.

core/haxlib.vcxproj
```diff
@@ -239,6 +256,8 @@
   <ItemGroup>
     <!-- We only add items (e.g. form ClSourceFiles) that do not already exist (e.g in the ClCompile list), this avoids duplication -->
     <ClCompile Include="@(ClSourceFiles)" Exclude="@(ClCompile)" />
+    <ClCompile Include="emulate.c" />
+    <MASM Include="emulate_ops.asm" />
```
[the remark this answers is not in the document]
AlexAltea: As mentioned above, the *.asm* may need to be rewritten in a language that we can easily integrate both in Xcode and Visual Studio build systems.

core/include/emulate.h
AlexAltea: https://github.com/intel/haxm/blob/428a9dae21742b4d8b3ccbacaeb11e6d37c542b5/include/windows/hax_types_windows.h#L139-L147

But probably `vaddr_t` is more convenient.

core/haxlib.vcxproj
```diff
@@ -269,4 +288,5 @@
   </ItemGroup>
   <!-- /Necessary to pick up proper files from local directory when in the IDE-->
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
+  <ImportGroup Label="ExtensionTargets" />
```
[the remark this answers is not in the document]
AlexAltea: Yes. I will check if this is necessary, or just auto-generated garbage.

AlexAltea:
To test the decoder/emulator, I'm planning to use Microsoft's Native Unit Test Framework [[1]](https://msdn.microsoft.com/en-us/library/hh598953.aspx) since it's integrated nicely into Visual Studio, but I'm open to any suggestions. If you prefer another framework or want the tests to run in macOS too, let me know.

image

The idea is mapping each instruction to a single `TEST_METHOD` where every instruction variant will be tested, e.g. `and reg,mem`, `and mem,reg`, `and reg,imm`, etc. In every variant, we will specify the initial state, the instruction will be assembled (using [Keystone](http://www.keystone-engine.org/)), then decoded+execute by the HAXM emulator and the output state will be verified.

An example of similar unit tests I've written to test emulators can be found here: https://github.com/AlexAltea/nucleus/blob/master/tests/cpu/ppc/ppc_integer.cpp.
Let me know if you agree with the overall design.

raphaelning:
I really like the idea of using a unit testing framework! It would surely benefit not only the new instruction emulator component, but the entire HAXM project :)

I've never used the Microsoft Unit Testing Framework. Here are my questions (some may be dumb):

1. Can we really run all tests in user space? That would be so cool, but I think most existing HAXM headers include `hax_types_windows.h` or `hax_types_mac.h`, which, in turn, includes headers from Windows DDK or Apple I/O Kit, so would there be a problem if you had to include such a header in your unit test `.cpp` (e.g. when you needed some data structures like `vcpu_t`)?
1.  Is it possible to run these unit tests outside Visual Studio, so that they can be automated?
1. According to [this article](https://docs.microsoft.com/en-us/visualstudio/test/writing-unit-tests-for-c-cpp), Visual Studio 2017 also supports other unit testing frameworks like Google Test. I wonder if switching to Google Test now would make it easier to port the unit tests to Mac later (as well as to automate them)?

The idea of using an assembler API like Keystone is also brilliant. But does that mean we'll add a build-time dependency on Keystone (although only for the unit tests)? According to [their GitHub page](https://github.com/keystone-engine/keystone), Keystone is licensed under GPLv2, so I'm not sure if we have a license issue here...

core/haxlib.vcxproj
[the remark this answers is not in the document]
raphaelning: OK don't worry about this. As long as it's auto-generated, it's fine. We created the VC++ project using VS2015, but we can upgrade to VS2017.

core/haxlib.vcxproj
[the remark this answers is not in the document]
raphaelning: I see. So basically this change is a result of putting MASM code in `core/`. I'd still prefer `core/` to host arch-independent code only.

> we still need to check whether we can use MASM for the macOS builds and if not, then switch to something else instead (NASM?)

Right. I doubt there's an assembler that understands MASM syntax and works on macOS, so maybe we should switch to a more portable syntax sooner rather than later? I've confirmed that NASM is supported on macOS: http://brewformulas.org/Nasm

But I also understand that your priority is Windows (and probably x86_64 only). If you want to keep the MASM code for now, could you move it to `windows/`? Eventually, we have to make the new instruction emulator work on Mac before we can release it.

core/include/emulate.h
raphaelning: Right! Sorry I missed the definition (never used it myself).

raphaelning:
I forgot to mention your sketch of the MMIO workflow - that's very nice! One of the biggest changes introduced by the new design is the "re-entry" arrow, so I'd like to understand it better. Basically, what is it about? What about the three small, black arrows?

I like the fact that the new design handles both `vcpu_setup_fastmmio()` and `handle_mmio_post()` in `em_emulate_insn()`, which is more consistent, probably because you now have a better per-vcpu emulation context.

AlexAltea:
> Can we really run all tests in user space?

Yes, the emulator is completely isolated from the rest of the driver thanks to a very simple API [[1]](https://github.com/AlexAltea/haxm/blob/52020a23f3d326e4205e507392803d4e64ce586f/core/vcpu.c#L2708-L2711) to perform simple tasks like reading/writing registers, memory, etc. The only external dependency would be some helper types. With few preprocessor lines we can alternate between `hax_types*` headers for kernel drivers and just a regular `<stdint.h>` typedef's. These are the only lines that we need in `emulate.h` to do this:

```c
#ifdef HAX_TESTS
#include <stdint.h>
#else
#include "../include/hax_types.h"
#endif
```

> Is it possible to run these unit tests outside Visual Studio, so that they can be automated?

Good point. It might require plenty of manual work to do so (I haven't found any article on the matter), so maybe using Google Test might be more convenient. I will look into it. :-)

> But does that mean we'll add a build-time dependency on Keystone (although only for the unit tests)?

Yes, I'd recommend using [vcpkg](https://github.com/Microsoft/vcpkg), so all you need to do is to run `vcpkg install keystone` and you should be ready to build the tests without polluting the solution/project.

> Keystone is licensed under GPLv2, so I'm not sure if we have a license issue here...

Apparently, they explicitly allowed compatibility with BSD licenses:
https://github.com/keystone-engine/keystone/blob/master/EXCEPTIONS-CLIENT

> One of the biggest changes introduced by the new design is the "re-entry" arrow, so I'd like to understand it better. Basically, what is it about? What about the three small, black arrows?

Every instruction might require **multiple** reads/writes, which cannot be handled in a single request to QEMU. For instance, after decoding `neg [addr]`, the emulator needs to: (1) read `addr`, (2) negate the value and (3) write it back to `addr`.

When some the emulator needs to do a special operation, e.g. read/write MMIO, an "*emulator exit*" return code will signal every function in the driver call stack to return immediately (that's the small black arrows), until we are back in userland/QEMU to handle that request.

Then, once we return to the driver, specifically in `vcpu_execute`, we will check if the emulator had pending work, so we can continue with it. This process will repeat as many times as necessary. This is possible with relatively low overhead since the emulator context stores the decoded instruction, and other intermediates.

AlexAltea:
I finished removing all Microsoft-specific code and using instead Google Test and NASM. The current revision builds now on Windows and macOS (of course, it's unusable at the current state since the decoder/emulator isn't finished yet).

raphaelning:
I now understand the arrows in the illustration, thanks!

Glad to know that Keystone is compatible with our license. `vcpkg` seems to be a good way to add Keystone to Windows build, although it means we are using more than one tool for dependency management (GTest seems to be managed by NuGet). That's not a big issue, especially if getting rid of NuGet would break the integration of GTest with VS2017. On Mac, I assume we could just install the Homebrew package for Keystone.

tests/test_emulator.cpp
```diff
@@ -0,0 +1,314 @@
+/*
```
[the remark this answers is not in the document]
AlexAltea: My plans are implementing all instructions currently supported, plus the ones needed for my guest kernel. Addind tests for all of those will take around 1000 LoC, so I won't split anything.

A full test suite for the entire x86 instruction set will certainly be much larger, so at some point it will need to be split. I'll consider it future work.

core/haxlib.vcxproj
```diff
@@ -135,6 +135,7 @@
   <!-- Needed by any VcxProj -->
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.props" />
   <ImportGroup Label="ExtensionSettings">
+    <Import Project="$(VCTargetsPath)\BuildCustomizations\nasm.props" />
```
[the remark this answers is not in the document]
AlexAltea: NASM requires manually installing: https://github.com/ShiftMediaProject/VSNASM on VS2017.

This is somewhat annoying. Although there is a NASM package for NuGet, unfortunately its build customizations that do not work on Visual Studio 2017. I will consider updating that package, so that we don't have to deal with 3 separate sources (vcpkg, nuget, manual installation) for build-time dependencies.

core/haxlib.vcxproj
```diff
@@ -269,4 +274,7 @@
   </ItemGroup>
   <!-- /Necessary to pick up proper files from local directory when in the IDE-->
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
+  <ImportGroup Label="ExtensionTargets">
+    <Import Project="$(VCTargetsPath)\BuildCustomizations\nasm.targets" />
```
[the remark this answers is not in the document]
AlexAltea: (See answer above)

core/haxlib.vcxproj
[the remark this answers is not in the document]
raphaelning: I see. We can live with the current approach, but it would be helpful if you could summarize the additional environment setup steps required to build and test this PR.

core/vcpu.c
```diff
@@ -1504,8 +1464,14 @@ int vcpu_execute(struct vcpu_t *vcpu)
...
     }
-    if (htun->_exit_status == HAX_EXIT_FAST_MMIO) {
-        handle_mmio_post(vcpu, (struct hax_fastmmio *)vcpu->io_buf);
+    // Continue until emulation finished
+    if (!em_ctxt->finished) {
+        switch (em_emulate_insn(em_ctxt)) {
+        case EM_EXIT_MMIO:
+            return HAX_EXIT_FAST_MMIO;
+        case EM_ERROR:
+            return HAX_EXIT;
```
raphaelning: `HAX_EXIT` indicates success. In case of an error, probably you should call `hax_panic_vcpu`, set `err` to some error code, and `goto out`?

core/vcpu.c
```diff
@@ -2012,606 +1946,132 @@ static bool is_mmio_address(struct vcpu_t *vcpu, paddr_t gpa)
...
-    // (2) For 64-bit code, 0x40 ~ 0x4f are size/reg (REX) prefix.
-
-    if (instr[num] >= 0x40 && instr[num] <= 0x4f) {
-        rex_w = (instr[num] & 0x08) != 0;
-        rex_r = (instr[num] & 0x04) != 0;
-        num++;
+    rc = em_emulate_insn(em_ctx);
+    switch (rc) {
+    case EM_EXIT_MMIO:
+        return HAX_EXIT_FAST_MMIO;
```
raphaelning: This is not a valid return value. I've explained it in another comment above.

core/vcpu.c
```diff
@@ -2012,606 +1946,132 @@ static bool is_mmio_address(struct vcpu_t *vcpu, paddr_t gpa)
...
-                override_segment = PF_SEG_OVERRIDE_GS;
-                break;
-            }
-            default: {
-                goto done_legacy_pf;
-            }
-        }
+    rc = em_decode_insn(em_ctx, instr);
+    if (rc != EM_CONTINUE) {
+        return HAX_EXIT;
```
raphaelning: This function is part of the `HAX_VCPU_IOCTL_RUN` handler where `HAX_EXIT` actually indicates success. See my other comment above on error handling.

core/vcpu.c
```diff
@@ -2012,606 +1946,132 @@ static bool is_mmio_address(struct vcpu_t *vcpu, paddr_t gpa)
...
-    if (instr[num] >= 0x40 && instr[num] <= 0x4f) {
-        rex_w = (instr[num] & 0x08) != 0;
-        rex_r = (instr[num] & 0x04) != 0;
-        num++;
+    rc = em_emulate_insn(em_ctx);
+    switch (rc) {
+    case EM_EXIT_MMIO:
+        return HAX_EXIT_FAST_MMIO;
+    case EM_ERROR:
+        return HAX_EXIT;
```
raphaelning: This actually indicates success. See my other comment above on error handling.

core/vcpu.c
```diff
@@ -2012,606 +1946,132 @@ static bool is_mmio_address(struct vcpu_t *vcpu, paddr_t gpa)
...
     }
+    return HAX_RESUME;
```
raphaelning: This actually indicates failure. See my other comment above.

AlexAltea:
> Sorry for the long wait!

No worries, I also went away last week to take a small break. I've addressed the issues you mentioned. Emulator functions, in particular `em_emulate_insn` return `EM_CONTINUE` (= 0) on success, `EM_EXIT*` (> 0) when QEMU handling is required, `EM_ERROR*` (< 0) on errors. Therefore I've mapped:
- `EM_CONTINUE`: Maps to `HAX_EXIT`, except when it's being called inside `vcpu_execute` as it means emulation is finished and we can continue executing the VM.
- `EM_EXIT*` (> 0): Maps to `HAX_EXIT`.
- `EM_ERROR*` (< 0): Maps to `HAX_RESUME`.

> So the last patch of this series has enabled the new instruction emulator, but have you run an integration test?

Not yet, and I'm not sure what would be the best way to make integration tests to test all of HAXM at once. Is it alright if I simply try to launch Debian (as shown in [[1]](https://www.qemu.org/2017/11/22/haxm-usage-windows/)) and Android image?

[1] https://www.qemu.org/2017/11/22/haxm-usage-windows/

core/vcpu.c
```diff
@@ -2012,606 +1911,146 @@ static bool is_mmio_address(struct vcpu_t *vcpu, paddr_t gpa)
...
-                       __func__, vcpu->vcpu_id, va, cs_base, rip, dc->gpa);
+            " gva=0x%llx (CS:IP=0x%llx:0x%llx)\n",
+            __func__, vcpu->vcpu_id, va, cs_base, rip);
         dump_vmcs(vcpu);
         return -1;
     }
 #else  // !CONFIG_HAX_EPT2
     if (!vcpu_read_guest_virtual(vcpu, va, &instr, INSTR_MAX_LEN, INSTR_MAX_LEN,
-                                 0)) {
+        0)) {
```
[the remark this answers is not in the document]
AlexAltea: Sorry, this might have been my IDE. I switch between few computers, and some editors aren't properly configured. I'll fix these (and other) lines.

core/vcpu.c
```diff
@@ -1504,8 +1465,15 @@ int vcpu_execute(struct vcpu_t *vcpu)
...
-    if (htun->_exit_status == HAX_EXIT_FAST_MMIO) {
-        handle_mmio_post(vcpu, (struct hax_fastmmio *)vcpu->io_buf);
+    // Continue until emulation finishes
+    if (!em_ctxt->finished) {
+        rc = em_emulate_insn(em_ctxt);
+        if (rc < 0) {
+            return HAX_RESUME;
+        }
+        if (rc > 0) {
+            return HAX_EXIT;
```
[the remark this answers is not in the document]
AlexAltea: > When rc == EM_EXIT_MMIO, is it safe to assume em_emulate_insn() has already set up htun for MMIO, e.g. htun->_exit_status = HAX_EXIT_FAST_MMIO?

Yes, the `EM_EXIT_MMIO` is returned after `HAX_EXIT_FAST_MMIO` is specified, see e.g. `vcpu_read_memory` and `vcpu_write_memory`. The same can be assumed for any other `EM_EXIT_*` event.

> but adding an assert() here wouldn't hurt.

Sure, you mean something like this, right?
```c
if (rc == EM_EXIT_MMIO)
    assert(htun->_exit_status == HAX_EXIT_FAST_MMIO);
```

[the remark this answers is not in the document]
raphaelning: Yes, exactly. Also I guess `rc == EM_EXIT_MMIO` itself could be asserted, since we expect no other value when `rc > 0`.

core/vcpu.c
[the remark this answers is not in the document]
raphaelning: Thanks! We don't have a script to check coding style or fix issues automatically, so I wouldn't push too hard on this :)

AlexAltea:
Ok. Thanks for your help, I've addressed all your comments.

> How much gap, in terms of functionality, is there between the old decoder/emulator and the new framework?

I've finished porting missing features today, including support for REP* prefixes. Now we should have the same functionality as the old emulator (plus a couple of new instructions). Regarding features, this PR can be considered "finished".

Now I'm testing Debian, but I'm still stuck at the BIOS stage: the OS freezes immediately after starting the QEMU and is followed by a Blue Screen. There might be other issues like this, so I'm going to test and fix them all. I will let you know once I consider this PR to be "ready to merge".

core/emulate.c
```diff
@@ -0,0 +1,807 @@
...
+    default:
+        return EM_ERROR;
+    }
+    ctxt->override_segment = SEG_NONE;
+    ctxt->override_operand_size = 0;
+    ctxt->override_address_size = 0;
+    ctxt->insn = insn;
+    ctxt->lock = 0;
+    ctxt->rep = 0;
+    ctxt->len = 0;
```
raphaelning: nit: Avoid trailing whitespace. I also see this issue elsewhere in the file since my editor is set to highlight trailing whitespace. You may consider doing the same.

core/emulate_ops.asm
```diff
@@ -0,0 +1,242 @@
...
+fastop3wcl shld
+fastop3wcl shrd
+fastop3d bextr
+fastop3d andn
+
+; Instruction dispatching
+global fastop_dispatch
+%ifidn __CONV__, x32_cdecl
+fastop_dispatch:
+    ; TODO: Unimplemented
```
raphaelning: Eventually we need this implemented, since we still support 32-bit Windows.

core/emulate.c
```diff
@@ -0,0 +1,807 @@
...
+    const struct em_opcode_t *opcode = &ctxt->opcode;
+    void(*fast_handler)();
+    void(*soft_handler)(em_context_t*);
+    uint64_t eflags;
+    em_status_t rc;
+    ctxt->finished = false;
+
+restart:
+    // TODO: Permissions, exceptions, etc.
+    if (opcode->flags & (INSN_REP)) {
```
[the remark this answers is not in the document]
AlexAltea: I forgot to add the flag to some instructions, this is fixed now. Thanks for pointing it out.

core/emulate.c
```diff
@@ -0,0 +1,807 @@
...
+        ctxt->modrm.value = insn_fetch_u8(ctxt);
+    }
+
+    /* Apply flags */
+    if (flags & INSN_BYTEOP) {
+        ctxt->operand_size = 1;
+    }
+    if (flags & INSN_GROUP) {
+        opcode_group = &opcode->group[ctxt->modrm.opc];
+        opcode->handler = opcode_group->handler;
```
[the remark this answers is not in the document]
AlexAltea: > Should make sure `opcode_group != NULL`.

Well `opcode_group = &opcode->group[ctxt->modrm.opc];` is always non-NULL because `opcode->group` is always non-NULL (otherwise it would make no sense to use `INSN_GROUP`).

However, you are completely right that opcode_group could point to a zero-initialized entry with a `NULL` handler (e.g. in case of unimplemented/unexpected ModRM opcodes), so I added a check for `opcode_group->handler != NULL`, and if it fails, the emulator decoder will throw `EM_ERROR`.

> IIUC, for unsupported ModR/M opcodes, you just initialize the corresponding group members to `NULL`

The opcode group is initialized as:
```c
static const struct em_opcode_t opcode_groupN[8] = {
    ...
};
```

AFAIK, C99 ensures that any element not explicitly defined (e.g. via `F(em_*, op_*, ...)`) will be zero-initialized.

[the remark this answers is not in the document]
raphaelning: You're right. I missed the `&` operator earlier ;-)

Also guess there's not much benefit of initializing opcode group entries to the `N` macro, because `handler` would still be `NULL`. BTW, currently `N` is almost the same as all-zero, since the `INSN_NOTIMPL` flag is not checked anywhere.

nit: Move declaration of `opcode_group` into this `if` block.

core/emulate.c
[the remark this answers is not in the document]
AlexAltea: Some instructions might require some early checks, so the first `if` is supposed to filter those instructions based on flags (for now it's only `INSN_REP`), but in the future have something like:
```c
    if (opcode->flags & (INSN_REP|INSN_XXX|INSN_YYY|INSN_ZZZ)) {
        if ((opcode->flags & INSN_XXX) && ...) {
            // ...
        }
        if ((opcode->flags & INSN_YYY) && ...) {
            // ...
        }
        // ...
        if ((opcode->flags & INSN_REP) && ctxt->rep) {
            // ...
        }
    }
```

However, I roll it back into a single `if` and leave this for a future PR.

AlexAltea:
Getting there... :-)

image

AlexAltea:
Another, question: do you want me to squash all commits into a single one, or are you fine with keeping the Git history as-is in this pull request? Do you need any special `Signed-off-by: ...` message?

raphaelning:
> Getting there... :-)

Awesome, congrats!!

> While I can successfully build HAXM on a macOS VM, I can't/won't test it.

No problem. You probably don't have the environment to test the 32-bit Windows driver either, so we'll run tests on these hosts.

I was trying to build this PR and ran into issues on both Windows and Mac. I'll describe them in a separate post.

> Is there any alternative to test Android?

Hmm, you could install the stock HAXM first, and then unload it before loading the test-signed driver: `sc stop intelhaxm`. Alternatively, it's possible to run Android Emulator from the command line without installing the Android Studio IDE:

1. Download Android SDK tools from [here](https://developer.android.com/studio/#downloads) (under **Command line tools only**)
1. Unzip the package.
1. Use `sdkmanager` (in `%ANDROID_SDK_ROOT%\tools\bin\`) to install `emulator` and at least one Android system image, e.g. `system-images;android-27;google_apis_playstore;x86`. (I don't think either of these would result in the force installation of stock HAXM.)
1. Use `avdmanager` (in `%ANDROID_SDK_ROOT%\tools\bin\`) to create an AVD (VM) configuration.
1. `cd emulator\`
1. List all AVD configurations: `emulator -list-avds`
1. Launch the AVD: `emulator -avd <avd_name> -verbose`

> Do you need any special Signed-off-by: ... message?

Probably not this time, since we haven't defined our sign-off policy yet. But thanks for your consideration!

> do you want me to squash all commits into a single one, or are you fine with keeping the Git history as-is in this pull request?

This is a great question. On one hand, having 20 commits for a big PR like this is better than having one huge commit. On the other hand, there's probably a better way to split the PR, so that each commit is self-contained and does not break the build. I'm thinking about squashing and re-splitting the PR into maybe 3 commits:

a) Upgrade the project to VS2017 format.
b) Implement the new instruction emulator.
c) Enable the new instruction emulator.

But I'm not sure if this idea is feasible, since I haven't given it much thought. E.g. should the emulator-vcpu interface go to _b_ or _c_? (I guess the former.) I'm open to suggestions, and I'm happy to help you with the rebasing and re-splitting. Bottom line: the code itself (and the feature it enables) is more important than the look of the git history. We can make a reasonable effort to improve the git history, but at the end of the day, as long as the code works well, we'll accept it :)

raphaelning:
On Mac I have just one issue, so let me describe that first.

The build was successful, but when I load the signed kext (we have an Apple-approved code-signing certificate) by `sudo kextutil intelhaxm.kext`, I get:

```
(kernel) kxld[com.intel.kext.intelhaxm]:        _em_adc
(kernel) kxld[com.intel.kext.intelhaxm]:        _em_add
(kernel) kxld[com.intel.kext.intelhaxm]:        _em_and
(kernel) kxld[com.intel.kext.intelhaxm]:        _em_cmp
(kernel) kxld[com.intel.kext.intelhaxm]:        _em_dec
(kernel) kxld[com.intel.kext.intelhaxm]:        _em_inc
(kernel) kxld[com.intel.kext.intelhaxm]:        _em_neg
(kernel) kxld[com.intel.kext.intelhaxm]:        _em_not
(kernel) kxld[com.intel.kext.intelhaxm]:        _em_or
(kernel) kxld[com.intel.kext.intelhaxm]:        _em_sbb
(kernel) kxld[com.intel.kext.intelhaxm]:        _em_sub
(kernel) kxld[com.intel.kext.intelhaxm]:        _em_test
(kernel) kxld[com.intel.kext.intelhaxm]:        _em_xor
(kernel) kxld[com.intel.kext.intelhaxm]:        _fastop_dispatch
(kernel) Failed to load executable for kext com.intel.kext.intelhaxm.
(kernel) Kext com.intel.kext.intelhaxm failed to load (0xdc008016).
(kernel) Failed to load kext com.intel.kext.intelhaxm (error 0xdc008016).
```

These symbols don't really exist in the code. Apparently they are derived from the NASM routines (defined in `core/emulate_ops.asm`, e.g. `em_add`) by adding the `_` prefix. It seems only the routines referenced by `core/emulate.c` have this problem. Any idea how to fix this?

You may not be able to reproduce the above error with an unsigned kext. But you can try this:

```
$ kextlibs -undef-symbols /path/to/intelhaxm.kext
For all architectures:
    com.apple.kpi.bsd = 17.5
    com.apple.kpi.iokit = 17.5
    com.apple.kpi.libkern = 17.5
    com.apple.kpi.mach = 17.5

For x86_64:
    17 symbols not found in any library kext:
        _em_add
        _em_xor
        _em_and
        _real_ncpus
        _em_test
        _em_neg
        _em_sbb
        _em_not
        _fastop_dispatch
        _em_dec
        _em_inc
        _em_or
        _em_cmp
        _cpu_number
        _em_sub
        _em_adc
        _mp_rendezvous_no_intrs
```

Note that a few symbols listed in this output (e.g. `_mp_rendezvous_no_intrs`) actually exist in the macOS kernel and can be ignored (they have nothing to do with this PR).

One more experiment:

```
$ nm -a build/intelhaxm.build/Release/intelhaxm.build/DerivedSources/emulate_ops.o
0000000000000180 T em_adc
0000000000000100 T em_add
0000000000000200 T em_and
0000000000000780 T em_andn
0000000000000740 T em_bextr
...
0000000000000280 T em_xor
00000000000007b6 T fastop_dispatch
```
(only original symbol names)

whereas:
```
$ nm -a build/intelhaxm.build/Release/intelhaxm.build/Objects-normal/x86_64/ia32.o
00000000000001ef T ___fls
00000000000001c5 T ___handle_cpuid
00000000000001e7 T ___nmi
000000000000015a T _btr
000000000000016d T _bts
0000000000000180 T _cpu_has_emt64_support
00000000000001ae T _cpu_has_nx_support
0000000000000197 T _cpu_has_vmx_support
0000000000000151 T _fxrstor
0000000000000148 T _fxsave
0000000000000000 T _get_cr0
0000000000000009 T _get_cr2
0000000000000012 T _get_cr4
...
00000000000000a2 T _set_dr7
00000000000000ef T _set_kernel_ds
00000000000000f7 T _set_kernel_es
00000000000000ff T _set_kernel_fs
00000000000000e7 T _set_kernel_gs
```
(only mangled symbol names)

So it seems we need to make NASM add the `_` prefix?

HaxmDriver.sln
```diff
@@ -15,38 +15,42 @@ Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "IntelHaxm", "windows\IntelH
 EndProject
 Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "haxlib", "core\haxlib.vcxproj", "{BC80D1E0-5738-4048-A742-8A20949A6587}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Tests", "Tests", "{4FF4BA9A-1A63-4CBF-A6C0-1D2D83B2146E}"
+EndProject
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "tests", "tests\tests.vcxproj", "{0458723C-5C3F-4509-BDAE-35FAE2A90C19}"
```
raphaelning: There doesn't seem to be an easy way to exclude `tests` from the VS solution build yet. I wonder how we can add a "no tests" option to the build. It's okay if `tests` is built by default, but it's not really a strict requirement for the driver build, so we should make it optional.

Actually I'm having a lot of trouble installing all the dependencies for `tests`... Still haven't got `keystone` installed via `vcpkg` due to a CMake failure :(

raphaelning:
I just tested this PR on Windows 10 x64. It seems there's still some debugging work to do ;-)

1. I also see the Debian boot menu as in your screenshot, but after I make a selection, the guest is stuck and never boots to shell or GUI.
1. The boot ROM test (`qemu-system-x86_64 -accel hax`) fails: the QEMU VGA screen stays black.
1. Android Emulator also boots to a hang (black screen).

darwin/hax_driver/com_intel_hax/intelhaxm.xcodeproj/project.pbxproj
```diff
@@ -73,8 +73,23 @@
...
+/* Begin PBXBuildRule section */
+		FA8F6523208BDF6600C8E91F /* PBXBuildRule */ = {
+			isa = PBXBuildRule;
+			compilerSpec = com.apple.compilers.proxy.script;
+			fileType = sourcecode.nasm;
+			isEditable = 1;
+			outputFiles = (
+				"$(DERIVED_FILE_DIR)/${INPUT_FILE_BASE}.o",
+			);
+			script = "/usr/local/bin/nasm -f macho64 ${INPUT_FILE_PATH} -o ${SCRIPT_OUTPUT_FILE_0}\n";
```
raphaelning: 1. Add `--prefix _` to generate symbol names with a leading underscore, so as to avoid symbol resolution errors when we load `intelhaxm.kext`. ([Reference](https://github.com/jonof/jfbuild/issues/3))
1. nit: Move the required parameter (input file) to the end.

In summary:

```
script = "/usr/local/bin/nasm -f macho64 --prefix _ -o ${SCRIPT_OUTPUT_FILE_0} ${INPUT_FILE_PATH}\n";
```

HaxmDriver.sln
raphaelning: That sounds great! Hopefully NuGet maintainers are more responsive. I like the idea of vcpkg, but right now I don't think it's mature enough for our project to depend on...

AlexAltea:
Sorry about the macOS issue: I'm unfamiliar with that system, so I blindly followed README.md and thought a successful build was enough. Thanks for helping me with this one!

Also, I'll try to come up with the following way of splitting commits. If you agree I'll rewrite the Git history for my fork (after fixing all remaining issues):
a) Update to VS2017
b) Replace old emulator with new one
c) Add tests

I'll tackle the Debian/BootROM/Android issues. Not sure how long it will take, since I'm busy with other ongoing projects at the moment, but I'll comment again at this pull request once its done.

raphaelning:
> Thanks for helping me with this one!

I just dotted the i's and crossed the t's ;-) The portable NASM code you wrote was the key.

> If you agree I'll rewrite the Git history for this PR (after fixing all remaining issues):
> 1. Update to VS2017
> 2. Add new emulator
> 3. Remove old emulator and implement interface to the new one.
> 4. Add tests

I absolutely agree. This is better than my proposal :-)

> I'll tackle the Debian/BootROM/Android issues. Not sure how long it will take, since I'm currently busy with other ongoing projects, but I'll comment again at this PR once its done.

Sure. Meanwhile I'll review the existing code more carefully and see if I can spot any bug.

raphaelning:
The latest couple of patches have fixed the hang (black screen) issue, thanks :-) Some guests (such as `debian_squeeze_i386_desktop.qcow2`) even boot to GUI now - that's massive progress in a day!

At least two issues remain:

1. Some guests still won't boot up. E.g. `debian_wheezy_amd64_standard.qcow2` seems to run into a disk error.
2. Text output to the QEMU VGA display is not fully correct. E.g. the line that is supposed to print the SeaBIOS version (_SeaBIOS (version XXX)_) only prints _S_.

My guess is that there's still a bug in either the decoding or emulation of some 16-bit (real mode) instruction.

AlexAltea:
> Some guests still won't boot up. E.g. `debian_wheezy_amd64_standard.qcow2` seems to run into a disk error.

I'm stuck with this last issue, it's hard to grasp what's wrong and I barely have time these days, so it might take few extra days.

I'm curious about debugging: So far I've just put `hax_warning`'s and logged messages with DebugView. However these `hax_*` logging functions are implemented in Windows relying on `vDbgPrintExWithPrefix` which is limited to a 512 byte buffer, so I frequently lose logs if I generate lots of messages. So far, I've worked around the issue inserting extra delays after every message, but that's suboptimal. Do you guys use any other approach? Probably I should use WinDbg to get what's going on...

I've

core/emulate.c
```diff
@@ -0,0 +1,899 @@
...
+{
+    op->type = OP_REG;
+    op->size = ctxt->operand_size;
+    op->reg.index = ctxt->modrm.reg | (ctxt->rex.r << 3);
+}
+
+static void decode_op_modrm_rm(em_context_t *ctxt,
+                               em_operand_t *op)
+{
+    uint32_t reg_base;
```
raphaelning: nit: Move these 3 declarations into the scope where they are used.

core/emulate.c
```diff
@@ -0,0 +1,899 @@
...
+        op->reg.index = ctxt->modrm.rm | (ctxt->rex.b << 3);
+        return;
+    }
+
+    // Memory operand
+    op->type = OP_MEM;
+    op->size = ctxt->operand_size;
+    op->mem.ea = 0;
+    op->mem.seg = SEG_DS;
+    if (ctxt->override_segment) {
```
raphaelning: nit: `if (ctxt->override_segment != SEG_NONE) {`

core/emulate.c
```diff
@@ -0,0 +1,899 @@
...
+        case 7:
+            op->mem.ea = BX;
+            break;
+        }
+
+        // Displacement
+        if (ctxt->modrm.mod == 1) {
+            op->mem.ea += insn_fetch_s8(ctxt);
+        }
+        if (ctxt->modrm.mod == 2) {
```
raphaelning: nit: `} else if`

core/emulate.c
```diff
@@ -0,0 +1,899 @@
...
+            op->mem.ea += insn_fetch_u16(ctxt);
+        }
+    }
+    if (ctxt->address_size == 4 || ctxt->address_size == 8) {
```
raphaelning: nit: `} else if`

core/emulate.c
```diff
@@ -0,0 +1,899 @@
...
+        /* Intel SDM Vol. 2A:
+         * Table 2-2. 32-Bit Addressing Forms with the ModR/M Byte */
+        if (ctxt->modrm.rm == 4) {
+            /* Intel SDM Vol. 2A:
+             * Table 2-3. 32-Bit Addressing Forms with the SIB Byte */
+            ctxt->sib.value = insn_fetch_u8(ctxt);
+            reg_base  = ctxt->sib.base  | (ctxt->rex.b << 3);
+            reg_index = ctxt->sib.index | (ctxt->rex.x << 3);
+            scale = 1 << ctxt->sib.scale;
+            op->mem.ea += READ_GPR(reg_base, ctxt->address_size);
```
raphaelning: There's a special case: when `mod == 0`, the base register is replaced by a 32-bit displacement. See the notes below Table 2-3 (Intel SDM Vol. 2A).

I've rewritten this part, for your reference:

```
            op->mem.ea = READ_GPR(reg_index, ctxt->address_size) * scale;
            if (ctxt->modrm.mod == 0) {
                op->mem.ea += insn_fetch_s32(ctxt);
            } else {
                uint32_t reg_base = ctxt->sib.base | (ctxt->rex.b << 3);
                op->mem.ea += READ_GPR(reg_base, ctxt->address_size);
            }
```

This seems to fix `debian_wheezy_amd64_standard.qcow2` boot!

core/emulate.c
```diff
@@ -0,0 +1,899 @@
...
+            op->mem.ea += READ_GPR(reg_index, ctxt->address_size) * scale;
+        } else if (ctxt->modrm.mod == 0 && ctxt->modrm.rm == 5) {
+            op->mem.ea += insn_fetch_s32(ctxt);
+        } else {
+            op->mem.ea += ctxt->ops->read_gpr(ctxt->vcpu, ctxt->modrm.rm,
```
raphaelning: nit: Use `READ_GPR` macro instead.

raphaelning:
All the changes made by the latest patch look good, but for some reason there's a regression: the boot ROM test now hangs at `Booting from floppy`.

I was able to spot and fix a bug in the SIB decoding logic. Now the `debian_wheezy_amd64_standard.qcow2` test is passing, but the regression is still there.

I also debug HAXM by inserting `hax_warning()`, etc. and capturing logs with DbgView, but I've never had the buffer overflow problem. There's a "Log to file" option in DbgView - maybe you could try that? Anyway, according to [MSDN](https://docs.microsoft.com/en-us/windows-hardware/drivers/devtest/reading-and-filtering-debugging-messages), the 512-byte buffer is for each message:

> Any single call to DbgPrint, DbgPrintEx, vDbgPrintEx, vDbgPrintExWithPrefix, KdPrint, or KdPrintEx transmits only 512 bytes of information. Any output longer than the 512 bytes is lost.

So you could always split a long message into several print calls. There's another 4KB buffer that may limit the rate at which `DbgPrint` can be called, but it seems adjustable:

> The DbgPrint buffer itself can hold up to 4 KB of data on a free build of Windows, and up to 32 KB of data on a checked build of Windows. On Windows Server 2003 and later versions of Windows, you can use the KDbgCtrl tool to alter the size of the DbgPrint buffer. This tool is part of Debugging Tools for Windows.

raphaelning:
I forgot to mention that, despite all my effort so far to fix the Win32 build, it's still failing :-(

(In EWDK 10 build environment)
```
> msbuild HaxmDriver.sln /p:Configuration="Win7 Debug" /p:Platform="Win32"
...
  haxlib.lib(emulate.obj) : error LNK2001: unresolved external symbol _em_not@0 [<redacted>\windows\IntelHaxm.vcxproj]
  haxlib.lib(emulate.obj) : error LNK2001: unresolved external symbol _em_neg@0 [<redacted>\windows\IntelHaxm.vcxproj]
  ...
  haxlib.lib(emulate.obj) : error LNK2019: unresolved external symbol _fastop_dispatch@20 referenced in function _em_emulate_insn@4 [<redacted>\windows\IntelHaxm.vcxproj]
```

`dumpbin /symbols core\Win7Debug\x86\emulate_ops.obj` clearly shows that the generated symbols have the correct prefix (`_`). But apparently the linker is also looking for a certain suffix, e.g. `@0` or `@20`. Any idea what that is about?

AlexAltea:
This looks like some sort of C++ name mangling? I'll investigate it.
Also, my latest commit (I force-pushed it 5 min ago solves the Floopy thing but Debian is still broken)

core/emulate.c
raphaelning: I got carried away and missed the `base == 5` check, haha. Thanks for fixing it!

core/emulate.c
```diff
@@ -571,36 +567,40 @@ static void decode_op_modrm_rm(em_context_t *ctxt,
...
-            op->mem.ea += READ_GPR(reg_base, ctxt->address_size);
             op->mem.ea += READ_GPR(reg_index, ctxt->address_size) * scale;
+            if ((reg_base & 7) == 5 && ctxt->modrm.mod == 0) {
+                op->mem.ea += insn_fetch_u32(ctxt);
+            } else {
+                op->mem.ea += READ_GPR(reg_base, ctxt->address_size);
+            }
         } else if (ctxt->modrm.mod == 0 && ctxt->modrm.rm == 5) {
-            op->mem.ea += insn_fetch_s32(ctxt);
+            op->mem.ea += insn_fetch_u32(ctxt);
```
raphaelning: Hmm, I'd use signed int for all displacements, even if the effective address is derived from nothing but a displacement. According to Intel SDM Vol. 2A, 2.2.1.3:

> Addressing in 64-bit mode uses existing 32-bit ModR/M and SIB encodings. The ModR/M and SIB displacement
sizes do not change. They remain 8 bits or 32 bits and are **sign-extended** to 64 bits.

HaxmDriver.sln
ras0219-msft: Sorry for dropping that PR on the floor! With our recent support for Linux and OSX, we built up a backlog of PRs that we're actively trying to clear.

Vcpkg can also be used to produce NuGet packages containing all your dependencies automatically: `vcpkg export --nuget keystone gtest`

core/emulate.c
AlexAltea: Ironically, the reason I reverted it back to `u32` was because it fixed the `Booting from Floppy...` issue (as well as the GRUB menu not responding to key presses). But you are completely right, the Intel SDM is clear about this being `s32`.

Probably these are just issues canceling each other out...

AlexAltea:
Good news: `debian_wheezy_amd64_standard.qcow2` boots a little further, although there's still some timeout issues are still printed to screen during initialization, and eventually gets stuck at `Starting MTA: exim4`.

As for the Win32 linking issues, some of them are fixed, but the tests still complain on `em_decode_insn`, `em_emulate_insn`. If I'm understanding this correctly, it seems that `haxlib` uses *stdcall* by default, while `tests` use *cdecl* by default. Changing either setting causes problems. :-/

I'll investigate both issues further.

core/emulate.c
```diff
@@ -0,0 +1,908 @@
...
+    return ((int64)addr << (64 - vaddr_bits)) >> (64 - vaddr_bits);
+}
+
+static em_status_t get_linear_address(struct em_context_t *ctxt,
+                                      struct operand_mem_t *mem,
+                                      uint64_t *la)
+{
+    uint64_t linear;
+
+    linear = ctxt->ops->get_segment_base(ctxt->vcpu, mem->seg) + mem->ea;
```
raphaelning: 1. Truncate `mem->ea` to `ctxt->address_size` before adding `ea` to segment base. This is because `ea` must be non-negative (0 <= `ea` <= segment limit).
2. nit: When in 64-bit mode, don't use the segment base unless `mem->seg` is FS or GS. Usually the guest makes sure the base of CS, DS, etc. is 0 before entering 64-bit mode, but we shouldn't rely on that.

core/emulate.c
```diff
@@ -0,0 +1,908 @@
...
+    if (ctxt->mode == EM_MODE_PROT64) {
+        linear = get_canonical_address(ctxt, linear, 48);
+    } else {
```
raphaelning: Add a case for `EM_MODE_REAL`, where `linear` should be truncated to 20 bits. See Intel SDM Vol. 1, 3.3.5 _32-Bit and 16-Bit Address and Operand Sizes_:

> An address-size override can be used in real-address mode to enable 32-bit addressing. However, the maximum allowable 32-bit linear address is still 000FFFFFH (2^20 - 1).

core/emulate.c
```diff
@@ -0,0 +1,908 @@
...
+                op->mem.ea += insn_fetch_s32(ctxt);
+            } else {
+                op->mem.ea += READ_GPR(reg_base, ctxt->address_size);
+            }
+            if (reg_index != 4) {
+                uint8_t scale = 1 << ctxt->sib.scale;
+                op->mem.ea += READ_GPR(reg_index, ctxt->address_size) * scale;
+            }
+        } else if (ctxt->modrm.mod == 0 && ctxt->modrm.rm == 5) {
+            op->mem.ea += insn_fetch_s32(ctxt);
```
raphaelning: After giving this more thought, I now believe we should use `_s32` when `address_size == 8`, but `_u32` when `address_size == 4`, because `ea` is a segment offset which must be non-negative (0 <= `ea` <= segment limit). Of course, we could also keep `_s32` and truncate `ea` to `address_size`.

raphaelning:
For Android, I've tried both an x86 (32-bit) AVD, `system-images;android-P;google_apis;x86`, and an x86_64 AVD, `system-images;android-P;google_apis;x86_64`. I launch Android Emulator from the command line:

```
> cd %ANDROID_SDK_ROOT%\emulator\
> emulator -avd <avd_name> -verbose -show-kernel -no-snapshot -memory 1536
```

The `-no-snapshot` flag disables AVD snapshot saving/loading (we can re-enable that later to test _quickboot_ after we get cold boot working). The `-memory` flag is necessary if the AVD was created using the command-line tool `avdmanager`, which does not specify a RAM size, and the default RAM size (1024MB) is not enough for recent versions of Android.

The x86 AVD boots up to home screen, but is not really usable, because it doesn't seem to know where my mouse cursor is (the cursor position is crucial for touch screen emulation).

The x86_64 AVD hangs at the Android boot animation, which seems to run at an incredibly slow speed.

core/vcpu.c
```diff
@@ -2067,551 +1957,164 @@ static int vcpu_simple_decode(struct vcpu_t *vcpu, struct decode *dc)
...
-            int imm_offset;
-            uint32 imm_value;
+static void vcpu_write_gpr(void *obj, uint32_t reg_index, uint64_t value, uint32_t size)
+{
+    struct vcpu_t *vcpu = obj;
+    if (reg_index >= 16) {
+        hax_panic_vcpu(vcpu, "vcpu_write_gpr: Invalid register index\n");
+        return;
+    }
+    vcpu->state->_regs[reg_index] = value;
```
raphaelning: Should honor `size`: `memcpy(&vcpu->state->_regs[reg_index], &value, size);`

core/vcpu.c
```diff
@@ -2067,551 +1957,164 @@ static int vcpu_simple_decode(struct vcpu_t *vcpu, struct decode *dc)
...
+        return EM_EXIT_MMIO;
     }
-
-    // Calculate instruction length
-    len = (uint8)(num + 1);  // 1 for opcode
-    if (has_modrm) {
-        len++;
-        if (has_sib) {
-            len++;
+    else {
```
raphaelning: nit: `} else {`

core/vcpu.c
```diff
@@ -2067,551 +1957,164 @@ static int vcpu_simple_decode(struct vcpu_t *vcpu, struct decode *dc)
...
+        return EM_CONTINUE;
     }
-    dc->advance = len;
-    return 0;
 }
 
-static int hax_setup_fastmmio(struct vcpu_t *vcpu, struct hax_tunnel *htun,
-                              struct decode *dec)
+static em_status_t vcpu_read_memory_post(void *obj,
+                                    uint64_t *value, uint32_t size)
```
raphaelning: nit: Align with the first parameter.

core/vcpu.c
```diff
@@ -2067,551 +1957,164 @@ static int vcpu_simple_decode(struct vcpu_t *vcpu, struct decode *dc)
...
-                }
-                break;
-            }
-            default: {
-                hax_panic_vcpu(vcpu, "Invalid address size %u\n",
-                               dec->addr_size);
-                dump_vmcs(vcpu);
-                return -1;
-            }
+    else {
```
raphaelning: nit: `} else {`

core/vcpu.c
```diff
@@ -2067,551 +1957,164 @@ static int vcpu_simple_decode(struct vcpu_t *vcpu, struct decode *dc)
...
-            hax_panic_vcpu(vcpu, "Unexpected MMIO instruction (opcode=0x%x,"
-                           " exit_instr_length=%u, num=%d, gpa=0x%llx,"
-                           " instr[0..5]=0x%x 0x%x 0x%x 0x%x 0x%x 0x%x)\n",
-                           instr[num], vcpu->vmx.exit_instr_length, num,
-                           dc->gpa, instr[0], instr[1], instr[2], instr[3],
-                           instr[4], instr[5]);
-            dump_vmcs(vcpu);
-            return -1;
-        }
+    vcpu_translate(vcpu, ea, 0, &pa, NULL, false);
```
raphaelning: As I said, in most cases we should be able to avoid this expensive call. I'd like to hear what you think of my previous proposal, i.e. to obtain `pa` directly from hardware:

```
pa = vmread(vcpu, VM_EXIT_INFO_GUEST_PHYSICAL_ADDRESS);
```

Of course, we should first identify the opcodes that allow more than one memory operand. E.g., we could define a new flag for `em_opcode_t::flags`, or add a new field to `em_opcode_t` that indicates the maximum number of memory operands allowed.

If we are able to fix all the remaining test failures soon, we can leave this optimization for a future PR. But I think doing it now can help debug the test failures.

core/include/emulate_ops.h
```diff
@@ -0,0 +1,77 @@
...
+
+#define FASTOP_ALIGN  0x10
+#define FASTOP_OFFSET(size) ( \
+    ((size) == 8) ? (3 * FASTOP_ALIGN) : \
+    ((size) == 4) ? (2 * FASTOP_ALIGN) : \
+    ((size) == 2) ? (1 * FASTOP_ALIGN) : \
+                    (0 * FASTOP_ALIGN))
+
+/* Instruction handlers */
+extern void *em_not;
```
raphaelning: nit: So now we are declaring all the handlers as plain pointers instead of function pointers. It's great that it fixes the linker errors, but why? The cost is code readability, although that's not a big issue. I can accept this change, but I just wonder if you have tried other approaches, like decorating the function declaration as you've done with `fastop_dispatch()`:

```
void __cdecl (*em_XXX)();
```

Does that not work for `em_XXX`?

core/emulate.c
```diff
@@ -0,0 +1,908 @@
...
+             * Table 2-3. 32-Bit Addressing Forms with the SIB Byte */
+            ctxt->sib.value = insn_fetch_u8(ctxt);
+            reg_base  = ctxt->sib.base  | (ctxt->rex.b << 3);
+            reg_index = ctxt->sib.index | (ctxt->rex.x << 3);
+            if ((reg_base & 7) == 5 && ctxt->modrm.mod == 0) {
+                op->mem.ea += insn_fetch_s32(ctxt);
+            } else {
+                op->mem.ea += READ_GPR(reg_base, ctxt->address_size);
+            }
+            if (reg_index != 4) {
```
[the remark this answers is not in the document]
AlexAltea: The SDM wasn't entirely clear about this. If we used the condition `(reg_index & 7) != 4`, then we wouldn't be able to use `r12` as index register (e.g. `mov eax, dword ptr [r12*4]`). I've checked other decoders, and they all allow `r12`, i.e. they filter only `reg_index != 4` where (4 == REG_RSP).

tests/test_emulator.cpp
```diff
@@ -0,0 +1,464 @@
...
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <vector>
+#include <type_traits>
+
+#include "gtest/gtest.h"
+#include "keystone/keystone.h"
+
+extern "C" {
```
raphaelning: Maybe a better approach is to move `extern "C"` into the header files (`emulate.h` and `emulate_ops.h`)?

```
#ifdef __cplusplus
extern "C" {
#endif  // __cplusplus

// Function declarations

#ifdef __cplusplus
}
#endif  // __cplusplus
```

core/emulate.c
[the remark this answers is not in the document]
raphaelning: I see! That makes sense. It would be helpful to explain this in the code with a short comment.

core/include/emulate.h
```diff
@@ -0,0 +1,212 @@
...
+        struct {
+            uint8_t base  : 3;
+            uint8_t index : 3;
+            uint8_t scale : 2;
+        };
+        uint8_t value;
+    } sib;
+} em_context_t;
+
+em_status_t __stdcall em_decode_insn(struct em_context_t *ctxt, const uint8_t *insn);
```
raphaelning: The `__stdcall` decoration introduces in a build warning on macOS:

```
warning: calling convention '__stdcall' ignored for this target [-Wignored-attributes]
```

So basically, do we have to decorate any C function that calls any of the ASM routines? Why?

core/vcpu.c
AlexAltea: Fixed (as of e6ff1d3). I've added a `uint32_t flags` argument on memory callbacks (`read_memory` and `write_memory`) to allow such optimizations. 

After checking how other hypervisors handle this, I think in the future we could drop this `flags` arguments and export several emulator "helper" functions (e.g. like KVM's `emulator_can_use_gpa(ctxt)`), so that it's the callback itself the one who checks whether the optimization is allowed, and not the emulator who preemptively checks for allowed optimizations.

core/include/emulate_ops.h
AlexAltea: Although replacing function pointers with plain `void*` pointers seems like an easy/dirty fix for linking issues, my reasoning was:
1. The `em_XXX` handlers don't respect any calling convention (they are 2-instruction functions with arbitrarily chosen input/output registers).
2. C/C++ code will never call these functions, since they are handled internally by `fastop_dispatch`. The compiler doesn't need to understand what these pointers are.

However, I agree that for the sake of readability (and to make clear we are pointing to *code*), we should replace it with function pointers. I will apply your suggestion in my next commit.

core/include/emulate.h
AlexAltea: The problem is that *haxlib.vcxproj* and *IntelHaxm.vcxproj* are stdcall by default, and *tests.vcxproj* is cdecl by default.

As mentioned earlier, the solution would be to explicitly stating the correct calling convention. However, changing *tests.vcxproj* to stdcall causes an issue with *Google Test*, and changing haxlib/IntelHaxm to cdecl causes an issue with MASM code.

The only solution left, was explicitly telling tests that these two functions are stdcall, but that's a suboptimal fix. I'm still thinking how to properly deal with this situation.

AlexAltea:
> One thing we should investigate in advance is whether NASM supports VMX instructions, e.g. `VMLAUNCH`, `INVEPT`, etc.,

Yes, they are supported by NASM: https://www.nasm.us/xdoc/2.13.03/html/nasmdocb.html#section-B.1.13

HaxmDriver.sln
AlexAltea: @ras0219-msft No worries, thanks for the `export --nuget` hint.

Include paths are not working with the exported package. This are the steps to repeat it:
1. `vcpkg export keystone:x86-windows keystone:x64-windows --nuget --nuget-id=KeystoneEngine --nuget-version=0.9.1`
2. Update the *.nupkg file just to add a description, icon, etc. to the *.nuspec file.
3, Create a new VS solution+project, and install the *.nupkg file from a local repository.
4. Add `#include <keystone/keystone.h>` as press build.

Somehow MSVC complains that it didn't find that file, on all {Debug,Release} {Win32,x86,x64} modes, which is strange considering that the *.vcxproj file automatically added the `packages\keystoneengine.0.9.1\build\native\keystoneengine.targets` file.

raphaelning:
> Yes, they are supported by NASM: https://www.nasm.us/xdoc/2.13.03/html/nasmdocb.html#section-B.1.13

Great! Currently, `INVEPT` is not implemented for Win32, since it is not supported by Win32 MASM. So rewriting the existing assembly code in NASM has the extra benefit of getting that fixed :-)

> As of 8acf0a6, NASM is a NuGet dependency. Meaning that as soon as you press "Build solution", Visual Studio will download and add build customizations automatically

This works perfectly, thanks! Previously, I had some trouble with the VSNASM installation script, and had to resort to manual installation.

> Not sure how NuGet works with EWDK (I've never used it).

Once I've resolved the NuGet dependencies in VS2017, I can also build the solution with EWDK, so it was really easy. It seems possible to manually install a NuGet dependency from outside VS2017 as well, using the NuGet CLI tool, and I'll give it a try. Either way, if we leave aside the optional `tests` project, where the Keystone dependency is yet to be converted to NuGet, setting up the build environment for this PR is a smooth process now.

core/emulate.c
```diff
@@ -0,0 +1,978 @@
...
+
+/* Emulate accesses to guest memory */
+static bool is_translation_required(struct em_context_t *ctxt)
+{
+    const struct em_opcode_t *opcode = &ctxt->opcode;
+
+    if (opcode->flags & INSN_TWOMEM) {
+        return true;
+    }
+    if (opcode->flags & INSN_STRING && ctxt->rep) {
```
raphaelning: Why do `REP` instructions require translation? E.g. if the destination of `REP STOS` is an MMIO region, we'll get as many EPT violations as the repetition count (*CX), right? Each time we should be able to just use the VMCS GPA instead of translating ES:*DI to GPA, no?

core/vcpu.c
```diff
@@ -2067,551 +1957,175 @@ static int vcpu_simple_decode(struct vcpu_t *vcpu, struct decode *dc)
...
+        vcpu_translate(vcpu, ea, 0, &pa, NULL, false);
     }
 
-    // Calculate instruction length
-    len = (uint8)(num + 1);  // 1 for opcode
-    if (has_modrm) {
-        len++;
-        if (has_sib) {
-            len++;
+    if (is_mmio_address(vcpu, pa)) {
```
raphaelning: nit: When we skip translation, we should skip this check as well, or assert it is true, because we assume the GPA is an MMIO address.

core/vcpu.c
```diff
@@ -2067,551 +1957,175 @@ static int vcpu_simple_decode(struct vcpu_t *vcpu, struct decode *dc)
...
-                dec->opcode_dir <= OPCODE_MOVS_IOMEM_TO_IOMEM) {
-                state->_rsi += dec->size;
-            }
+    if (flags & EM_OPS_NO_TRANSLATION) {
+        pa = vmread(vcpu, VM_EXIT_INFO_GUEST_PHYSICAL_ADDRESS);
+    } else {
+        vcpu_translate(vcpu, ea, 0, &pa, NULL, false);
+    }
+
+    if (is_mmio_address(vcpu, pa)) {
```
raphaelning: nit: Same here.

core/include/emulate_ops.h
raphaelning: I see, that's really thoughtful of you! One last thing I'm curious about: why does decorating `em_XXX` with `__cdel` fix the linking error?

nit: Now that these functions don't really conform to the cdecl calling convention, could you add a short comment above the `typedef`, so people don't find the `__cdecl` part misleading?

core/include/emulate.h
raphaelning: Thanks. IIUC, your other proposal about replacing existing MASM code with NASM should be a proper fix. But for this PR, we can accept the current solution. The only issue is the macOS build warning, which I think can be fixed by something like:

```
#ifdef __i386__
#define EMCALL __stdcall
#else  // !__i386__
#define EMCALL
#endif  // __i386__

em_status_t EMCALL em_decode_insn(..);
em_status_t EMCALL em_emulate_insn(..)
```

HaxmDriver.sln
raphaelning: I noticed something last week, although I'm not sure if it's related to this problem: `HaxmDriver.sln` uses build configurations like `Win7 Debug`, `Win7 Release`, etc., but not `Debug`/`Release`. However, vcpkg seems to look for the latter:

```
> msbuild HaxmDriver.sln /p:Configuration="Win7 Debug" /p:Platform="x64"
...
VcpkgTripletSelection:
  Using triplet "x64-windows" from "<redacted>\vcpkg\installed\x64-windows\"
  Vcpkg is unable to link because we cannot decide between Release and Debug libraries. Please define the property VcpkgConfiguration to be 'Release' or 'Debug' (currently 'Win7 Debug').
...
```

Although these warnings do not fail the build.

This PR has actually added `Debug` and `Release` to `HaxmDriver.sln`, but I don't think they are usable now. Probably the right thing to do is map them to `Win7 Debug` and `Win7 Release`, respectively. Then I guess our VS solution will work better with vcpkg.

core/emulate.c
AlexAltea: > Why do `REP` instructions require translation?

Because with each iteration ES:*DI changes (`+= (!DF ? +1 : -1) * ctxt->address_size`), we will have access memory at different GPA's, but the optimization only works when the instruction accesses a single GPA. Additionally these GPAs can't be trivially computed by just increasing/decreasing the first GPA, since we might hit different pages.

raphaelning: I see. I was assuming we would resume guest execution at each iteration, which is what the old instruction emulator used to do. But it seems we are now bouncing back and forth between QEMU and the new emulator without entering guest mode at all, until we finish emulating the entire `REP` instruction. The two approaches differ in how many times we ask the hardware to execute the instruction for us (previously *CX times, now only once). Both work, but the new approach, as I now realize, is exactly the design you illustrated (i.e. the little arrows) ;-)

core/emulate.c
```diff
@@ -0,0 +1,978 @@
...
+    X8(F(em_inc, op_modrm_reg, op_none, op_none, 0)),
+    /* 0x48 - 0x4F */
+    X8(F(em_dec, op_modrm_reg, op_none, op_none, 0)),
+    /* 0x50 - 0x7F */
+    X16(N), X16(N), X16(N),
+    /* 0x80 - 0x8F */
+    G(opcode_group1, op_modrm_rm, op_imm, op_none, INSN_BYTEOP),
+    G(opcode_group1, op_modrm_rm, op_imm, op_none, 0),
+    G(opcode_group1, op_modrm_rm, op_imm, op_none, INSN_BYTEOP),
+    G(opcode_group1, op_modrm_rm, op_simm8, op_none, 0),
```
raphaelning: I checked all 8 instructions in this group. 0x80-0x83 all require the immediate operand to be sign-extended to the operand size. So instead of `op_simm8`, we probably need `op_simm`?

AlexAltea:
I've published a Keystone package at NuGet: https://www.nuget.org/packages/keystoneengine/.
Now building on Windows, including tests, is completely straightforward.

Regarding [your earlier comment](https://github.com/intel/haxm/pull/42/files/bf80191d162e72d9a4649a2ea0cb87e54dd13216#r189861281): While it's true that we have Configurations that don't match naming conventions expected by `vcpkg.targets`, the project `tests.vcxproj` (which is the only consumer of Keystone) **only** uses *Debug* and *Release* configurations, so we are good.

core/include/emulate_ops.h
raphaelning: > why does decorating em_XXX with __cdecl fix the linking error?

OK I think I get it. We add `__cdecl` just for its simple name mangling scheme. The error is due to stdcall name mangling, or more precisely, its use of the `@N` suffix. The `__cdecl` decoration forces C references to these symbols to drop the suffix.

raphaelning:
> Now building on Windows, including tests, is completely straightforward.

Indeed, thanks! This time I didn't rely on VS2017 to resolve the new NuGet dependency. It's very simple:

1. Download `nuget.exe` from https://www.nuget.org/downloads (no installation required).
2. Run `\path\to\nuget.exe restore` from the root of the HAXM source tree.

raphaelning:
The new patch looks good, thanks! The only remaining issue is the Clang warning about `__stdcall` being ignored for Mac (x86_64) build. I've suggested a possible fix:

https://github.com/intel/haxm/pull/42#discussion_r189856377

Could you take care of it? Then we can go ahead with the Git history revamp.

raphaelning:
> It turns out there are 2 more warnings, due to the use of __stdcall in emulate.c

Actually, there's no need to fix this now if you've already started cleaning up the history. Please feel free to address it in one of the final commits.

AlexAltea:
@raphaelning Done! :-)

I've checked that every commit along the way builds successfully on 32-bit and 64-bit mode (under Windows), so that we don't break `git bisect` or anything.

raphaelning:
Thank you so much! The final patches look great, and there's no warning from the macOS build. I'll proceed with the merge.

PR #49, opened 2018-05-02
Fixes #29. Ensure cr0/cr4 fixed bit settings prior vmxon.

### LINKED ISSUE #29: #GP on vmxon because cr0/cr4 fixed bits are not set
Hi,

The hax_create_vm() function when invoking invept() and then invept_smpfunc(), does not ensure before a "vmxon" that the fixed bits in CR0/CR4 are correctly set. It results in a general protection exception.

This may not happen under MacOS or Windows, because host CR0 is set by default with ET and NE. But trying to use HAXM in another environment can trigger the bug.

sduverger:
Hi,

Sorry for the delay. Holidays :) It fixes the vmxon issue #29 with code refactoring as discussed in PR46 (you can now ignore).

### DIFF
--- a/core/cpu.c
+++ b/core/cpu.c
@@ -502,7 +502,15 @@ uint8 is_vmcs_loaded(struct vcpu_t *vcpu)
 
 int debug_vmcs_count = 0;
 
-void restore_host_cr4_vmxe(struct per_cpu_data *cpu_data);
+void check_vmxon_coherency(struct per_cpu_data *cpu_data)
+{
+    if ((cpu_data->host_cr4 & CR4_VMXE) &&
+        (cpu_data->vmm_flag & VMXON_HAX)) {
+        // TODO: Need to understand why this happens
+        // (on both Windows and macOS)
+        hax_debug("HAX: VMM flag (VMON_HAX) is not clear!\n");
+    }
+}
 
 uint32 log_host_cr4_vmxe = 0;
 uint64 log_host_cr4 = 0;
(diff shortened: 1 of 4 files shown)


core/cpu.c
```diff
@@ -548,13 +556,107 @@ void hax_panic_log(struct vcpu_t *vcpu)
...
+ * Must be smp_call_function() safe.
+ * The success/failure logic is done in the caller.
+ */
+vmx_error_t vmxroot_enter(struct per_cpu_data *cpu_data)
+{
+    uint64 fc_msr;
+    uint64 cr0, cr0_f0, cr0_f1;
+    uint64 cr4, cr4_f0, cr4_f1;
+
+    smp_mb();
```
raphaelning: I'm not really familiar with memory barriers, and I'm curious why you use it here. Because of this line, `smp_mb()` is called twice for both `load_vmcs()` and `invept_smpfunc()` after the refactoring, but only once in the original code.

core/cpu.c
```diff
@@ -548,13 +556,107 @@ void hax_panic_log(struct vcpu_t *vcpu)
...
+/*
+ * Try to leave VMX-ROOT operations.
+ * Must be smp_call_function() safe.
+ * The success/failure logic is done in the caller.
+ */
+vmx_error_t vmxroot_leave(struct per_cpu_data *cpu_data)
+{
+    vmx_error_t err;
+
+    smp_mb();
```
raphaelning: Same here (see my comment on line 574).

core/cpu.c
```diff
@@ -502,7 +502,15 @@ uint8 is_vmcs_loaded(struct vcpu_t *vcpu)
 
 int debug_vmcs_count = 0;
 
-void restore_host_cr4_vmxe(struct per_cpu_data *cpu_data);
+void check_vmxon_coherency(struct per_cpu_data *cpu_data)
```
raphaelning: Make this function `static`? Actually, a better idea is to merge this function into `vmxroot_leave()` (after modifying `vmxroot_leave()` as I propose below).

raphaelning (commented):
Thanks a lot! I've made quite a few comments, but I wouldn't ask you to address all of them, since most of them are out of the scope of this patch ("refactoring") and can be addressed later. I'd pick out the one at `cpu.c:723` as the only required change.

If you don't have time to improve this patch, let me know if it would be easier if we took over and updated the PR directly.

sduverger:
Hello,

I'm really sorry for the answer delay. I had no more time slot to work on the subject :( and i think it will be difficult by the end of the year. I'll answer point by point to your comments, but please feel free to do what ever you think it's best with the PR. I did it a little bit in a hurry, to help, but based on your comments it tends to become a full-time dev job :) and i totally agree some part would have to be completely rewritten.

core/cpu.c
```diff
@@ -548,13 +556,107 @@ void hax_panic_log(struct vcpu_t *vcpu)
...
+}
+
+/*
+ * Restore host control registers value before entering vmxroot
+ * operations. Must be smp_func_call() safe.
+ */
+void restore_host_cr(struct per_cpu_data *cpu_data)
+{
+    if (cpu_data->vmm_flag & VMXON_RESTORE_CR0)
+        set_cr0(cpu_data->host_cr0);
```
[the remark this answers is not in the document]
sduverger: yeah, no debate :)

core/cpu.c
```diff
@@ -548,13 +556,107 @@ void hax_panic_log(struct vcpu_t *vcpu)
...
+
+    if (cpu_data->vmm_flag & VMXON_RESTORE_CR4)
+        set_cr4(cpu_data->host_cr4);
```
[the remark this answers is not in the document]
sduverger: this one too

core/cpu.c
sduverger: It's only compiler stuff optimization to prevent memory reordering operations ... Please refer to https://yarchive.net/comp/linux/memory_barriers.html, if you want to know more about it. You don't have to see it as a real function call.
In this situation we have to be safe toward smp function calls (like it's the case in HAXM code when you meet xxx_smpfunc()). Even if the 2 functions use to be called along side, if one day someone calls them appart ... it may fail.
As for all of your previous comments, i tried to keep the philosophy of the existing code without rewriting everything actually.

raphaelning:
Thanks, Stephane. I guess we should have kept this PR simple and leave the refactoring to another PR. Now we have created #69 for the latter. When that is merged, it should be easy to fix #29. But I understand that you may not have time to rebase your original patch and resubmit it. We can do that later and will give you credit in the commit message.

xianxiaoyin:
Build finished.

xianxiaoyin:

yiqisiben:

yiqisiben:

PR #61, opened 2018-06-13
Added IA32_TSC_AUX MSR

AlexAltea:
My guest system relies on `IA32_TSC_AUX` and this patch allows it to boot further.

Since `handle_cpuid_virtual` always returns a constant set of features, I have skipped the check for the `feat_rdtscp` flag before reading/writing this register and simply added it to `gmsr_list`. Note that the `rdtscp` instruction not yet implemented in the emulator, since none of my guests used it, but I can implement it for the sake of consistency if you want.

### DIFF
--- a/core/vcpu.c
+++ b/core/vcpu.c
@@ -929,6 +929,9 @@ void save_guest_msr(struct vcpu_t *vcpu)
         msr = (uint32)(IA32_PERFEVTSEL0 + i);
         gstate->apm_pes_msrs[i] = ia32_rdmsr(msr);
     }
+
+    // TODO: Check if host supports RDTSCP
+    gstate->tsc_aux = ia32_rdmsr(IA32_TSC_AUX);
 }
 
 void load_guest_msr(struct vcpu_t *vcpu)
@@ -952,6 +955,9 @@ void load_guest_msr(struct vcpu_t *vcpu)
         msr = (uint32)(IA32_PERFEVTSEL0 + i);
         ia32_wrmsr(msr, gstate->apm_pes_msrs[i]);
     }
+
+    // TODO: Check if host supports RDTSCP
+    ia32_wrmsr(IA32_TSC_AUX, gstate->tsc_aux);
 }
 
 static void save_host_msr(struct vcpu_t *vcpu)
@@ -976,6 +982,9 @@ static void save_host_msr(struct vcpu_t *vcpu)
         msr = (uint32)(IA32_PERFEVTSEL0 + i);
         hstate->apm_pes_msrs[i] = ia32_rdmsr(msr);
     }
+
+    // TODO: Check if host supports RDTSCP
+    hstate->tsc_aux = ia32_rdmsr(IA32_TSC_AUX);
 }
 
 static void load_host_msr(struct vcpu_t *vcpu)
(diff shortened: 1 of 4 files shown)


core/vcpu.c
```diff
@@ -2337,6 +2338,7 @@ static void handle_cpuid_virtual(struct vcpu_t *vcpu, uint32 a, uint32 c)
     uint32 cpu_features_ext =
             feat_execute_disable |
             feat_syscall         |
+            feat_rdtscp          |
```
raphaelning: 1. Since this effectively enables the guest to use the `RDTSCP` and `RDPID` instructions, we should also set the `ENABLE_RDTSCP` VM-execution control (Intel SDM Vol. 3C Table 24-7 _Definitions of Secondary Processor-Based VM-Execution Controls_):

https://github.com/intel/haxm/blob/415b5e463fd7fd59ac76972213ae4808ee302b2f/core/vcpu.c#L1147

2. nit: This feature is available in Nehalem and later (i.e. all generations of Intel Core i-3/5/7) Intel CPUs (see Intel SDM Vol. 3B 17.16.2). However, officially, HAXM also supports a few pre-Nehalem models, e.g. Core 2 Duo. Therefore, it would be better to check the host CPUID for availability of this feature before exposing it to the guest. But I guess that's too much burden for you. So you could just add a TODO comment, we'll address it in a later patch.

raphaelning: BTW, HAXM disables `RDTSC_EXITING`, but enables `USE_TSC_OFFSETTING` VM-execution control. According to Intel SDM Vol. 3C 25.3, `RDTSCP` won't incur a VM exit in this case.

AlexAltea:
I've addressed all issues except checking whether the host supports `rdtscp`, thus breaking pre-Nehalem hosts. I was thinking in adding the following changes to test for host features:

1. During HAXM initialization get relevant features via `cpuid` and store them in `uint32_t[N]` array.
2. Define `X86_FEATURE_*` values as *(index << 5) | (bit)*, where index is an index into the previous array, and bit says the bit-index of the feature flag.
3. Define a `cpu_has_feature` function in `cpu.c` (this is the source file for host-related functions, right?).

This would be similar to what KVM does by relying on Linux' `boot_cpu_has` function. Of course, since HAXM is multiplatform, it'd be to check for the features by ourselves (as described above) rather than relying in OS utility functions.

raphaelning:
The patch looks good now. We'll run a few tests before merging it.

Indeed, currently HAXM lacks an API to easily query host CPU features, and what you have proposed is definitely a good design.

> 2. Define X86_FEATURE_* values as (index << 5) | (bit), where index is an index into the previous array, and bit says the bit-index of the feature flag.

This is a scalable way to organize all the features. I was wondering if there's BSD-compatible code that defines similar macros for us to reuse. I checked FreeBSD, but it basically puts `index` in the macro name (e.g. `CPUID2_SSE3`), which is not very convenient. Anyway, we don't need many `index`es at first, so defining them manually should be okay.

> 3. Define a cpu_has_feature function in cpu.c to do queries by relying on said array (I'm assuming cpu.c is the source file for host-related functions).

Yes, we could put the new API in `core/cpu.c`. But what about creating a new pair of source files, `core/cpuid.{c, h}`? In the beginning we only need the host CPUID query API,  but we might want to add more over time, e.g. caching the CPUID result for each guest, and eventually allowing each guest to customize its own vCPU features.

AlexAltea:
> The patch looks good now. We'll run a few tests before merging it.

Great! If tests are passed, feel free to merge and I'll address the host feature checks in a separate PR.

> Yes, we could put the new API in core/cpu.c. But what about creating a new pair of source files, core/cpuid.{c, h}?

That sounds much better, I will do that then.

raphaelning:
> I'll address the host feature checks in a separate PR.

Thanks!

PR #62, opened 2018-06-14
Added bextr/andn instructions to emulator

AlexAltea:
This adds support for decoding VEX prefixes and emulating the `bextr` and `andn` instructions, with which some some guest kernels do MMIO accesses.

Since these instructions are part of the BMI1 extension, using the "*fast handlers*": `em_bextr` and  `em_andn`, i.e. running a register-only version of the instruction would be an issue on pre-Haswell CPUs. Until we implement a way to detect host features, I've added software-emulated handlers `em_bextr_soft` and `em_andn_soft`, respectively.

### DIFF
--- a/core/emulate.c
+++ b/core/emulate.c
@@ -92,6 +92,7 @@
 DECL_DECODER(op_none);
 DECL_DECODER(op_modrm_reg);
 DECL_DECODER(op_modrm_rm);
+DECL_DECODER(op_vex_reg);
 DECL_DECODER(op_moffs);
 DECL_DECODER(op_simm);
 DECL_DECODER(op_simm8);
@@ -137,6 +138,8 @@ DECL_DECODER(op_si);
     F2_BV(_handler, op_acc, op_simm, op_none, (_flags))
 
 /* Soft-emulation */
+static void em_andn_soft(struct em_context_t *ctxt);
+static void em_bextr_soft(struct em_context_t *ctxt);
 static void em_mov(struct em_context_t *ctxt);
 static void em_movzx(struct em_context_t *ctxt);
 static void em_xchg(struct em_context_t *ctxt);
@@ -239,11 +242,17 @@ static const struct em_opcode_t opcode_table_0F[256] = {
 };
 
 static const struct em_opcode_t opcode_table_0F38[256] = {
-    /* 0x00 - 0xFF */
+    /* 0x00 - 0xEF */
     X16(N), X16(N), X16(N), X16(N),
     X16(N), X16(N), X16(N), X16(N),
     X16(N), X16(N), X16(N), X16(N),
-    X16(N), X16(N), X16(N), X16(N),
+    X16(N), X16(N), X16(N),
+    /* 0xF0 - 0xFF */
+    X2(N),
+    I(em_andn_soft, op_modrm_reg, op_vex_reg, op_modrm_rm, INSN_MODRM),
+    X4(N),
+    I(em_bextr_soft, op_modrm_reg, op_modrm_rm, op_vex_reg, INSN_MODRM),
+    X8(N),
 };
 
 static const struct em_opcode_t opcode_table_0F3A[256] = {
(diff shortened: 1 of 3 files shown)


core/emulate.c
```diff
@@ -854,10 +905,49 @@ em_status_t EMCALL em_decode_insn(struct em_context_t *ctxt, const uint8_t *insn
...
+        } else {
+            ctxt->vex.value = insn_fetch_u8(ctxt);
+            ctxt->vex.r = ctxt->vex.w;
+            ctxt->vex.w = 1;
+            ctxt->vex.x = 1;
+            ctxt->vex.b = 1;
+            ctxt->vex.m = 1;
+        }
+        if (ctxt->mode != EM_MODE_PROT64 && (!ctxt->vex.r || !ctxt->vex.x)) {
+            return EM_ERROR;
```
raphaelning: nit: According to Intel SDM Vol. 2A 2.3.5.{2, 3, 6}, in this case we should reinterpret `vex.prefix` as either the `LDS` (0xC5) or `LES` (0xC4) instruction. We don't have to support emulating these two uncommon instructions (although, in theory, they may do MMIO), but it would be useful to add a comment to explain the situation or leave it as a TODO.

core/emulate.c
```diff
@@ -854,10 +905,49 @@ em_status_t EMCALL em_decode_insn(struct em_context_t *ctxt, const uint8_t *insn
...
+        }
+
+        ctxt->rex.r = ~ctxt->vex.r;
+        ctxt->rex.w = ~ctxt->vex.w;
```
raphaelning: [Unsure] Do you mean `b` instead of `w`? I'm not sure if we always have REX.W = ~VEX.W, but the SDM is pretty clear about REX.B = ~VEX.B (Vol. 2A 2.3.5.4).

core/emulate.c
```diff
@@ -854,10 +905,49 @@ em_status_t EMCALL em_decode_insn(struct em_context_t *ctxt, const uint8_t *insn
...
+        break;
+
+    case 0xC4:
+    case 0xC5:
+        /* Intel SDM Vol. 2A: 2.3.5 The VEX Prefix */
+        ctxt->vex.prefix = b;
+        if (b == 0xC4) {
+            ctxt->vex.value = insn_fetch_u16(ctxt);
+        } else {
+            ctxt->vex.value = insn_fetch_u8(ctxt);
```
raphaelning: [Unsure] Should the right-hand side be `insn_fetch_u8(ctxt) << 8` instead?

AlexAltea: True, thank you. My brain was accidentally in big-endian mode. :-)

core/emulate.c
AlexAltea: Yeah, that's a typo. I should have checked the code more carefully.

core/emulate.c
```diff
@@ -765,6 +783,38 @@ static em_status_t decode_op_si(em_context_t *ctxt,
...
+{
+    uint64_t temp;
+    uint64_t signbit = 63;
+    temp = ~ctxt->src1.value & ctxt->src2.value;
+    if (ctxt->operand_size == 4) {
+        temp &= 0xFFFFFFFF;
+        signbit = 31;
+    }
+    ctxt->dst.value = temp;
+    ctxt->rflags = 0;
```
raphaelning: Hmm, probably we should only touch the status flags (OSZAPC), since this instruction is similar to the _fastop_ instructions? Clearing other flags may have side effects, e.g. bit 1 (reserved) is supposed to be fixed to 1.

core/emulate.c
```diff
@@ -765,6 +783,38 @@ static em_status_t decode_op_si(em_context_t *ctxt,
...
+    start = ctxt->src2.value & 0xFF;
+    len = (ctxt->src2.value >> 8) & 0xFF;
+    temp = ctxt->src1.value;
+    if (ctxt->operand_size == 4) {
+        temp &= 0xFFFFFFFF;
+    }
+    temp >>= start;
+    temp &= ~(~0ULL << len);
+    ctxt->dst.value = temp;
+    ctxt->rflags = (temp == 0) ? RFLAGS_ZF : 0;
```
raphaelning: Same here. Although the SDM says "All other flags are cleared", I think it means all _status_ flags.

core/emulate.c
raphaelning: :-) I was wondering why the unit tests didn't catch this. Now I see--both `ANDN` and `BEXTR` require a 3-byte VEX prefix...

AlexAltea:
All fixed. Sorry for the status flags issue: I forgot that only *fastop*'s had masked eflags accesses. :-)

raphaelning:
Thanks! We'll test the patch and merge it.

> I forgot that only fastop's had masked eflags accesses.

Right, it would be convenient if there was a helper function that only updates status flags. I guess we can add it when we support more ALU instructions that require software emulation.

PR #63, opened 2018-06-14
Refactored host feature detection via CPUID

AlexAltea:
This patch allows checking for host features to conditionally enable guest features added in https://github.com/intel/haxm/pull/61 and https://github.com/intel/haxm/pull/62, since the old code only checked for host-support for few hardcoded features (VMX, NX, EMT64T).

## Design

The newer one is designed following the comments following https://github.com/intel/haxm/pull/61#issuecomment-396995254:
- Old features (`feat_*`) have been replaced by unique 32-bit integer keys (`X86_FEATURE_*`) which are much more flexible than the approach used in [Linux](https://elixir.bootlin.com/linux/v4.17.1/source/arch/x86/include/asm/cpufeatures.h) and [FreeBSD](http://web.mit.edu/freebsd/head/sys/x86/include/specialreg.h): These keys encode the entire CPUID input state (both EAX and leaf registers), the cache index, and the bit position in just 25-bits.
- All CPUID-related definitions/code have been moved to `cpuid.h`/`cpuid.c`. These files are self-contained and provide a simple interface to: (1) initialize a CPUID cache, (2) check for features (either in the cache or re-querying CPUID). This interface is consumed by `cpu.c`.

---

Aditionally, I had few questions:
- I have noticed that during each (non-APM) iteration of *save_guest_msr*, *load_guest_msr*, *save_host_msr*, *load_host_msr*, HAXM checks if `cpu_has_emt64_support()` is true. This means a total of _2 * (NR_GMSR + NR_HMSR) = 22_ queries. Is it really necessary to query it every time, i.e. does the feature set really change? This probably causes some overhead, specially without cache.
- Some code (see `cpu_init_vmx`) flags host features for each CPU: Is it safe to assume that all host CPU cores present the same features? That would be nice, because otherwise we would need to have N CPUID caches for N cores.


raphaelning:
Thanks for initiating this effort! I haven't reviewed the code yet, but let me answer the questions first:

> Is it really necessary to query it every time, i.e. does the feature set really change? This probably causes some overhead, specially without cache.

This is a good catch! It's just the _lm_ (Long Mode) feature flag, which should never change and should be cached.

> Is it safe to assume that all host CPU cores present the same features?

I believe so. It's safe to assume nobody would run HAXM on a system with heterogeneous CPUs. As you said, KVM only checks the boot CPU (BSP), but I guess we can pick an arbitrary CPU.

core/vcpu.c
```diff
@@ -2314,43 +2323,43 @@ static void handle_cpuid_virtual(struct vcpu_t *vcpu, uint32 a, uint32 c)
...
+            FEATURE(POPCNT);
 
     uint32 cpu_features_ext =
-            feat_execute_disable |
-            feat_syscall         |
-            feat_rdtscp          |
-            feat_em64t;
+            FEATURE(NX)         |
+            FEATURE(SYSCALL)    |
+            FEATURE(RDTSCP)     |
```
raphaelning: We need to make this conditional on `cpu_has_feature(X86_FEATURE_RDTSCP)`.

core/cpuid.c
```diff
@@ -0,0 +1,114 @@
...
+typedef union cpuid_feature_t {
+    struct {
+        uint32_t index        : 5;
+        uint32_t root_key_lo  : 5;
+        uint32_t root_key_hi  : 1;
+        uint32_t leaf_reg     : 2;
+        uint32_t leaf_key     : 5;
+        uint32_t value_reg    : 2;
+        uint32_t value_bit    : 5;
+    };
```
raphaelning: nit: Add a `reserved` field for the remaining bits, so the reader won't assume all 32 bits are used.

core/include/cpuid.h
```diff
@@ -0,0 +1,249 @@
...
+    X86_FEATURE_PDPE1GB       = LEAF(26), /* 0x04000000  Gibibyte pages */
+    X86_FEATURE_RDTSCP        = LEAF(27), /* 0x08000000  RDTSCP Instruction */
+    X86_FEATURE_EM64T         = LEAF(29), /* 0x20000000  Long Mode */
+    X86_FEATURE_3DNOWEXT      = LEAF(30), /* 0x40000000  Extended 3DNow! */
+    X86_FEATURE_3DNOW         = LEAF(31), /* 0x80000000  3DNow! */
+#undef LEAF
+};
+
+// Functions
+void cpuid_cache_init(uint32_t* cache);
```
raphaelning: 1. nit: Usually, if we allow the caller to pass a dynamic array like `cache`, we should also require them to provide the array size in a separate parameter, instead of assuming a minimum size (`CPUID_CACHE_SIZE`).
2. nit: Use C-style pointer declaration: `uint32_t *cache`

core/include/cpuid.h
```diff
@@ -0,0 +1,249 @@
...
+bool cpuid_cache_has_feature(uint32_t* cache, uint32_t feature_key);
```
raphaelning: nit: Pointer declaration style

core/cpuid.c
```diff
@@ -0,0 +1,114 @@
...
+                                 uint32_t leaf_val, uint32_t leaf_reg)
+{
+    cpuid_args_t args;
+    args.regs[leaf_reg] = leaf_val;
+    args.eax = eax;
+    // __handle_cpuid(&args); // TODO
+    return args.regs[value_reg];
+}
+
+void cpuid_cache_init(uint32_t* cache)
```
raphaelning: nit: Use C-style pointer declaration: `uint32_t *cache`

core/cpuid.c
```diff
@@ -0,0 +1,114 @@
...
+{
+    cache[0] = cpuid_query_root(0x00000001, CPUID_REG_ECX);
+    cache[1] = cpuid_query_root(0x00000001, CPUID_REG_EDX);
+    cache[2] = cpuid_query_leaf(0x00000007, CPUID_REG_ECX, 0x00, CPUID_REG_ECX);
+    cache[3] = cpuid_query_leaf(0x00000007, CPUID_REG_EBX, 0x00, CPUID_REG_ECX);
+    cache[4] = cpuid_query_root(0x80000001, CPUID_REG_ECX);
+    cache[5] = cpuid_query_root(0x80000001, CPUID_REG_EBX);
+}
+
+bool cpuid_cache_has_feature(uint32_t* cache, uint32_t feature_key)
```
raphaelning: nit: Pointer declaration style

core/cpuid.c
```diff
@@ -0,0 +1,114 @@
...
+    cpuid_args_t args;
+    args.regs[leaf_reg] = leaf_val;
+    args.eax = eax;
+    // __handle_cpuid(&args); // TODO
+    return args.regs[value_reg];
+}
+
+void cpuid_cache_init(uint32_t* cache)
+{
+    cache[0] = cpuid_query_root(0x00000001, CPUID_REG_ECX);
```
raphaelning: nit: Maybe we could adjust the API and make fewer CPUID calls:

```
cpuid_args_t res;

// res is output only
cpuid_query_root(&res, 0x00000001);
cache[0] = res.ecx;
cache[1] = res.edx;

// Assuming (sub)leaf register is always ECX
cpuid_query_leaf(&res, 0x00000007, 0x00);
cache[2] = res.ecx;
cache[3] = res.ebx;
...
```

core/include/cpuid.h
```diff
@@ -0,0 +1,249 @@
...
+bool cpuid_has_feature(uint32_t feature_key);
```
raphaelning: I'm considering renaming these functions to:

* `cpuid_cache_init` => `cpuid_host_init`
* `cpuid_cache_has_feature` => `cpuid_host_has_feature`
* `cpuid_has_feature` => `cpuid_host_has_feature_uncached`

Reasoning:
1. We may want to cache the CPUID result for each guest later (instead of wasting time to recompute the result at each guest CPUID call), so we need to distinguish between host and guest CPUID APIs.
2. Since the CPU feature set should never change for a given host, the uncached variant of the host CPUID query API should be rarely used and can have a long name. Meanwhile, we assign the short name to the more common, cached variant.

core/include/cpuid.h
```diff
@@ -0,0 +1,249 @@
...
+ * - value_reg    : 2;  Output register index
+ * - value_bit    : 5;  Bit position of the feature
+ *
+ * Additionally root_key is derived from the EAX register as follows:
+ * - root_key_lo  : 5;  Masked least-significant 5-bits of EAX
+ * - root_key_hi  : 1;  Masked most-significant 1-bit of EAX
+ *
+ * This allows representing any feature obtained via:
+ * - Root register (EAX): 00h-1Fh and 80000000h-8000001Fh
+ * - Leaf register (ECX, EDX, EBX): 00h-1Fh
```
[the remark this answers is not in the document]
AlexAltea: Regarding to your points:

1. You are right, I will use the leaf/subleaf naming to remain consistent with the Intel SDM (which is also what I quoted in the code comments). My previous terms were chosen arbitrarily.

2. True, I have transformed that field into boolean 1-bit field named `subleaf_used`.

3. True, I have extended `leaf_hi` (previously `root_key_hi`) to 2-bits. However, I have kept the remaining "key" fields unmodified (5-bits), since no leaf or subleaf has a lower part larger than 0x1F, both on Intel and AMD reference manuals. I think we should only extend fields when necessary to avoid extending the wrong fields and later running out of extra bits for other fields. Since these fields are only consumed **internally** by *cpuid.h*, increasing their size can be done at any time easily.

core/include/cpuid.h
```diff
@@ -0,0 +1,249 @@
...
+
+    /*
+     * Intel SDM Vol. 2A: Table 3-8. Information Returned by CPUID Instruction
+     * Extended Function CPUID Information
+     * Features for CPUID with EAX=80000001h stored in ECX
+     */
+#define LEAF(bit) \
+    FEATURE_KEY_ROOT(4, 0x80000001, CPUID_REG_ECX, bit)
+    X86_FEATURE_LAHF          = LEAF(0),  /* 0x00000001  LAHF/SAHF Instructions */
+    X86_FEATURE_SVM           = LEAF(2),  /* 0x00000004  Secure Virtual Machine */
```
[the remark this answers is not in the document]
AlexAltea: Yes, some of these came from AMD manuals. But on second thoughts, I will remove all AMD-specific features, at least until we add AMD-support at a later point (if).

core/include/cpuid.h
AlexAltea: Good point. I thought of addressing it in a different way though:
```c
typedef uint32_t cpuid_cache_t[CPUID_CACHE_SIZE];
```
And then replacing all `uint32_t *cache` occurrences with `cpuid_cache_t *cache`. Therefore, all consumers of the *cpuid* interface can simply allocate a `cpuid_cache_t` object and stop worrying about sizes. What do you think?

core/include/cpuid.h
AlexAltea: I fully agree with the names you suggested.

Regarding the 2nd point, although it's true, I don't think these names are too important. After all, these functions are going to be wrapped by *cpu.c*, so the actual functions that will appear all over the codebase are `cpu_has_feature` (for host features).

AlexAltea:
All fixed (on dfa8810). I will update things as needed. Once you give green-light to the patch, I'll squash them into the following two commits:
1. Added x86 feature definitions and new CPUID interface.
2. Updated host/guest CPUID feature management.

The only thing missing to this PR is:
1. Updating the `__handle_cpuid` assembly function to take an `cpuid_args_t` pointer.
2. Initializing and using the *cpuid* cache in `cpu.c`.

core/cpuid.c
```diff
@@ -0,0 +1,115 @@
...
+typedef union cpuid_feature_t {
+    struct {
+        uint32_t index        : 5;
+        uint32_t leaf_lo      : 5;
+        uint32_t leaf_hi      : 2;
+        uint32_t subleaf_key  : 5;
+        uint32_t subleaf_used : 1;
+        uint32_t reg          : 2;
+        uint32_t bit          : 5;
+        uint32_t /*reserved*/ : 8;
```
raphaelning: 8 => 7

core/cpu.c
```diff
@@ -58,6 +59,17 @@ static int cpu_nx_enable()
     return effer & 0x800;
 }
 
+bool cpu_has_feature(uint32_t feature)
+{
+    static cpuid_cache_t cache = {
+        .initialized = 0
+    };
+    if (!cache.initialized) {
+        cpuid_host_init(&cache);
```
raphaelning: nit: Is there still a race condition, when two host CPUs both see the cache as uninitialized and try to fill it at the same time? This may not cause a problem, since they only write to the cache and should write the same data. Still, it would be nice if only one CPU performed the initialization. C++11 actually guarantees that static local initializers are only called once:

https://en.cppreference.com/w/cpp/language/storage_duration#Static_local_variables

I'm not sure if we can take advantage of that, because this is C code, and our compilers may not support or be configured to use C++11.

A simple fix would be to call `cpu_has_feature()` with an arbitrary feature very early, e.g. before this call:

https://github.com/intel/haxm/blob/2c0ac6c203e30f61f09184f8d36d6d2b5396234d/core/hax.c#L543

so as to force CPU cache initialization in a non-`smp_call_function()` context.

raphaelning:
[haxm-PR63-mac-build-fix.zip](https://github.com/intel/haxm/files/2113842/haxm-PR63-mac-build-fix.zip)

The first patch is not really related to this PR, but my second patch relies on it, so I've also uploaded it to GitHub: #65. I'm not sure what works best for you: I could merge it into master and then you rebase your work on top of it, or you could pick it into your final patch set for this PR.

The second patch should probably be incorporated into your final patches for this PR.

AlexAltea:
I applied your second patch `0002-darwin-Fix-CPUID-module-build.patch`, since it's relevant to this PR. The other one can be merged via #65 (maybe will require rebasing).

The race condition is fixed by adding a `cpu_init_feature_cache` function, since I felt that calling `cpu_has_feature` for no reason was not good enough.

I'll rewrite the Git history for this PR now.

PR #64, opened 2018-06-19
Codebase cleanup and assembly refactoring

AlexAltea:
I just cleaned up parts of the codebase, hoping to make an improvement to overall readability. I hope it's not too biased with my own tastes. If so, please feel free to reject any change and I will rewrite the patch cherry-picking only the commits you deem acceptable.

## Summary

- On e74a4ec:
  - Added `VMX_EXIT_` prefix to VMX exit reasons enum. Goal is to avoid polluting the global namespace with ambiguous names such as "*CPUID_INSTRUCTION*" that are confusing without context (e.g. it could refer to the "cpuid" opcode, or the "cpuid" exit reason, or anything else involving that instruction).
  - Removed `_INSTRUCTION` suffix from VMX exit reasons enum. Goal is to keep consistency with `exit_*` handlers (which don't have such such suffix). It's also less verbose this way.
  - Added a couple of missing exit reasons (57-64), for the sake of completeness.
  - Renamed exceptions to `VECTOR_XX`, to remain consistent with the Intel SDM.
  - Fixed few English grammar issues.
- On 6f693d4:
  - Updated the exit handler list and `config_t` to use designated initializers
  - Fixed small typos.
- On 5ea877d:
  - Replaced platform-specific **.asm* files with general NASM files.
  - Using NASM macros to make most assembly functions cross-platform (32-bit/64-bit), and to automate the generation of certain functions, via `function_*` macros.
  - Replaced the `vmx_error_t` with `vmx_result_t` enum type with *VMX_SUCCEED* (0), *VMX_FAIL_INVALID* (1), *VMX_FAIL_VALID* (2) values. Transforming EFLAGS into these values happens in the assembly functions via the macro `vmx_check`. The new enum type `vmx_error_t` contains the error codes for *VMX_FAIL_VALID*. This change required renaming a couple of variables.
  - Replaced `vmx_error_t cpu_vmentry_failed(...)` to `void cpu_vmentry_failed(...)` since the return value is useless and always ignored by the caller.
  - Renamed `load_kernel_ldt` to `set_kernel_ldt` to remain consistent with IDT/GDT setters.
  - Moved few VMX-specific function declarations from `asm.h` to `vmx.h`.
  - Replaced platform-specific wrappers for IA32 assembly functions with generic wrapper.

As always, based on the changes you accept, I'll rewrite the Git history to avoid polluting the repository.


raphaelning:
> I hope it's not too biased with my own tastes.

Not at all! On the contrary, I'm really grateful that you took your time to improve the code quality of HAXM. I can see that you spot something wrong and actually went on to fix it--that's really admirable.

I've reviewed the first three patches and I think they are good to go. Please allow me more time on the last one, which is especially exciting because it is another big step toward a unified assembler for all supported platforms.

core/ia32.asm
```diff
@@ -0,0 +1,314 @@
...
+    mov edx, esi
+    wrmsr
+    push esi
+    push edi
+    ret
+%endif
+
+function get_kernel_rflags
+    pushf
+    pop ax
```
raphaelning: `ax` => `reg_ret_16`

core/ia32.asm
```diff
@@ -0,0 +1,314 @@
...
+function asm_fxsave
+    fxsave [reg_arg1]
+    ret
+
+function asm_rdmsr
+%ifidn __BITS__, 64
+    mov rcx, reg_arg1
+    rdmsr
+    shl rdx, 32
+    or rax, rdx
```
raphaelning: `rax` => `reg_ret`

core/ia32.asm
```diff
@@ -0,0 +1,314 @@
...
+    mov [reg_arg2 + qword_struct.lo], eax
+    mov [reg_arg2 + qword_struct.hi], edx
+    ret
+%endif
+
+function asm_rdtsc
+%ifidn __BITS__, 64
+    rdtsc
+    shl rdx, 32
+    or rax, rdx
```
raphaelning: `rax` => `reg_ret`

core/haxlib.vcxproj
```diff
@@ -180,6 +180,7 @@
     <NASM Include="emulate_ops.asm">
       <FileType>Document</FileType>
     </NASM>
+    <NASM Include="ia32.asm" />
```
raphaelning: nit: So we don't really need to specify the `FileType` tag? Then shall we just delete line 181?

AlexAltea: Not sure what caused the `<FileType>` tag to be added, but it's unnecessary. It's deleted now.

AlexAltea:
Sorry for the issues, I was waiting for the new PR to be merged to rebase and review+test the code. There's still an annoying issue under 32-bit configuration regarding calling conventions. I'll fix it and let you know when everything's ready.

raphaelning:
Sure.

> There's still an annoying issue under 32-bit configuration regarding calling conventions.

I see. I managed to dig out [your old comment](https://github.com/intel/haxm/pull/42#discussion_r189585746) about the previous Win32 linking issue. Not sure if it's related to this one:

> The problem is that haxlib.vcxproj and IntelHaxm.vcxproj are stdcall by default, and tests.vcxproj is cdecl by default.
> 
> As mentioned earlier, the solution would be to explicitly stating the correct calling convention. However, changing tests.vcxproj to stdcall causes an issue with Google Test, and changing haxlib/IntelHaxm to cdecl causes an issue with MASM code.
> 
> The only solution left was explicitly stating that these two emulator functions are stdcall, but that's a suboptimal fix. I'm still thinking how to properly deal with this situation.

I also noticed that the new `ia32.asm` uses the _fastcall_ convention for Win32, whereas `emulate_ops.asm` (`fastop_dispatch()`) uses _cdecl_. How do you choose the calling convention?

core/ia32.asm
```diff
@@ -0,0 +1,351 @@
...
+function get_kernel_tr_selector
+    str reg_ret_16
+    ret
+
+function get_kernel_ldt
+    sldt reg_ret_16
+    ret
+
+function get_kernel_gdt
+    sgdt [reg_arg1]
```
raphaelning: nit: Does `sgdt` not require the operand to be labeled `fword ptr` like `lgdt` does?

core/ia32.asm
```diff
@@ -0,0 +1,351 @@
...
+    ret
+
+function get_kernel_idt
+    sidt [reg_arg1]
```
raphaelning: nit: Is `fword ptr` not necessary?

darwin/hax_driver/com_intel_hax/asm/vmcs.c
```diff
@@ -1,970 +0,0 @@
...
- * POSSIBILITY OF SUCH DAMAGE.
- */
-
-#include "../../../../core/include/ia32.h"
-#include "../../../../core/include/vmx.h"
-#include "../../../../core/include/types.h"
-#include "../../../../include/hax.h"
-#include "../../../../core/include/vcpu.h"
-
-void hax_disable_preemption(preempt_flag *eflags)
```
raphaelning: The Windows version of this function is defined in `windows/hax_wrapper.c` and implemented with pure C code. I have no idea why the Mac version is defined in `vmcs.c`, but we can't just delete it. Instead, let's move it into `darwin/..../hax_wrapper.cpp`, and I think it's okay to keep the inline assembly.

darwin/hax_driver/com_intel_hax/asm/vmcs.c
```diff
@@ -1,970 +0,0 @@
...
-        "pushfd         \n\t"
-        "pop %0"
-        : "=r" (flags)
-    );
-#endif
-    *eflags = flags;
-    hax_disable_irq();
-}
-
-void hax_enable_preemption(preempt_flag *eflags)
```
raphaelning: Same here. Let's move this one into `hax_wrapper.cpp` as well.

raphaelning:
I wonder if we should split this PR into multiple ones, so we can merge commits early, as soon as they become ready. Right now we have a small PR #69, which is ready to merge but has some conflicts with this PR, at least with commit `b67284a` (`vmx_ops.asm`). If we can merge part of this PR first, I think you'll have an easier time rebasing the rest of it.

AlexAltea:
Sorry, last week and part of this week I was/will be mostly AFK for work reasons. Don't worry about merging other PRs first, I don't mind rebasing. Also, I expect to finish this PR next weekend. Thanks for your other comments, I will try to update my patch as soon as possible :)

core/ia32.asm
```diff
@@ -0,0 +1,351 @@
...
+%endif
+
+;
+; Helpers
+;
+%macro function 1
+    global %1
+    %1:
+%ifidn __CONV__, x32_cdecl
+    mov reg_arg1, [esp + 0x4]
```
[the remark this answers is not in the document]
AlexAltea: > How about adding a second parameter to `%macro function`, for declaring the number of parameters the function takes?

Agreed. I will do that.

core/vmx_ops.asm
```diff
@@ -0,0 +1,194 @@
...
+    xor reg_ret, reg_ret
+    vmread reg_ret, reg_arg1
+    ret
+
+function asm_vmwrite
+    vmwrite reg_arg1, reg_arg2
+    ret
+
+function asm_vmxrun
+    ; TODO
```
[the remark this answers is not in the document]
AlexAltea: This is implemented now.

core/vmx_ops.asm
```diff
@@ -0,0 +1,194 @@
...
+    ret
+
+function asm_invept
+    ; TODO
```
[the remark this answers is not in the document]
AlexAltea: This is implemented now.

AlexAltea:
Thanks for merging the recent PRs, they had really useful stuff! Rebasing this patch was really simple. Sorry for the delays, I don't like keeping PRs idling, but I got overwhelmed by real-life work.

I've finished all the features I wanted to implement so far. There's more to come, but that's for another PR since this one is already big enough. Since your last review, I've fixed the issues brought up by your comments and additionally:
- On `881c2f9`:
  - Implemented `asm_vmxrun` and `asm_invept`.
- On `d555074`:
  - Replaced the `vmx_error_t` with `vmx_result_t` enum type with *VMX_SUCCEED* (0), *VMX_FAIL_INVALID* (1), *VMX_FAIL_VALID* (2) values. Transforming flags into these values occurs in the assembly functions via the macro `vmx_check`. Consequently, `vmx_error_t` contains the error codes for *VMX_FAIL_VALID*. This change required renaming a couple of variables.
  - Replaced `vmx_error_t cpu_vmentry_failed(...)` to `void cpu_vmentry_failed(...)` since the return value is useless and always ignored by the callet.
  - Renamed `load_kernel_ldt` to `set_kernel_ldt` to remain consistent with IDT/GDT setters.
  - Moved few VMX-specific function declarations from `asm.h` to `vmx.h`.

core/vmx_ops.asm
```diff
@@ -0,0 +1,358 @@
...
+    mov ebx, 6C14h
+    mov eax, esp
+    sub eax, 4h
+    vmwrite ebx, eax
+    pop ebx
+    pop eax
+    push eax
+    push ebx
+    ; push the state
+    push reg_arg1
```
raphaelning: Here we need to copy the `state` pointer to `eax` as the original 32-bit MASM code used to do:

```
mov eax, reg_arg1
push eax
```

Otherwise, since `reg_arg1 == ecx != eax`, we'll be loading ECX with the wrong data at line 317. And we can't fix that by simply replacing `eax` with `reg_arg1` in lines 317-323, because line 317 will modify ECX and thus `reg_arg1`.

core/ia32.asm
raphaelning: Interesting. But it makes sense, because `lgdt`, etc. don't really use other (standard) operand sizes, so removing `fword ptr` doesn't really create any ambiguity.

raphaelning:
> Once you approve the patch, I'll rewrite the history into three commits

Yes, that's a good plan. Please feel free to incorporate the miscellaneous changes (spelling, coding style, etc.) into these three at your convenience.

AlexAltea:
All done, patch history is clean now. An overview of every change is updated at https://github.com/intel/haxm/pull/64#issue-195689921.

raphaelning:
Thanks! Clang was able to spot a few more issues when we tried to build this PR on Mac (sorry I still don't know how to comment on unchanged code):

1. `paniced` => `panicked`
   * https://github.com/AlexAltea/haxm/blob/10597c3a90c0d41b26d23bcd17a69c83633c7d9e/include/darwin/hax_mac.h#L200
2. `vmx_error_t` => `vmx_result_t` (I'm actually very surprised that VC++ didn't catch this--we should check our compiler settings)
   * Multiple occurrences in `core/include/cpu.h` and `core/cpu.c`
3. `%lx` => `%x` (because `vmx_result_t` is essentially an `int`):
   * https://github.com/AlexAltea/haxm/blob/10597c3a90c0d41b26d23bcd17a69c83633c7d9e/core/cpu.c#L453
   * https://github.com/AlexAltea/haxm/blob/10597c3a90c0d41b26d23bcd17a69c83633c7d9e/core/cpu.c#L553
   * https://github.com/AlexAltea/haxm/blob/10597c3a90c0d41b26d23bcd17a69c83633c7d9e/core/cpu.c#L561
   * https://github.com/AlexAltea/haxm/blob/10597c3a90c0d41b26d23bcd17a69c83633c7d9e/core/cpu.c#L699

In addition, my colleague @wcwang ran a few tests on Windows, and some of them resulted in a host freeze. He can tell you more about that.

wcwang:
When I launched Debian in QEMU, which guest image is _debian_squeeze_i386_desktop.qcow2_, the host would be frozen once the launch command was entered. No QEMU window popped up or logs output in command prompt. Similar results occurred on Android virtual device launch. You may try to reproduce this issue on Windows.

darwin/hax_driver/com_intel_hax/intelhaxm.xcodeproj/project.pbxproj
```diff
@@ -35,9 +35,6 @@
 		43C9A9E7138DDA93000A1071 /* hax_host.h in Headers */ = {isa = PBXBuildFile; fileRef = 43C9A9E6138DDA93000A1071 /* hax_host.h */; };
 		43F857E013931E75008A93D6 /* com_intel_hax_mem.h in Headers */ = {isa = PBXBuildFile; fileRef = 43F857DE13931E75008A93D6 /* com_intel_hax_mem.h */; };
 		43F857E113931E75008A93D6 /* com_intel_hax_mem.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 43F857DF13931E75008A93D6 /* com_intel_hax_mem.cpp */; };
-		4BCC4E0513FB6729005E4BE4 /* ia32.c in Sources */ = {isa = PBXBuildFile; fileRef = 4BCC4E0213FB6729005E4BE4 /* ia32.c */; };
```
raphaelning: `ia32.asm` and `vmx_ops.asm` haven't been added to the Xcode project. We're trying to do that and see if we can reproduce the host freeze issue on Mac too.

raphaelning:
I was able to fix the compiler errors and add the two new `.asm` files to Xcode. Here are the patches: 
[haxm-PR64-mac-build-fix.zip](https://github.com/intel/haxm/files/2175734/haxm-PR64-mac-build-fix.zip)

Still, when loading `intelhaxm.kext` I get:

```
(kernel) kxld[com.intel.kext.intelhaxm]:        __vmx_vmwrite
(kernel) kxld[com.intel.kext.intelhaxm]:        __vmx_vmwrite_64
(kernel) kxld[com.intel.kext.intelhaxm]:        __vmx_vmwrite_natural
(kernel) kxld[com.intel.kext.intelhaxm]:        _btr
(kernel) kxld[com.intel.kext.intelhaxm]:        _fxrstor
(kernel) kxld[com.intel.kext.intelhaxm]:        _fxsave
(kernel) kxld[com.intel.kext.intelhaxm]:        _ia32_rdmsr
(kernel) kxld[com.intel.kext.intelhaxm]:        _ia32_wrmsr
(kernel) kxld[com.intel.kext.intelhaxm]:        _rdtsc
(kernel) kxld[com.intel.kext.intelhaxm]:        _vmx_vmread
(kernel) kxld[com.intel.kext.intelhaxm]:        _vmx_vmread_64
(kernel) kxld[com.intel.kext.intelhaxm]:        _vmx_vmread_natural

```

I looked into a few of them, and they are actually C wrappers around `asm_xxx`, but only exist for Windows (`windows/{i386, amd64}/wrapper.c`). We could add a `asm/wrapper.c` for Mac too, but I guess it would be better to move these wrappers to `core/` so as to further reduce code duplication?

AlexAltea:
Thanks for the report, testing and for the macOS-specific patch. I'll update my PR, and try to find and fix the bugs found by @wcwang.

> I'm actually very surprised that VC++ didn't catch this--we should check our compiler settings

Seems like a compiler bug in MSVC, see minimal example at: https://godbolt.org/g/KDjcGJ. Other compilers (gcc/clang/icc) either throw warnings/errors for this. Even MSVC throws an error when building as C++. I've reported the issue, but let's see if they even bother with it, as they tend to heavily prioritize C++ over C.

> I guess it would be better to move these wrappers to `core/` so as to further reduce code duplication?

Will do so.

raphaelning:
> Seems like a compiler bug in MSVC, [...] I've reported the issue, but let's see if they even bother with it, as they tend to heavily prioritize C++ over C.

Thanks for the information! Before they fix this bug, we'll just continue to rely on Clang to detect coding errors.

Regarding the host freeze issue, I wonder how you plan to debug it and narrow it down. We'd like to help, but all we are sure about now is that the bug lies in the third/last commit, which is very big. To make debugging easier, could you consider splitting it further, e.g.:

1. Replacing the old `vmcs.{asm, c}` with `vmx_ops.asm`, `vmx_error_t` with `vmx_result_t`.
   * With some more effort, this one could be split into 2-3 commits, beginning with the introduction of `vmx_result_t` and its integration with the old assembly code, before plugging in `vmx_ops.asm`.
2. Replacing the old `ia32.{asm, c}`, `segments.{asm, c}` with `ia32.asm`, moving `wrapper.c` into `core/`.

Or if you have a better idea, let us know :-)

raphaelning:
BTW:

1. You can reproduce the host freeze bug on Windows regardless of the guest OS used, or even with no guest OS (i.e. the boot ROM test): `qemu-system-x86_64 -accel hax`
2. My colleague @junxiaoc managed to get rid of the host freeze issue by restoring the old `vmcs.asm`, but the guest still wouldn't boot up. This means we probably have bugs in both `vmx_ops.asm` and the new `ia32.asm`. If they are in separate patches, we can deal with them one at a time.

junxiaoc:
We just tried to only pick up ia32.asm(original ia32.asm and segments.asm) related change, qemu could boot up with guest OS.
So freeze issue might be related with vmx_ops or vmx_result_t related changes.

core/vmx_ops.asm
```diff
@@ -0,0 +1,362 @@
...
+    ._r10    resq 1
+    ._r11    resq 1
+    ._r12    resq 1
+    ._r13    resq 1
+    ._r14    resq 1
+    ._r15    resq 1
+endstruc
+
+%macro vmx_check 0
+    pushf
```
raphaelning: `pushf => pushfw`

This turns out to be the root cause of the host freeze issue! Although the `pushf` mnemonic is used by the Intel SDM for the 16-bit (operand size) variant of the instruction, it is still ambiguous, and in this case gets assembled into `pushfq` by NASM for win64. `pushfw`, on the other hand, is unambiguous.

core/ia32.asm
```diff
@@ -0,0 +1,346 @@
...
+function get_kernel_gdt, 1
+    sgdt [reg_arg1]
+    ret
+
+function get_kernel_idt, 1
+    sidt [reg_arg1]
+    ret
+
+function get_kernel_rflags, 0
+    pushf
```
raphaelning: `pushf` => `pushfw` (see my other comment)

AlexAltea:
Thanks a lot for tracking the bug down, I've fixed and the last commit and pushed it back.

> We'd like to help, but all we are sure about now is that the bug lies in the third/last commit, which is very big. To make debugging easier, could you consider splitting it further [...]

Sorry I couldn't made it in time, I will be mostly AFK until tomorrow.

core/ia32.asm
```diff
@@ -0,0 +1,346 @@
...
+%ifidn __BITS__, 64
+    mov rcx, reg_arg1
+    rdmsr
+    shl rdx, 32
+    or reg_ret, rdx
+    ret
+%else
+    mov ecx, reg_arg1
+    rdmsr
+    mov [reg_arg2 + qword_struct.lo], eax
```
raphaelning: The Win32 driver causes a BSOD as soon as it's loaded, and this is the culprit: Here, `reg_arg2 == edx`, but EDX was just overwritten by `rdmsr`. So we have to save `reg_arg2` to another register (e.g. EBX, which itself may need to be saved to the stack first) before calling `rdmsr`, and use that register as the base address here.

nit: To keep the code portable (e.g. in case we want to support other calling conventions), we could surround cdecl-specific logic with `%ifidn __CONV__, x32_cdecl` .. `%endif`

core/ia32.asm
```diff
@@ -0,0 +1,346 @@
...
+    mov [reg_arg2 + qword_struct.hi], edx
```
raphaelning: Same here.

core/ia32.asm
raphaelning: > we could surround cdecl-specific logic with %ifidn __CONV__, x32_cdecl .. %endif

Or perhaps `%ifidn reg_arg2, edx` .. `%endif`. Anyway, it's not easy to achieve portability across different calling conventions, since we probably have cdecl-specific logic elsewhere in the code. So it's perfectly fine if you just explain in a comment why we must save `reg_arg2`.

BTW, I've noticed that line 234, which is equivalent to `mov ecx, ecx`, isn't optimized out by NASM, even for the Release build, and even if I manually run `nasm` with `-Ox`. This isn't a big issue at all, but do you think NASM should be smarter in recognizing this no-op? Or maybe it doesn't want to be too smart, because we could just want to waste some CPU cycles?

AlexAltea: > but do you think NASM should be smarter in recognizing this no-op? Or maybe it doesn't want to be too smart, because we could just want to waste some CPU cycles?

I believe that's out-of-scope for NASM: it assembles the source as-is and only optimizes code when the instruction mnemonics are ambiguous and allow different variants. E.g. it will prioritize *imm8* over *imm32* when the constants is in range *[0x00, 0xFF]*. There's more info at:
https://www.nasm.us/doc/nasmdoc2.html#section-2.1.23

As you suggested, the only real solution is writing calling-convention specific implementations of functions that have side-effects on "specific" GPRs (just `asm_rdmsr` and `asm_rdmsr`). There might be more such pseudo-nop's out there, but I don't see them as a big issue either. Fortunately, most functions are simple enough that can be implemented just with the `reg_*` macros.

raphaelning: I see, thanks! BTW, with `asm_rdmsr` fixed, the new code seems to work well on Win32, so I think we are very close to getting this PR merged!

core/ia32.c
```diff
@@ -0,0 +1,196 @@
...
+
+void bts(uint8 *addr, uint bit)
+{
+    uint8 *base = addr + bit / 8;
+    uint offset = bit % 8;
+    asm_bts(base, offset);
+}
+
+void _vmx_vmwrite(struct vcpu_t *vcpu, const char *name,
+    component_index_t component,
```
raphaelning: nit: Insert more spaces to align with the first parameter (`struct vcpu_t *`).

core/ia32.c
```diff
@@ -0,0 +1,196 @@
...
+    mword source_val)
```
raphaelning: nit: Same here.

core/ia32.c
```diff
@@ -0,0 +1,196 @@
...
+{
+    asm_vmwrite(component, source_val);
+}
+
+void _vmx_vmwrite_64(struct vcpu_t *vcpu, const char *name,
+    component_index_t component,
```
raphaelning: nit: Align with the first parameter.

core/ia32.c
```diff
@@ -0,0 +1,196 @@
...
+    uint64 source_val)
```
raphaelning: nit: Same here.

core/ia32.c
```diff
@@ -0,0 +1,196 @@
...
+#ifdef _M_IX86
+    asm_vmwrite(component, (uint32)source_val);
+    asm_vmwrite(component + 1, (uint32)(source_val >> 32));
+#else
+    asm_vmwrite(component, source_val);
+#endif
+}
+
+void _vmx_vmwrite_natural(struct vcpu_t *vcpu, const char *name,
+    component_index_t component,
```
raphaelning: nit: Align with the first parameter.

core/ia32.c
```diff
@@ -0,0 +1,196 @@
...
+    uint64 source_val)
```
raphaelning: nit: Same here.

core/include/vmx.h
```diff
@@ -484,10 +487,26 @@ typedef enum encode_t encode_t;
...
+uint64 ASMCALL asm_vmxrun(struct vcpu_state_t *state, uint16 launch);
+
+mword ASMCALL vmx_get_rip(void);
+
+uint64 vmx_vmread(struct vcpu_t *vcpu, component_index_t component);
+uint64 vmx_vmread_natural(struct vcpu_t *vcpu, component_index_t component);
+uint64 vmx_vmread_64(struct vcpu_t *vcpu, component_index_t component);
+
+void _vmx_vmwrite(struct vcpu_t *vcpu, const char *name,
+    component_index_t component, mword source_val);
```
raphaelning: nit: Insert more spaces to align with the first parameter (`struct vcpu_t *`).

core/include/vmx.h
```diff
@@ -484,10 +487,26 @@ typedef enum encode_t encode_t;
...
+void _vmx_vmwrite_natural(struct vcpu_t *vcpu, const char *name,
+    component_index_t component, uint64 source_val);
```
raphaelning: nit: Same here.

core/include/vmx.h
```diff
@@ -484,10 +487,26 @@ typedef enum encode_t encode_t;
...
+void _vmx_vmwrite_64(struct vcpu_t *vcpu, const char *name,
+    component_index_t component, uint64 source_val);
```
raphaelning: nit: Same here.

core/ia32.c
```diff
@@ -0,0 +1,196 @@
...
+extern void asm_btr(uint8 *addr, uint bit);
+extern void asm_bts(uint8 *addr, uint bit);
+
+extern void asm_fxinit(void);
+extern void asm_fxsave(mword *addr);
+extern void asm_fxrstor(mword *addr);
+
+
+#ifdef _M_IX86
+extern uint64 ASMCALL asm_vmread(uint32 component);
```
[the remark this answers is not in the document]
raphaelning: BTW, the 64-bit version of `asm_vmread` should also return `mword` instead of `uint32`. According to @junxiaoc , this change alone is enough to make this patch work on Windows.

AlexAltea:
All fixed.

> I'd actually prefer calling the macOS kernel function, even though it would create platform-dependent code paths

I fully agree with your reasoning. I've reverted that change and used `#ifndef __MACH__` to remove the generic assembly-based function from macOS.

raphaelning:
Thanks! The new code looks all good, and works well on Windows. However, most Mac tests are broken, resulting in a host kernel panic. We're still trying to root cause the issue. For now, let me post the macOS kernel panic logs, in case you can spot an obvious problem.

So far we've seen two kinds of macOS kernel panic:

a)
```
*** Panic Report ***
panic(cpu 3 caller 0xffffff8008987d3d): Kernel trap at 0x0000000000000189, type 14=page fault, registers:
CR0: 0x000000008001003b, CR2: 0x0000000000000189, CR3: 0x0000000040a1d055, CR4: 0x00000000003627e0
RAX: 0x0000000000000000, RBX: 0xffffff8044613800, RCX: 0x0000000000000000, RDX: 0x0000000000000663
RSP: 0xffffff913b4839b0, RBP: 0x0000000000000000, RSI: 0x0000000000000018, RDI: 0x0000000000000000
R8:  0x0000000000000000, R9:  0x0000000000000000, R10: 0x0000000000000000, R11: 0x0000000000000000
R12: 0x000000002c100801, R13: 0xffffff913b483a40, R14: 0xffffff802bdd9c00, R15: 0x0000000000000000
RFL: 0x0000000000010086, RIP: 0x0000000000000189, CS:  0x0000000000000008, SS:  0x0000000000000010
Fault CR2: 0x0000000000000189, Error code: 0x0000000000000010, Fault CPU: 0x3, PL: 0, VF: 0

Backtrace (CPU 3), Frame : Return Address
0xffffff913b483460 : 0xffffff800886d166 
0xffffff913b4834b0 : 0xffffff8008995824 
0xffffff913b4834f0 : 0xffffff8008987b10 
0xffffff913b483570 : 0xffffff800881f180 
0xffffff913b483590 : 0xffffff800886cbdc 
0xffffff913b4836c0 : 0xffffff800886c99c 
0xffffff913b483720 : 0xffffff8008987d3d 
0xffffff913b4838a0 : 0xffffff800881f180 
0xffffff913b4838c0 : 0x189 

BSD process name corresponding to current thread: qemu-system-x86_

Mac OS version:
17F77

Kernel version:
Darwin Kernel Version 17.6.0: Tue May  8 15:22:16 PDT 2018; root:xnu-4570.61.1~1/RELEASE_X86_64
Kernel UUID: CC1699EE-71B5-3AF1-AAA3-6DBD16222B92
Kernel slide:     0x0000000008600000
Kernel text base: 0xffffff8008800000
__HIB  text base: 0xffffff8008700000
System model name: MacBookPro14,2 (Mac-CAD6701F7CEA0921)

System uptime in nanoseconds: 35247773639496
last loaded kext at 35152996138307: com.intel.kext.intelhaxm	1 (addr 0xffffff7f8e030000, size 229376)
last unloaded kext at 35152574121238: com.intel.kext.intelhaxm	7.2.0 (addr 0xffffff7f8e030000, size 126976)
```

b)
```
*** Panic Report ***
mp_kdp_enter(): 2, 3, 4 TIMED-OUT WAITING FOR NMI-ACK, PROCEEDING
panic(cpu 2 caller 0xffffff801fb87d3d): Kernel trap at 0xffffff801faba11c, type 14=page fault, registers:
CR0: 0x000000008001003b, CR2: 0x0000000000000016, CR3: 0x00000001b34d410b, CR4: 0x00000000003627e0
RAX: 0x0000000000000000, RBX: 0xffffff802029c670, RCX: 0xffffff9156a83900, RDX: 0x0000000000000004
RSP: 0xffffff9156a83820, RBP: 0xffffff9156a838f0, RSI: 0x0000000000000004, RDI: 0xffffff802029c670
R8:  0x0000000000000ff4, R9:  0x0000000000000000, R10: 0xffffc90000022000, R11: 0xffffffff81141e00
R12: 0x0000000000000000, R13: 0x0000000000000000, R14: 0x0000000000000004, R15: 0x0000000000000000
RFL: 0x0000000000010046, RIP: 0xffffff801faba11c, CS:  0x0000000000000008, SS:  0x0000000000000010
Fault CR2: 0x0000000000000016, Error code: 0x0000000000000000, Fault CPU: 0x2, PL: 0, VF: 0

Backtrace (CPU 2), Frame : Return Address
0xffffff9156a832d0 : 0xffffff801fa6d166 
0xffffff9156a83320 : 0xffffff801fb95824 
0xffffff9156a83360 : 0xffffff801fb87b10 
0xffffff9156a833e0 : 0xffffff801fa1f180 
0xffffff9156a83400 : 0xffffff801fa6cbdc 
0xffffff9156a83530 : 0xffffff801fa6c99c 
0xffffff9156a83590 : 0xffffff801fb87d3d 
0xffffff9156a83710 : 0xffffff801fa1f180 
0xffffff9156a83730 : 0xffffff801faba11c 
0xffffff9156a838f0 : 0xffffff801fa78e9a 
0xffffff9156a83940 : 0xffffff802005e456 
0xffffff9156a83960 : 0xffffff7fa55407ba 
No mapping exists for frame pointer
Backtrace terminated-invalid frame pointer 0xffff88005fc03f98
      Kernel Extensions in backtrace:
         com.intel.kext.intelhaxm(1.0)[C0020148-1BDC-365A-8793-D893A66DA7CB]@0xffffff7fa553c000->0xffffff7fa5573fff

BSD process name corresponding to current thread: qemu-system-x86_

Mac OS version:
17F77

Kernel version:
Darwin Kernel Version 17.6.0: Tue May  8 15:22:16 PDT 2018; root:xnu-4570.61.1~1/RELEASE_X86_64
Kernel UUID: CC1699EE-71B5-3AF1-AAA3-6DBD16222B92
Kernel slide:     0x000000001f800000
Kernel text base: 0xffffff801fa00000
__HIB  text base: 0xffffff801f900000
System model name: MacBookPro14,2 (Mac-CAD6701F7CEA0921)

System uptime in nanoseconds: 262242529783
last loaded kext at 253616550554: com.intel.kext.intelhaxm	1 (addr 0xffffff7fa553c000, size 229376)
last unloaded kext at 253108159779: com.intel.kext.intelhaxm	7.2.0 (addr 0xffffff7fa553c000, size 126976)
```

raphaelning:
The manual address/offset patching method works, so this `vmx_get_rip` issue is probably the last bug we need to fix :-)

I inspected `vmx_ops.o` with `otool` and `nm`, but did not spot anything abnormal...

If we compare the original (`vmcs.{asm, c}`) and current implementations of this assembly routine, there are two main differences:

i. The old code uses the `lea` instruction instead of `mov`.
ii. The old code declares `exit_entry` as a global label (MASM `EXIT_ENTRY::` rather than `EXIT_ENTRY:`, GAS `EXIT_ENTRY:` rather than something like `1b:`).

Changing `exit_entry` to a global label doesn't seem to help. I'm now trying to figure out how to make `lea` work...

core/vmx_ops.asm
```diff
@@ -0,0 +1,364 @@
...
+    pop ecx
+    vmx_check
+    popfd
+    ret
+%else
+    %error "Unimplemented function"
+%endif
+
+function vmx_get_rip, 0
+    mov reg_ret, asm_vmxrun.exit_entry
```
raphaelning: => `lea reg_ret, [rel asm_vmxrun.exit_entry]`

This is enough to fix the macOS kernel panic! (A lot of credit goes to @junxiaoc )

Basically, the use of `lea` with the [REL keyword](https://www.nasm.us/doc/nasmdoc6.html#section-6.2.1) causes the address to be computed using RIP-relative addressing. The disassembly looks like this:

```asm
0000000000004a62        48 8d 05 7a ff ff ff    leaq    asm_vmxrun.exit_entry(%rip), %rax
```

Although `rel` is specific to 64-bit mode, it seems to be just ignored for the Win32 build, and this `lea` approach works across all platforms we support. Nevertheless, for readability (especially since `mov` is more intuitive than `lea`), we could save the `mov` approach for 32-bit.

raphaelning: Please feel free to incorporate
[haxm-PR64-mac-kernel-panic-fix.zip](https://github.com/intel/haxm/files/2201035/haxm-PR64-mac-kernel-panic-fix.zip) into your 3rd commit.

AlexAltea: @raphaelning @junxiaoc Thank you so much for debugging+fixing the issue, and also for your patience with this PR. :-)

I'll add the patch and squash everything into the 3rd commit (it will be big, but as mentioned earlier it's convenient to avoid breaking git-bisect).

AlexAltea:
All done!

PR #67, opened 2018-06-21
Fixed mov/movzx instruction emulation

AlexAltea:
This patch fixes two issues:
- Fixed `mov` (and potentially other instructions) due to top 32 bits of general purpose registers not being cleared while writing to the lower 32 bits.
- Fixed `movzx` bug caused by mismatching operand sizes.

Additionally, this patch adds:
- Support for `movsx` instructions.
- Tests for both `movzx` and `movsx`.

### DIFF
--- a/core/emulate.c
+++ b/core/emulate.c
@@ -92,6 +92,8 @@
 DECL_DECODER(op_none);
 DECL_DECODER(op_modrm_reg);
 DECL_DECODER(op_modrm_rm);
+DECL_DECODER(op_modrm_rm8);
+DECL_DECODER(op_modrm_rm16);
 DECL_DECODER(op_vex_reg);
 DECL_DECODER(op_moffs);
 DECL_DECODER(op_simm);
@@ -142,6 +144,7 @@ static void em_andn_soft(struct em_context_t *ctxt);
 static void em_bextr_soft(struct em_context_t *ctxt);
 static void em_mov(struct em_context_t *ctxt);
 static void em_movzx(struct em_context_t *ctxt);
+static void em_movsx(struct em_context_t *ctxt);
 static void em_xchg(struct em_context_t *ctxt);
 
 static const struct em_opcode_t opcode_group1[8] = {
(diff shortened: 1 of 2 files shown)


core/emulate.c
```diff
@@ -823,9 +850,19 @@ static void em_mov(struct em_context_t *ctxt)
...
-    memcpy(&value, &ctxt->src1, ctxt->operand_size);
-    ctxt->dst.value = value;
+    ctxt->dst.value = 0;
+    memcpy(&ctxt->dst.value, &ctxt->src1.value, ctxt->src1.size);
+}
+
+static void em_movsx(struct em_context_t *ctxt)
+{
+    uint64_t extension = 0;
+    if (ctxt->src1.value & (1 << ((8 * ctxt->src1.size) - 1))) {
```
raphaelning: The `<<` operation may cause an int overflow when `ctxt->src1.size == 8`. Two possible ways to fix that:

1. `1 <<` => `1ULL <<`
1. `if (ctxt->src1.value >> (8 * ctxt->src1.size - 1))`

AlexAltea: Sorry I overlooked this edge case, `movsx` was defined so that `ctxt->src1.size` can only be *1* or *2*.
This is fixed now. I have picked your first suggestion (`1 <<` => `1ULL <<`), as it will work regardless of whether the higher bits (above the sign bit) are zero.

AlexAltea:
> Just curious, is this behavior specified by Intel SDM?

Yes, this is specified in *Intel SDM Vol. 1: 3.4.1.1 General-Purpose Registers in 64-Bit Mode*:

> 32-bit operands generate a 32-bit result, zero-extended to a 64-bit result in the destination general-purpose register

core/emulate.c
raphaelning: Ah, right, it's the source operand. Sorry about the fuss!

raphaelning:
Thanks, that's very good to know! We'll test the patch and merge it soon.

PR #70, opened 2018-06-29
darwin: Move preemption helpers out of vmcs.c

raphaelning:
hax_disable_preemption() and hax_enable_preemption() do not belong
to vmcs.c, since they do not use any VMX instruction. Move them to
hax_wrapper.cpp, in the same way that their  Windows counterparts
are defined in windows/hax_wrapper.c.

### DIFF
--- a/darwin/hax_driver/com_intel_hax/asm/vmcs.c
+++ b/darwin/hax_driver/com_intel_hax/asm/vmcs.c
@@ -28,37 +28,9 @@
  * POSSIBILITY OF SUCH DAMAGE.
  */
 
-#include "../../../../core/include/ia32.h"
 #include "../../../../core/include/vmx.h"
 #include "../../../../core/include/types.h"
-#include "../../../../include/hax.h"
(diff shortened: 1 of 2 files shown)


raphaelning:
> Change looks good. Only a small question is why it needs extern "C"? My impression whole project is in C.

Actually, we have quite a few .cpp files in `darwin/`, and this is one of them. Without `extern "C"`, I got an error when loading `intelhaxm.kext`:

```
(kernel) kxld[com.intel.kext.intelhaxm]:        _hax_disable_preemption
(kernel) kxld[com.intel.kext.intelhaxm]:        _hax_enable_preemption
```

which is not really surprising, since these functions are called from C code. Also, if you look at how other similar functions are defined in `hax_wrapper.cpp`, e.g. `hax_enable_irq()`, you'll also see the `extern "C"` modifier.

junxiaoc:
Yes, just noticed there are cpp files in darwin. No further comment from me.

PR #71, opened 2018-07-03
tests: Fix linker warning

raphaelning:
When building the tests project with MSVC, we always get a warning:

 LNK4098: defaultlib 'LIBCMT' conflicts with use of other libs

Since the project uses the /MDd (MultiThreadedDebugDLL) and /MD
(MultiThreadedDLL) compiler flags for Debug and Release
configurations, respectively, it should not link against LIBCMT
(which corresponds to /MT) or any other incompatible C runtime
library (CRT). According to:

 https://docs.microsoft.com/en-us/cpp/error-messages/tool-errors/linker-tools-warning-lnk4098
 https://docs.microsoft.com/en-us/cpp/build/reference/nodefaultlib-ignore-libraries

we just need to instruct the linker to explicitly ignore each
incompatible CRT.

### DIFF
--- a/tests/tests.vcxproj
+++ b/tests/tests.vcxproj
@@ -64,6 +64,7 @@
     <Link>
       <GenerateDebugInformation>true</GenerateDebugInformation>
       <SubSystem>Console</SubSystem>
+      <IgnoreSpecificDefaultLibraries>libcmt.lib</IgnoreSpecificDefaultLibraries>
     </Link>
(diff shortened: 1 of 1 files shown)


raphaelning:
> maybe we could only add libcmt.lib in ignored libs?

Yes, that's a great idea :-) I just tried removing all other libs except `libcmt.lib`, and the patch still works.

PR #85, opened 2018-07-31
core: save/restore FPU registers in VM entry/exit

### LINKED ISSUE #39: QEMU/haxm: Multiple segmentation faults on Ubuntu 18.04
While installation of a Ubuntu 16.04 works fine with qemu 2.11 and haxm 7.1, 18.04 produces multiple user mode segmentation faults in the guest on early steps and eventually hangs. The netboot images can be obtained from http://de.archive.ubuntu.com/ubuntu/dists/artful/main/installer-amd64/current/images/netboot/netboot.tar.gz . 

Parameters:
`-m 1024 -smp 2 -rtc base=utc -drive file=temp.vmdk,if=virtio,detect-zeroes=unmap,discard=unmap -drive file=fat:rw:fat-type=32:label=kernel:C:\temp,format=raw,if=virtio -vga std -device virtio-net,netdev=natted -netdev user,id=natted -kernel linux_initrd/linux -initrd linux_initrd/initrd.gz -append "nofb fb=false pti=off interface=auto auto=true" -nodefaults`

### LINKED ISSUE #74: Cannot start ubuntu, archlinux ... with hax
I have qemu windows build 2.12.90, haxm 7.2.0. Ubuntu, nor arch linux does not works when i turn on hax acceleration. Permanent kernel panics, black screen freezing and other crashes happens when i run qemu.
Qemu crashed with hax - when i ran it from iso. It crashed on already installed system - it's not matters.

Versions:
archlinux-2018.07.01-x86_64
ubuntu-18.04-live-server-amd64.iso

I run qemu-system-x86_64.exe binary.

`qemu-system-x86_64.exe -drive file=build_slave.img,index=0,media=disk,format=qcow2 -m 2G -accel hax -name build_slave`

My CPU:
core i7 2600k

My Host:
Windows 7 64bit

junxiaoc:
Guest OS kernel/app might use SSE instruction and registers.
When Guest OS VM exits, these registers should be saved, or else
it might be corrupted by host OS/app. In next time guest VM
enter, guest's SSE registers context might be corrupted.

Guest app segfault and kernel panic were reported
which should be related with this issue.

This change is to remove is_fpu_used flag so guest FPU registers
could be saved in VM exit and restored in VM entry unconditionally.

Fixes #39, fixes #74.

### DIFF
--- a/core/vcpu.c
+++ b/core/vcpu.c
@@ -1198,7 +1198,7 @@ static void fill_common_vmcs(struct vcpu_t *vcpu)
         }
     }
 
-    exc_bitmap = (1u << VECTOR_MC) | (1u << VECTOR_NM);
+    exc_bitmap = 1u << VECTOR_MC;
 
 #ifdef __x86_64__
     exit_ctls = EXIT_CONTROL_HOST_ADDR_SPACE_SIZE | EXIT_CONTROL_LOAD_EFER |
@@ -1895,10 +1895,8 @@ static void vcpu_enter_fpu_state(struct vcpu_t *vcpu)
     struct fx_layout *hfx = (struct fx_layout *)hax_page_va(hstate->hfxpage);
     struct fx_layout *gfx = (struct fx_layout *)hax_page_va(gstate->gfxpage);
 
-    if (vcpu->is_fpu_used) {
-        fxsave((mword *)hfx);
-        fxrstor((mword *)gfx);
-    }
+    fxsave((mword *)hfx);
+    fxrstor((mword *)gfx);
 }
 
 static void vcpu_exit_fpu_state(struct vcpu_t *vcpu)
@@ -1908,10 +1906,8 @@ static void vcpu_exit_fpu_state(struct vcpu_t *vcpu)
     struct fx_layout *hfx = (struct fx_layout *)hax_page_va(hstate->hfxpage);
     struct fx_layout *gfx = (struct fx_layout *)hax_page_va(gstate->gfxpage);
 
-    if (vcpu->is_fpu_used) {
-        fxsave((mword *)gfx);
-        fxrstor((mword *)hfx);
-    }
+    fxsave((mword *)gfx);
+    fxrstor((mword *)hfx);
 }
 
 // Instructions are never longer than 15 bytes:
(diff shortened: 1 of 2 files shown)


core/vcpu.c
```diff
@@ -1895,10 +1895,8 @@ static void vcpu_enter_fpu_state(struct vcpu_t *vcpu)
     struct fx_layout *hfx = (struct fx_layout *)hax_page_va(hstate->hfxpage);
     struct fx_layout *gfx = (struct fx_layout *)hax_page_va(gstate->gfxpage);
 
-    if (vcpu->is_fpu_used) {
-        fxsave((mword *)hfx);
-        fxrstor((mword *)gfx);
-    }
+    fxsave((mword *)hfx);
```
raphaelning: To minimize the performance penalty that the unconditional FXSAVE/FXRSTOR incurs, we should replace them with XSAVEOPT/XRSTOR if the host supports them. According to the description of XSAVEOPT in Intel SDM Vol. 2c, it is pretty smart in avoiding unnecessary memory writes. In addition, only the XSAVE instructions can save/restore AVX context.

Of course, this optimization can and should be implemented in a future patch.

wcwang:
A new warning is generated.
core/vcpu.c Ln 2168: unused variable 'cr0'

core/vcpu.c
raphaelning: As to the question of whether we should utilize CR0.TS and \#NM (Device Not Found exception) to implement lazy FPU (including SSE, etc.) context switching, here's the official recommendation from Intel SDM Vol. 3A 13.4 (_Designing OS Facilities for Saving x87 FPU, SSE and Extended States on Task or Context Switches_). Basically the answer is no:

> * The operating system can take the responsibility for saving the states as part of the task switch process, but delay the restoring of the states until an instruction operating on the states is actually executed by the new task. See Section 13.4.1, “Using the TS Flag to Control the Saving of the x87 FPU and SSE State,” for more information. This approach is called lazy restore.
>
>   The use of lazy restore mechanism in context switches is not recommended when XSAVE feature set is used to save/restore states for the following reasons.
>
>   — With XSAVE feature set, Intel processors have optimizations in place to avoid saving the state components that are in their initial configurations or when they have not been modified since they were restored last. These optimizations eliminate the need for lazy restore. See section 13.5.4 in Intel® 64 and IA-32 Architectures Software Developer’s Manual, Volume 1.
>   — Intel processors have power optimizations when state components are in their initial configurations. Use of lazy restore retains the non-initial configuration of the last thread and is not power efficient.
>   — Not all extended states support lazy restore mechanisms. As such, when one or more such states are enabled it becomes very inefficient to use lazy restore as it results in two separate state restore, one in context switch for the states that does not support lazy restore and one in the #NM handler for states that support lazy restore.

PR #87, opened 2018-08-08
darwin: fix Debug build and install issue

junxiaoc:
In Mac, when install Debug version haxm driver, it reports several
function symbols missing. They are inline function.

Mac Debug build uses "-O0" for build optimazation. With this flag,
some inline functions which don't have static prefix have build
issue, they don't perform inline expansion.

This change is to add static prefix to several inline functions.

Signed-off-by: Junxiao Chang <junxiao.chang@intel.com>


junxiaoc:
Sign off message added.

PR #91, opened 2018-08-31
Refine is_fpu_used logic to handle eager fpu.

Taogle2018:
Hi, Ning,

This refines detection of guest fpu usage. Old haxm seems to rely on guest setting CR0.TS and trapping exception 7.
I know there is a patch for forcefully turning on fpu save/restore. I do not have any preference to either solution. I just uploaded mine in case you want a friendly version towards non-fpu usage guests.

Haitao

### DIFF
--- a/core/include/vcpu.h
+++ b/core/include/vcpu.h
@@ -178,6 +178,7 @@ struct vcpu_t {
         uint64 paused                          : 1;
         uint64 panicked                        : 1;
         uint64 is_running                      : 1;
+        uint64 is_fpu_used                     : 1;
         uint64 is_vmcs_loaded                  : 1;
         uint64 event_injected                  : 1;
         /* vcpu->state is valid or not */
@@ -189,7 +190,8 @@ struct vcpu_t {
         uint64 vmcs_pending_entry_instr_length : 1;
         uint64 vmcs_pending_entry_intr_info    : 1;
         uint64 vmcs_pending_guest_cr3          : 1;
-        uint64 padding                         : 53;
+        uint64 is_ts_set_by_haxm               : 1;
+        uint64 padding                         : 51;
     };
 
     /* For TSC offseting feature*/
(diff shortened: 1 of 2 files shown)


raphaelning:
This is indeed a nicely crafted patch that honors the original HAXM design and fixes `vcpu->is_fpu_used`. However, we still prefer the existing solution (#85), which is to unconditionally save/restore FPU state at every VM exit/entry, after considering both code complexity and performance.

1. The unconditional approach is definitely simpler :)
1. The unconditional approach does not depend on guest behavior, and therefore is compatible with all guest OSes.
1. Performance-wise, the unconditional approach should be more or less the same as the conditional approach:
   i. Note that once `is_fpu_used` is set, it's never reset. Therefore, for a guest that does use FPU/MMX/SSE instructions, after the first use of such an instruction is detected by HAXM (`is_fpu_used = 1`), which usually happens very early during the lifetime of guest, both approaches will behave like the unconditional approach, and will have the same FXSAVE/FXRSTOR overhead.
   ii. While reviewing PR #85, I actually thought about minimizing FXSAVE/FXRSTOR calls by only saving the guest FPU state when it's dirty. That would require properly clearing `is_fpu_used` and trapping \#NM / CR0.TS = 1 for every "session" of guest execution (i.e. between every VM entry and exit), essentially implementing something similar to Linux's lazy FPU mode. However, Linux itself has dropped lazy FPU mode in favor of eager FPU mode, mainly because the former doesn't really offer better performance, especially on recent hardware. See commit messages [here](https://lore.kernel.org/patchwork/patch/322488/) and [here](https://lore.kernel.org/patchwork/patch/638220/). Therefore, we thought we should also take the eager/unconditional approach.

Taogle2018:
I agree and I prefer the simple fix too. I believe it is not a perf issue on modern CPUs for eager save/restore. Going further to runtime "eager save, lazy restore" is far too complex. Some time being simple is the king, even by paying a little bit for that.
However, I think I have shown where HAXM has done wrong by this fix. Be it simple or complex, hypervisor should not make assumptions on guests behavior. It is not about which way to go. It is about doing things right. :)

raphaelning:
> I think I have shown where HAXM has done wrong by this fix. Be it simple or complex, hypervisor should not make assumptions on guests behavior. It is not about which way to go. It is about doing things right. :)

True :) Closing this PR now.

PR #94, opened 2018-09-06
HaxmLoader cleanup and fix

raphaelning:


HaxmLoader/haxm_loader.c
```diff
@@ -166,74 +174,38 @@ int DriverInstall(char *sysFileName)
...
                 if (argc == 2)
                     return DriverUninstall();
                 break;
-            case 's':
-                if (argc == 4)
-                    return WriteToReg(atoi(argv[2]),
-                                      atoi(argv[3]));
-                break;
             default:
                 break;
```
junxiaoc: Adding a PrintUsage here?

HaxmLoader/haxm_loader.c
```diff
@@ -100,55 +99,64 @@ int DriverUninstall(void)
...
+    if (StringCchCatW(driverLocation, MAX_PATH, L"\\") != S_OK ||
+        StringCchCatW(driverLocation, MAX_PATH, sysFileName) != S_OK) {
+        printf("%s(): Failed to get absolute path:\r\n",
+                __FUNCTION__);
+        wprintf(L"  sysFileName='%s'\r\n", sysFileName);
+        goto cleanup;
+    }
 
     // check if sysFile exists
-    fileHandle = CreateFile((LPCSTR)driverLocation,
```
junxiaoc: I'm not very familiar with unicode coding, my impression is there is a macro "_T" which could handle this as well?

HaxmLoader/haxm_loader.c
junxiaoc: Yes, in line 244 which is out of this display, PrintUsage is called in "default" case. :)

HaxmLoader/haxm_loader.c
raphaelning: `_T` is probably just another form of the `TEXT("some string")` macro, which actually expands to `L"some string"` when `_UNICODE` is defined:

https://docs.microsoft.com/en-us/windows/desktop/learnwin32/working-with-strings

The same MSDN article also deprecates the use of `TEXT`:

> The **TEXT** and **TCHAR** macros are less useful today, because all applications should use Unicode.

That's why I'm using Windows APIs with the `W` suffix explicitly--we don't need to compile the code for non-Unicode versions Windows, because all the Windows versions that HAXM can run on support Unicode. I also tried to follow guidelines from section 10 of this article:

https://utf8everywhere.org/

HaxmLoader/HaxmLoader.vcxproj
```diff
@@ -118,4 +118,4 @@
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
   <ImportGroup Label="ExtensionTargets">
   </ImportGroup>
-</Project>
+</Project>
```
wcwang: There is no end of new line. Is it reasonable?

raphaelning: Not desirable but I don't think it was me... I think I just let Visual Studio upgrade the project.

junxiaoc: I ever met this as well, no change, but it was marked.

PR #95, opened 2018-09-07
Fix TEST instruction emulation

raphaelning:
This PR fixes two issues with the instruction emulator:

1. Many unsupported MMIO instructions (e.g. `TEST r/mN, rN`, `XCHG r/mN, rN`, etc.) lead to a host crash.
2. `TEST r/mN, rN` is not supported.

### DIFF
--- a/core/emulate.c
+++ b/core/emulate.c
@@ -881,7 +881,6 @@ em_status_t EMCALL em_decode_insn(struct em_context_t *ctxt, const uint8_t *insn
     uint8_t b;
     uint64_t flags;
     struct em_opcode_t *opcode;
-    const struct em_opcode_t *opcode_group;
 
     switch (ctxt->mode) {
     case EM_MODE_REAL:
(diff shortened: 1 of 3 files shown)


core/vcpu.c
```diff
@@ -1956,6 +1956,10 @@ static int vcpu_emulate_insn(struct vcpu_t *vcpu)
     uint64 rip = vcpu->state->_rip;
     uint64 va;
 
+    // Clean up the emulation context of the previous MMIO instruction, so that
+    // even if things go wrong, the behavior will still be predictable.
+    vcpu_init_emulator(vcpu);
```
junxiaoc: vcpu_init_emulator only needs to be called in EM_ERROR case?

core/vcpu.c
```diff
@@ -1991,8 +1995,11 @@ static int vcpu_emulate_insn(struct vcpu_t *vcpu)
     em_ctxt->rip = rip;
     rc = em_decode_insn(em_ctxt, instr);
     if (rc < 0) {
-        hax_panic_vcpu(vcpu, "%s: em_decode_insn() failed: vcpu_id=%u",
-                       __func__, vcpu->vcpu_id);
+        hax_panic_vcpu(vcpu, "em_decode_insn() failed: vcpu_id=%u,"
```
junxiaoc: Should we print these info as well?
ctxt->mode, ctxt->override_operand_size

core/vcpu.c
raphaelning: If we failed to decode/emulate the previous instruction, we should have terminated the guest already and wouldn't have arrived here. The intention here is to always reset the emulation context (`vcpu->emulate_ctxt`) before we start working on a new instruction, because `emulate_ctxt` is shared among all MMIO instructions executed by this vCPU, and we don't want to accidentally reuse anything from the previous context.

BTW, I have made some changes to this patch, and will upload the new version later on.

core/vcpu.c
raphaelning: Ah, sorry I missed this comment. I'll instead add a call to `dump_vmcs()`, which will give us (among other info) guest CR0 (`GUEST_CR0` of VMCS) and CS (`GUEST_CS_AR` of VMCS), from which we'll be able to derive `ctxt->mode`, etc.

raphaelning:
@junxiaoc Ah, right, the wording isn't the best. Let's wait for the test result, in case there's also something in the code that needs to be fixed as well.

core/emulate.c
```diff
@@ -1010,22 +1010,30 @@ em_status_t EMCALL em_decode_insn(struct em_context_t *ctxt, const uint8_t *insn
         ctxt->operand_size = 1;
     }
     if (flags & INSN_GROUP) {
-        opcode_group = &opcode->group[ctxt->modrm.opc];
-        opcode->handler = opcode_group->handler;
-        if (!opcode_group->handler) {
+        const struct em_opcode_t *group_opcode;
```
PatrickvL: Moving the variable declaration here isn't necessary, plus it's probably declared above for maximum compatibly across compilers

mborgerson:
Confirmed (a) in my comment above, that the new updated driver no longer causes a kernel panic. Have not explicitly verified correctness of (b) yet, but my target appears to be getting much farther.

core/emulate.c
PatrickvL: C89-compilant compilers and gcc -pendantic would fail on local-scoped variable declarations.

Googling around, most people, like you and me, do seem to prefer local-scoped variable declarations. (Especially when C++ variables are initialized with constructor calls, limiting scope can be advantageous.) However, one remark that did stand out was "[Reusing variable names with a different type is prevented](https://softwareengineering.stackexchange.com/a/119212)". 

Other than that, most compilers will try to limit scope and move variables into registers anyway, so it's a moot point.

raphaelning: Thanks, that's interesting. Although I do have some vague memory of C89, I guess C99 was already prevalent when I learned coding. The compilers we use for HAXM (MSVC 2017 and Clang) are new enough to accept C99 syntax, so I wouldn't worry about compatibility with C89. In fact, we may consider enabling C11 support in these compilers some day.

> However, one remark that did stand out was "Reusing variable names with a different type is prevented"

To achieve that, I'd rather avoid writing long functions, so we could easily catch variable naming conflicts in code review ;-)

raphaelning:
@mborgerson Thanks! If you have time for one more test, could you apply **only** the first patch 40330993d8c2c6738903acc7c2dd924fe6bc31bf and capture HAXM logs? The purpose of this patch is not only to avoid a host kernel panic, but also to log useful information about the unsupported MMIO instruction.

You can run the following command to catch live HAXM output:

`log stream --predicate 'eventMessage contains "hax"' --info`

PR #108, opened 2018-10-16
Added support for Linux hosts

AlexAltea:
## Motivation

Although previously (#4) Linux support for HAXM was dismissed due to KVM being around. There's still reasons for having Linux support in this project:
* **Single cross-platform hypervisor API**: Although QEMU supports most hypervisors, smaller cross-platform projects need to deal with multiple backends: WHPX on Windows, KVM on Linux, HVF on MacOS. HAXM can finally fill that gap and cover the 3 major platforms.
* **Extra features**: Although KVM is a more mature hypervisor, HAXM is **not** a subset of it: As of today, there's certain guests that *only* run on HAXM [[1]](https://github.com/AlexAltea/orbital), due to the recent emulator changes which added support for x86 instructions missing in KVM/Xen.
* **Kernel module**: KVM is deeply integrated into Linux. Patches requires recompiling/installing the kernel. There's [kvm-kmod](https://github.com/bonzini/kvm-kmod) but it's high-maintenance: Linux/KVM patches easily break *kvm-kmod*, and *kvm-kmod* patches easily break older kernels. Instead, HAXM can be built and installed in few seconds.
* **Permissive licensing**: There's no QEMU-compatible hypervisor that supports Linux *and* offers a permissive license (e.g. BSD, MIT).

Additionally, we could take the chance to refactor a bit the codebase.

## Changes

- **Normalized prefixes**: Some platform-specific functions such as `smp_mb` or `__fls` had no `hax_`/`asm_`-prefix, clashing against functions declared in kernel headers.
- **Added Linux-specific code**: Headers at `include/linux/*.h`, sources at `linux/*.c`, alongside Makefile+Kbuild files.
- **Added Linux to TravisCI build matrix**: It should make maintenance easier.

## Pending

Essential tasks:
- [x] HAXM builds/links under Linux.
- [x] Finishing implementation of Linux-specific functions.
- [x] Testing changes.
- [x] Adding build instructions.

Optional tasks that imply major changes to the codebase (**needs discussion!**):
- **Single ioctl definitions**: Define ioctl's only once at `hax_interface.h` via `HAX_IOCTL(name, code, size)`. Then implement the `HAX_IOCTL` macro on `hax_interface_{darwin,windows,linux}.h` to make the actual platform-specific *ioctl* definitions. This will reduce code duplication.

### DIFF
--- a/include/hax.h
+++ b/include/hax.h
@@ -39,8 +39,6 @@
 // declaration
 struct vcpu_t;
 
-extern int hax_page_size;
-
 #define HAX_CUR_VERSION    0x0004
 #define HAX_COMPAT_VERSION 0x0001
 
@@ -207,7 +205,7 @@ void hax_set_page(phax_page page);
 
 static inline uint64_t hax_page2pa(phax_page page)
 {
-    return hax_page2pfn(page) << PAGE_SHIFT;
+    return hax_page2pfn(page) << HAX_PAGE_SHIFT;
 }
 
 #define hax_page_pa hax_page2pa
@@ -232,15 +230,15 @@ static inline unsigned char *hax_page_va(struct hax_page *page)
 #define HAX_MAX_CPUS (sizeof(uint64_t) * 8)
 
 /* Host SMP */
-extern cpumap_t cpu_online_map;
+extern hax_cpumap_t cpu_online_map;
 extern int max_cpus;
 
 #ifdef __cplusplus
 extern "C" {
 #endif
 
-int smp_call_function(cpumap_t *cpus, void(*scfunc)(void *param), void *param);
-extern int cpu_number(void);
+int hax_smp_call_function(hax_cpumap_t *cpus, void(*scfunc)(void *param),
+                          void *param);
 
 uint32_t hax_cpuid(void);
 int proc_event_pending(struct vcpu_t *vcpu);
--- a/include/hax.h
+++ b/include/hax.h
@@ -266,6 +266,9 @@ int hax_em64t_enabled(void);
 #ifdef HAX_PLATFORM_DARWIN
 #include "darwin/hax_mac.h"
 #endif
+#ifdef HAX_PLATFORM_LINUX
+#include "linux/hax_linux.h"
+#endif
 #ifdef HAX_PLATFORM_WINDOWS
 #include "windows/hax_windows.h"
 #endif
--- a/include/linux/hax_interface_linux.h
+++ b/include/linux/hax_interface_linux.h
@@ -1,0 +1,79 @@
+/*
+ * Copyright (c) 2011 Intel Corporation
+ * Copyright (c) 2018 Kryptos Logic
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ *   1. Redistributions of source code must retain the above copyright notice,
+ *      this list of conditions and the following disclaimer.
+ *
+ *   2. Redistributions in binary form must reproduce the above copyright
+ *      notice, this list of conditions and the following disclaimer in the
+ *      documentation and/or other materials provided with the distribution.
+ *
+ *   3. Neither the name of the copyright holder nor the names of its
+ *      contributors may be used to endorse or promote products derived from
+ *      this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef HAX_LINUX_HAX_INTERFACE_LINUX_H_
+#define HAX_LINUX_HAX_INTERFACE_LINUX_H_
+
+#include <asm/ioctl.h>
+
+/* The mac specific interface to qemu because of mac's
+ * special handling like hax tunnel allocation etc */
+/* HAX model level ioctl */
+#define HAX_IOCTL_VERSION _IOWR(0, 0x20, struct hax_module_version)
+#define HAX_IOCTL_CREATE_VM _IOWR(0, 0x21, uint32_t)
+#define HAX_IOCTL_DESTROY_VM _IOW(0, 0x22, uint32_t)
+#define HAX_IOCTL_CAPABILITY _IOR(0, 0x23, struct hax_capabilityinfo)
+#define HAX_IOCTL_SET_MEMLIMIT _IOWR(0, 0x24, struct hax_set_memlimit)
+
+// Only for backward compatibility with old Qemu.
+#define HAX_VM_IOCTL_VCPU_CREATE_ORIG _IOR(0, 0x80, int)
+
+#define HAX_VM_IOCTL_VCPU_CREATE _IOWR(0, 0x80, uint32_t)
+#define HAX_VM_IOCTL_ALLOC_RAM _IOWR(0, 0x81, struct hax_alloc_ram_info)
+#define HAX_VM_IOCTL_SET_RAM _IOWR(0, 0x82, struct hax_set_ram_info)
+#define HAX_VM_IOCTL_VCPU_DESTROY _IOR(0, 0x83, uint32_t)
+#define HAX_VM_IOCTL_ADD_RAMBLOCK _IOW(0, 0x85, struct hax_ramblock_info)
+#define HAX_VM_IOCTL_SET_RAM2 _IOWR(0, 0x86, struct hax_set_ram_info2)
+#define HAX_VM_IOCTL_PROTECT_RAM _IOWR(0, 0x87, struct hax_protect_ram_info)
+
+#define HAX_VCPU_IOCTL_RUN _IO(0, 0xc0)
+#define HAX_VCPU_IOCTL_SET_MSRS _IOWR(0, 0xc1, struct hax_msr_data)
+#define HAX_VCPU_IOCTL_GET_MSRS _IOWR(0, 0xc2, struct hax_msr_data)
+
+#define HAX_VCPU_IOCTL_SET_FPU _IOW(0, 0xc3, struct fx_layout)
+#define HAX_VCPU_IOCTL_GET_FPU _IOR(0, 0xc4, struct fx_layout)
+
+#define HAX_VCPU_IOCTL_SETUP_TUNNEL _IOWR(0, 0xc5, struct hax_tunnel_info)
+#define HAX_VCPU_IOCTL_INTERRUPT _IOWR(0, 0xc6, uint32_t)
+#define HAX_VCPU_SET_REGS _IOWR(0, 0xc7, struct vcpu_state_t)
+#define HAX_VCPU_GET_REGS _IOWR(0, 0xc8, struct vcpu_state_t)
+
+/* API 2.0 */
+#define HAX_VM_IOCTL_NOTIFY_QEMU_VERSION _IOW(0, 0x84, struct hax_qemu_version)
+
+#define HAX_IOCTL_VCPU_DEBUG _IOW(0, 0xc9, struct hax_debug_t)
+
+#define HAX_KERNEL64_CS 0x80
+#define HAX_KERNEL32_CS 0x08
+
+#define is_compatible() 0
+
+#endif  // HAX_LINUX_HAX_INTERFACE_LINUX_H_
--- a/platforms/linux/components.c
+++ b/platforms/linux/components.c
@@ -1,0 +1,612 @@
+/*
+ * Copyright (c) 2018 Kryptos Logic
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ *   1. Redistributions of source code must retain the above copyright notice,
+ *      this list of conditions and the following disclaimer.
+ *
+ *   2. Redistributions in binary form must reproduce the above copyright
+ *      notice, this list of conditions and the following disclaimer in the
+ *      documentation and/or other materials provided with the distribution.
+ *
+ *   3. Neither the name of the copyright holder nor the names of its
+ *      contributors may be used to endorse or promote products derived from
+ *      this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <linux/cred.h>
+#include <linux/dm-ioctl.h>
+#include <linux/miscdevice.h>
+#include <linux/namei.h>
+#include <linux/slab.h>
+#include <linux/uaccess.h>
+
+#include <linux/fs.h>
+
+#include "../../core/include/hax_core_interface.h"
+
+#define HAX_VM_DEVFS_FMT    "hax_vm/vm%02d"
+#define HAX_VCPU_DEVFS_FMT  "hax_vm%02d/vcpu%02d"
+
+typedef struct hax_vm_linux_t {
+    struct vm_t *cvm;
+    int id;
+    struct miscdevice dev;
+    char *devname;
+} hax_vm_linux_t;
+
+typedef struct hax_vcpu_linux_t {
+    struct vcpu_t *cvcpu;
+    struct hax_vm_linux_t *vm;
+    int id;
+    struct miscdevice dev;
+    char *devname;
+} hax_vcpu_linux_t;
+
+static int hax_vm_open(struct inode *inodep, struct file *filep);
+static int hax_vm_release(struct inode *inodep, struct file *filep);
+static long hax_vm_ioctl(struct file *filp, unsigned int cmd,
+                         unsigned long arg);
+
+static int hax_vcpu_open(struct inode *inodep, struct file *filep);
+static int hax_vcpu_release(struct inode *inodep, struct file *filep);
+static long hax_vcpu_ioctl(struct file *filp, unsigned int cmd,
+                           unsigned long arg);
+
+static struct file_operations hax_vm_fops = {
+    .owner          = THIS_MODULE,
+    .open           = hax_vm_open,
+    .release        = hax_vm_release,
+    .unlocked_ioctl = hax_vm_ioctl,
+    .compat_ioctl   = hax_vm_ioctl,
+};
+
+static struct file_operations hax_vcpu_fops = {
+    .owner          = THIS_MODULE,
+    .open           = hax_vcpu_open,
+    .release        = hax_vcpu_release,
+    .unlocked_ioctl = hax_vcpu_ioctl,
+    .compat_ioctl   = hax_vcpu_ioctl,
+};
+
+/* Component management */
+
+static void hax_component_perm(const char *devname, struct miscdevice *misc)
+{
+    int err;
+    struct path path;
+    struct inode *inode;
+    const struct cred *cred;
+    char devpath[DM_NAME_LEN];
+
+    if (!misc || !misc->this_device)
+        return;
+
+    snprintf(devpath, sizeof(devpath), "/dev/%s", devname);
+    err = kern_path(devpath, LOOKUP_FOLLOW, &path);
+    if (err || !path.dentry) {
+        hax_error("Could not obtain device inode\n");
+        return;
+    }
+    cred = get_current_cred();
+    inode = path.dentry->d_inode;
+    inode->i_uid.val = cred->uid.val;
+    inode->i_gid.val = cred->gid.val;
+    inode->i_mode |= 0660;
+}
+
+static hax_vcpu_linux_t* hax_vcpu_create_linux(struct vcpu_t *cvcpu,
+                                               hax_vm_linux_t *vm, int vcpu_id)
+{
+    hax_vcpu_linux_t *vcpu;
+
+    if (!cvcpu || !vm)
+        return NULL;
+
+    vcpu = kmalloc(sizeof(hax_vcpu_linux_t), GFP_KERNEL);
+    if (!vcpu)
+        return NULL;
+
+    memset(vcpu, 0, sizeof(hax_vcpu_linux_t));
+    vcpu->cvcpu = cvcpu;
+    vcpu->id = vcpu_id;
+    vcpu->vm = vm;
+    set_vcpu_host(cvcpu, vcpu);
+    return vcpu;
+}
+
+static void hax_vcpu_destroy_linux(hax_vcpu_linux_t *vcpu)
+{
+    struct vcpu_t *cvcpu;
+
+    if (!vcpu)
+        return;
+
+    cvcpu = vcpu->cvcpu;
+    hax_vcpu_destroy_hax_tunnel(cvcpu);
+    set_vcpu_host(cvcpu, NULL);
+    vcpu->cvcpu = NULL;
+    kfree(vcpu);
+}
+
+int hax_vcpu_create_host(struct vcpu_t *cvcpu, void *vm_host, int vm_id,
+                         int vcpu_id)
+{
+    int err;
+    hax_vcpu_linux_t *vcpu;
+    hax_vm_linux_t *vm;
+
+    vm = (hax_vm_linux_t *)vm_host;
+    vcpu = hax_vcpu_create_linux(cvcpu, vm, vcpu_id);
+    if (!vcpu)
+        return -1;
+
+    vcpu->devname = kmalloc(DM_NAME_LEN, GFP_KERNEL);
+    snprintf(vcpu->devname, DM_NAME_LEN, HAX_VCPU_DEVFS_FMT, vm_id, vcpu_id);
+    vcpu->dev.minor = MISC_DYNAMIC_MINOR;
+    vcpu->dev.name = vcpu->devname;
+    vcpu->dev.fops = &hax_vcpu_fops;
+
+    err = misc_register(&vcpu->dev);
+    if (err) {
+        hax_error("Failed to register HAXM-VCPU device\n");
+        hax_vcpu_destroy_linux(vcpu);
+        return -1;
+    }
+    hax_component_perm(vcpu->devname, &vcpu->dev);
+    hax_info("Created HAXM-VCPU device with minor=%d\n", vcpu->dev.minor);
+    return 0;
+}
+
+int hax_vcpu_destroy_host(struct vcpu_t *cvcpu, void *vcpu_host)
+{
+    hax_vcpu_linux_t *vcpu;
+
+    vcpu = (hax_vcpu_linux_t *)vcpu_host;
+    misc_deregister(&vcpu->dev);
+    kfree(vcpu->devname);
+
+    hax_vcpu_destroy_linux(vcpu);
+    return 0;
+}
+
+static hax_vm_linux_t *hax_vm_create_linux(struct vm_t *cvm, int vm_id)
+{
+    hax_vm_linux_t *vm;
+
+    if (!cvm)
+        return NULL;
+
+    vm = kmalloc(sizeof(hax_vm_linux_t), GFP_KERNEL);
+    if (!vm)
+        return NULL;
+
+    memset(vm, 0, sizeof(hax_vm_linux_t));
+    vm->cvm = cvm;
+    vm->id = vm_id;
+    set_vm_host(cvm, vm);
+    return vm;
+}
+
+static void hax_vm_destroy_linux(hax_vm_linux_t *vm)
+{
+    struct vm_t *cvm;
+
+    if (!vm)
+        return;
+
+    cvm = vm->cvm;
+    set_vm_host(cvm, NULL);
+    vm->cvm = NULL;
+    hax_vm_free_all_ram(cvm);
+    kfree(vm);
+}
+
+int hax_vm_create_host(struct vm_t *cvm, int vm_id)
+{
+    int err;
+    hax_vm_linux_t *vm;
+
+    vm = hax_vm_create_linux(cvm, vm_id);
+    if (!vm)
+        return -1;
+
+    vm->devname = kmalloc(DM_NAME_LEN, GFP_KERNEL);
+    snprintf(vm->devname, DM_NAME_LEN, HAX_VM_DEVFS_FMT, vm_id);
+    vm->dev.minor = MISC_DYNAMIC_MINOR;
+    vm->dev.name = vm->devname;
+    vm->dev.fops = &hax_vm_fops;
+
+    err = misc_register(&vm->dev);
+    if (err) {
+        hax_error("Failed to register HAXM-VM device\n");
+        hax_vm_destroy_linux(vm);
+        return -1;
+    }
+    hax_component_perm(vm->devname, &vm->dev);
+    hax_info("Created HAXM-VM device with minor=%d\n", vm->dev.minor);
+    return 0;
+}
+
+/* When coming here, all vcpus should have been destroyed already. */
+int hax_vm_destroy_host(struct vm_t *cvm, void *vm_host)
+{
+    hax_vm_linux_t *vm;
+
+    vm = (hax_vm_linux_t *)vm_host;
+    misc_deregister(&vm->dev);
+    kfree(vm->devname);
+
+    hax_vm_destroy_linux(vm);
+    return 0;
+}
+
+/* No corresponding function in Linux side, it can be cleaned later. */
+int hax_destroy_host_interface(void)
+{
+    return 0;
+}
+
+/* VCPU operations */
+
+static int hax_vcpu_open(struct inode *inodep, struct file *filep)
+{
+    int ret;
+    struct vcpu_t *cvcpu;
+    struct hax_vcpu_linux_t *vcpu;
+    struct miscdevice *miscdev;
+
+    miscdev = filep->private_data;
+    vcpu = container_of(miscdev, struct hax_vcpu_linux_t, dev);
+    cvcpu = hax_get_vcpu(vcpu->vm->id, vcpu->id, 1);
+
+    hax_log_level(HAX_LOGD, "HAX vcpu open called\n");
+    if (!cvcpu)
+        return -ENODEV;
+
+    ret = hax_vcpu_core_open(cvcpu);
+    if (ret)
+        hax_error("Failed to open core vcpu\n");
+    hax_put_vcpu(cvcpu);
+    return ret;
+}
+
+static int hax_vcpu_release(struct inode *inodep, struct file *filep)
+{
+    int ret = 0;
+    struct vcpu_t *cvcpu;
+    struct hax_vcpu_linux_t *vcpu;
+    struct miscdevice *miscdev;
+
+    miscdev = filep->private_data;
+    vcpu = container_of(miscdev, struct hax_vcpu_linux_t, dev);
+    cvcpu = hax_get_vcpu(vcpu->vm->id, vcpu->id, 1);
+
+    hax_log_level(HAX_LOGD, "HAX vcpu close called\n");
+    if (!cvcpu) {
+        hax_error("Failed to find the vcpu, is it closed already?\n");
+        return 0;
+    }
+
+    /* put the one for vcpu create */
+    hax_put_vcpu(cvcpu);
+    /* put the one just held */
+    hax_put_vcpu(cvcpu);
+    return ret;
+}
+
+static long hax_vcpu_ioctl(struct file *filp, unsigned int cmd,
+                           unsigned long arg)
+{
+    int ret = 0;
+    void *argp = (void *)arg;
+    struct vcpu_t *cvcpu;
+    struct hax_vcpu_linux_t *vcpu;
+    struct miscdevice *miscdev;
+
+    miscdev = filp->private_data;
+    vcpu = container_of(miscdev, struct hax_vcpu_linux_t, dev);
+    cvcpu = hax_get_vcpu(vcpu->vm->id, vcpu->id, 1);
+    if (!cvcpu)
+        return -ENODEV;
+
+    switch (cmd) {
+    case HAX_VCPU_IOCTL_RUN:
+        ret = vcpu_execute(cvcpu);
+        break;
+    case HAX_VCPU_IOCTL_SETUP_TUNNEL: {
+        struct hax_tunnel_info info;
+        ret = hax_vcpu_setup_hax_tunnel(cvcpu, &info);
+        if (copy_to_user(argp, &info, sizeof(info))) {
+            ret = -EFAULT;
+            break;
+        }
+        break;
+    }
+    case HAX_VCPU_IOCTL_SET_MSRS: {
+        struct hax_msr_data msrs;
+        struct vmx_msr *msr;
+        int i, fail;
+
+        if (copy_from_user(&msrs, argp, sizeof(msrs))) {
+            ret = -EFAULT;
+            break;
+        }
+        msr = msrs.entries;
+        /* nr_msr needs to be verified */
+        if (msrs.nr_msr >= 0x20) {
+            hax_error("MSRS invalid!\n");
+            return -EFAULT;
+        }
+        for (i = 0; i < msrs.nr_msr; i++, msr++) {
+            fail = vcpu_set_msr(cvcpu, msr->entry, msr->value);
+            if (fail) {
+                break;
+            }
+        }
+        msrs.done = i;
+        break;
+    }
+    case HAX_VCPU_IOCTL_GET_MSRS: {
+        struct hax_msr_data msrs;
+        struct vmx_msr *msr;
+        int i, fail;
+
+        if (copy_from_user(&msrs, argp, sizeof(msrs))) {
+            ret = -EFAULT;
+            break;
+        }
+        msr = msrs.entries;
+        if(msrs.nr_msr >= 0x20) {
+            hax_error("MSRS invalid!\n");
+            return -EFAULT;
+        }
+        for (i = 0; i < msrs.nr_msr; i++, msr++) {
+            fail = vcpu_get_msr(cvcpu, msr->entry, &msr->value);
+            if (fail) {
+                break;
+            }
+        }
+        msrs.done = i;
+        if (copy_to_user(argp, &msrs, sizeof(msrs))) {
+            ret = -EFAULT;
+            break;
+        }
+        break;
+    }
+    case HAX_VCPU_IOCTL_SET_FPU: {
+        struct fx_layout fl;
+        if (copy_from_user(&fl, argp, sizeof(fl))) {
+            ret = -EFAULT;
+            break;
+        }
+        ret = vcpu_put_fpu(cvcpu, &fl);
+        break;
+    }
+    case HAX_VCPU_IOCTL_GET_FPU: {
+        struct fx_layout fl;
+        ret = vcpu_get_fpu(cvcpu, &fl);
+        if (copy_to_user(argp, &fl, sizeof(fl))) {
+            ret = -EFAULT;
+            break;
+        }
+        break;
+    }
+    case HAX_VCPU_SET_REGS: {
+        struct vcpu_state_t vc_state;
+        if (copy_from_user(&vc_state, argp, sizeof(vc_state))) {
+            ret = -EFAULT;
+            break;
+        }
+        ret = vcpu_set_regs(cvcpu, &vc_state);
+        break;
+    }
+    case HAX_VCPU_GET_REGS: {
+        struct vcpu_state_t vc_state;
+        ret = vcpu_get_regs(cvcpu, &vc_state);
+        if (copy_to_user(argp, &vc_state, sizeof(vc_state))) {
+            ret = -EFAULT;
+            break;
+        }
+        break;
+    }
+    case HAX_VCPU_IOCTL_INTERRUPT: {
+        uint8_t vector;
+        if (copy_from_user(&vector, argp, sizeof(vector))) {
+            ret = -EFAULT;
+            break;
+        }
+        vcpu_interrupt(cvcpu, vector);
+        break;
+    }
+    case HAX_IOCTL_VCPU_DEBUG: {
+        struct hax_debug_t hax_debug;
+        if (copy_from_user(&hax_debug, argp, sizeof(hax_debug))) {
+            ret = -EFAULT;
+            break;
+        }
+        vcpu_debug(cvcpu, &hax_debug);
+        break;
+    }
+    default:
+        // TODO: Print information about the process that sent the ioctl.
+        hax_error("Unknown VCPU IOCTL 0x%lx\n", cmd);
+        ret = -ENOSYS;
+        break;
+    }
+    hax_put_vcpu(cvcpu);
+    return ret;
+}
+
+/* VM operations */
+
+static int hax_vm_open(struct inode *inodep, struct file *filep)
+{
+    int ret;
+    struct vm_t *cvm;
+    struct hax_vm_linux_t *vm;
+    struct miscdevice *miscdev;
+
+    miscdev = filep->private_data;
+    vm = container_of(miscdev, struct hax_vm_linux_t, dev);
+    cvm = hax_get_vm(vm->id, 1);
+    if (!cvm)
+        return -ENODEV;
+
+    ret = hax_vm_core_open(cvm);
+    hax_put_vm(cvm);
+    hax_log_level(HAX_LOGI, "Open VM\n");
+    return ret;
+}
+
+static int hax_vm_release(struct inode *inodep, struct file *filep)
+{
+    struct vm_t *cvm;
+    struct hax_vm_linux_t *vm;
+    struct miscdevice *miscdev;
+
+    miscdev = filep->private_data;
+    vm = container_of(miscdev, struct hax_vm_linux_t, dev);
+    cvm = hax_get_vm(vm->id, 1);
+
+    hax_log_level(HAX_LOGI, "Close VM\n");
+    if (cvm) {
+        /* put the ref get just now */
+        hax_put_vm(cvm);
+        hax_put_vm(cvm);
+    }
+    return 0;
+}
+
+static long hax_vm_ioctl(struct file *filp, unsigned int cmd,
+                         unsigned long arg)
+{
+    int ret = 0;
+    void *argp = (void *)arg;
+    struct vm_t *cvm;
+    struct hax_vm_linux_t *vm;
+    struct miscdevice *miscdev;
+
+    miscdev = filp->private_data;
+    vm = container_of(miscdev, struct hax_vm_linux_t, dev);
+    cvm = hax_get_vm(vm->id, 1);
+    if (!cvm)
+        return -ENODEV;
+
+    switch (cmd) {
+    case HAX_VM_IOCTL_VCPU_CREATE:
+    case HAX_VM_IOCTL_VCPU_CREATE_ORIG: {
+        uint32_t vcpu_id, vm_id;
+        struct vcpu_t *cvcpu;
+
+        vm_id = vm->id;
+        if (copy_from_user(&vcpu_id, argp, sizeof(vcpu_id))) {
+            ret = -EFAULT;
+            break;
+        }
+        cvcpu = vcpu_create(cvm, vm, vcpu_id);
+        if (!cvcpu) {
+            hax_error("Failed to create vcpu %x on vm %x\n", vcpu_id, vm_id);
+            ret = -EINVAL;
+            break;
+        }
+        break;
+    }
+    case HAX_VM_IOCTL_ALLOC_RAM: {
+        struct hax_alloc_ram_info info;
+        if (copy_from_user(&info, argp, sizeof(info))) {
+            ret = -EFAULT;
+            break;
+        }
+        hax_info("IOCTL_ALLOC_RAM: vm_id=%d, va=0x%llx, size=0x%x, pad=0x%x\n",
+                 vm->id, info.va, info.size, info.pad);
+        ret = hax_vm_add_ramblock(cvm, info.va, info.size);
+        break;
+    }
+    case HAX_VM_IOCTL_ADD_RAMBLOCK: {
+        struct hax_ramblock_info info;
+        if (copy_from_user(&info, argp, sizeof(info))) {
+            ret = -EFAULT;
+            break;
+        }
+        if (info.reserved) {
+            hax_error("IOCTL_ADD_RAMBLOCK: vm_id=%d, reserved=0x%llx\n",
+                      vm->id, info.reserved);
+            return -EINVAL;
+        }
+        hax_info("IOCTL_ADD_RAMBLOCK: vm_id=%d, start_va=0x%llx, size=0x%llx\n",
+                 vm->id, info.start_va, info.size);
+        ret = hax_vm_add_ramblock(cvm, info.start_va, info.size);
+        break;
+    }
+    case HAX_VM_IOCTL_SET_RAM: {
+        struct hax_set_ram_info info;
+        if (copy_from_user(&info, argp, sizeof(info))) {
+            ret = -EFAULT;
+            break;
+        }
+        ret = hax_vm_set_ram(cvm, &info);
+        break;
+    }
+#ifdef CONFIG_HAX_EPT2
+    case HAX_VM_IOCTL_SET_RAM2: {
+        struct hax_set_ram_info2 info;
+        if (copy_from_user(&info, argp, sizeof(info))) {
+            ret = -EFAULT;
+            break;
+        }
+        if (info.reserved1 || info.reserved2) {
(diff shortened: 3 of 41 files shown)


raphaelning:
Thanks, Alexandro. Great to see you back with another big PR!

Windows and Mac have been and will still be the Intel HAXM team's priority, and realistically, we wouldn't add a third host OS to our support matrix. However, if the open source community wants this feature and can provide enough help, there's no reason for us to stand in the way, not to mention that this is a great chance to improve the overall code quality of the project ;-)

We'll accept this PR on these conditions:

1. It doesn't cause any regression on Windows and Mac hosts.
1. You will develop the QEMU-side patch for testing this PR, and do your own testing on Linux. (We'll try to run some simple tests on a Linux host before merging it, but that won't be enough coverage.)
1. You will be the maintainer of all the Linux-specific code that this PR adds to the source tree. In other words, in case someone else submits a PR for such code in the future, we will invite you to the review and will accept or reject the change based on your feedback.

What do you think? I have some other tasks in hand at the moment, but I'll try to respond to the major changes you have proposed in the next few days. The minor changes all look fine to me :-)

AlexAltea:
Sounds good to me! I will provide support for Linux frontend. Additionally, the planned changes will ensure easier maintenance in the future (e.g. by reducing code duplication in platform-specific code).

I will comment on this PR to discuss any further major changes that might affect generic or Windows/macOS-specific code. Of course, once the patch is ready, I will run tests on the three platforms (of course, this implies taking care of QEMU-side patches myself) and let you know once everything is ready.

Also on the topic of testing: After completing this PR, it might be a good moment to setup something like [Travis CI](https://travis-ci.org/)? This should also reduce maintenance effort.

raphaelning:
> Sounds good to me! I will provide support for Linux frontend. Additionally, the planned changes will ensure easier maintenance in the future (e.g. by reducing code duplication in platform-specific code).

Thanks! Indeed.

Regarding the first proposed major change:

> Combine platform-specific folders. Moving {darwin,windows,linux}/ to X/{darwin,windows,linux}/, to avoid polluting the root folder. Folder "X" could be called drivers, platforms, etc.

Agreed. Right now we have three kinds of files in `{darwin,windows}/`:

a) Makefiles (e.g. `windows/IntelHaxm.vcxproj`).
b) C/C++ source files that implement for the host OS the required interface specified by a shared header (e.g. both `windows/hax_host_mem.c` and `darwin/.../hax_host_mem.cpp` implement `include/hax_host_mem.h`).
c) Other C/C++ source files specific to the host OS, mainly to implement the driver interface specified by the host OS (e.g. `windows/hax_entry.c`).

Since `drivers` only covers (c), I'd prefer `platforms`.

Two more questions related to this topic:

1. Currently, shared headers (as described by (b) above) exist in both `core/include/` and `include/`, and it's not very clear to me how to choose between them when we create a new shared header. Is there a better scheme?
1. Have you thought about using CMake to generate the Linux makefiles, with an eye to eventually do so for Windows and Mac as well using a shared CMake configuration? This could be a crazy idea, because last time I checked, CMake didn't have mature support for WDK projects.

I'll comment on the other proposed major changes later.

raphaelning:
> Combine documentation: Moving API.md to X/api.md, and add platform-specific build instructions in X/building.md or X/building-{darwin,windows,linux}.md to avoid polluting README.md. Folder "X" could be called docs, etc.

Agreed. `docs` is a good name.

> Single ioctl definitions: Define ioctl's only once at hax_interface.h via HAX_IOCTL(name, code, size). Then implement the HAX_IOCTL macro on hax_interface_{darwin,windows,linux}.h to make the actual platform-specific ioctl definitions. This will reduce code duplication.

Good idea! Agreed.

> Normalized integer types: The codebase mixes intN_t and intN. It would convenient to pick one variant and consistently use it.

Agreed. I see that you have started doing this and have chosen `[u]intN_t` over `[u]intN`, which is the harder choice (because the latter types are more heavily used throughout the codebase), but also the right one, I'd say (because the former types are more familiar, as they resemble the standard types in `<stdint.h>`).

The question is what's the best strategy to implement such a ubiquitous change, so as to avoid rebase headaches. Here's what I think:

1. Don't mix it with other changes.
1. Create separate PR(s) for it, which can be merged before this one.
1. Split it into multiple commits/PRs, each one covering as few as one source file (e.g. `vcpu.c`) or folder, which can be reviewed and merged quickly.

Please let me know if this is reasonable or if you have better ideas.

raphaelning:
> Also on the topic of testing: After completing this PR, it might be a good moment to setup something like Travis CI? This should also reduce maintenance effort.

Good idea! I just started reading the Travis CI documentation, and it seems to work best with Linux Makefile projects. Since there's a risk of a common (`core`) patch causing a regression on Linux hosts, maybe we can set up Travis CI for Linux first, which will at least tell us right away if a patch breaks the build on Linux. I'm not sure if it can help us run QEMU-based tests, though. Neither do I know if it's possible to use it for Windows and Mac builds.

AlexAltea:
> Have you thought about using CMake to generate the Linux makefiles, with an eye to eventually do so for Windows and Mac as well using a shared CMake configuration?

I didn't considered it yet, but that is much more convenient. For now, we could have each platform-specific build script (Makefile, Xcode-files, VS-files) in their corresponding `platforms/*` folder, and merge everything under a single root-level CMakeLists for a future PR.

> The question is what's the best strategy to implement such a ubiquitous change, so as to avoid rebase headaches. Here's what I think:

Agreed, I'll submit all refactoring-related changes as a separate PR.

> I'm not sure if it can help us run QEMU-based tests, though. 

We can execute arbitrary scripts on Travis (build scripts run in sudo-enabled virtual machines), so it should be possible as long as their hardware supports it and they virtualize EPT to allow nested virtualization.

> Neither do I know if it's possible to use it for Windows and Mac builds.

Travis can handle Linux, macOS and Windows:
https://docs.travis-ci.com/user/reference/overview/#virtualization-environments

linux/hax_entry.c
```diff
@@ -0,0 +1,94 @@
...
+#include <linux/init.h>
+#include <linux/module.h>
+#include <linux/kernel.h>
+#include <linux/fs.h>
+#include <linux/miscdevice.h>
+
+#include "../include/hax_interface.h"
+
+MODULE_LICENSE("BSD");
+MODULE_AUTHOR("Intel Corporation");
```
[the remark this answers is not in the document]
AlexAltea: It's possible to add multiple `MODULE_AUTHOR` statements.

Listing both will probably be the best approach to reflect the authorship of the Linux kernel module (`Kryptos Logic LLC`) and the HAXM core code (`Intel Corporation`).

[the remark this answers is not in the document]
raphaelning: Sounds good to me :-) Let's do that for now, and I'll let you know if Intel Legal has other ideas.

AlexAltea:
> Currently, shared headers (as described by (b) above) exist in both core/include/ and include/, and it's not very clear to me how to choose between them when we create a new shared header. Is there a better scheme?

Include folders give me quite some headaches. There's two issues:

1. The current approach to include headers is weird: We explicitly mention the "include" folders directly in the paths, e.g. `#include "../include/*.h"`, which most C/C++ codebases rarely do. Instead, those include directories should be added to the Makefile/Projects to simply write `#include <*.h>`.

2. It's not clear what each `include` folder should contain. I see 5 types of headers:
    - Headers of **core** consumed by **platforms** (e.g.: `core/include/hax_core_interface.h`):
      Right now they are in `core/include`, which might be the appropriate place.
    - Headers of **core** consumed by **core** (e.g.: `core/include/vmx.h`).
      Right now they are in `core/include`. Maybe they should be next to their sources: that is, moving them to `core`, e.g. `core/vmx.h`. There's no reason for having them separate from sources, since there are no external consumers. This is *already* the case for platforms headers consumed by platforms (see below).
    - Headers for **platforms** consumed by **core**, (e.g.: `include/hax_host_mem.h`).
      Right now they are in `include`... Does that make sense?
    - Headers for **platforms** consumed by **platforms**, (e.g.: `windows/hax_win.h`).
      Right now they are each next to their sources.
    - Headers of **platforms** consumed by QEMU (e.g.: `include/hax_interface.h`).
      Right now they are in `include`, which seems alright since they are the only definitions that HAXM needs to export for 3rd party software.

Anyway, this comment is just brainstorming on my side. I still need to give this problem more thought.

raphaelning:
Ah, great question. Yes, I do know what they are.

When I first took over the project, it didn't even have a Visual Studio project (`.sln` or `.vcxproj`). Instead, it was using the WDK 7.1 build system to build the Windows driver. That build system was pretty straightforward/basic: you just need to provide very simple makefiles like `dirs` and `sources`, and then run `build.exe`.

Later we decided to migrate to WDK 10 and Visual Studio 2015. I did that in two steps:

1. Convert the `dirs`-based project to a Visual Studio solution, using a tool available in Visual Studio 2012. I think it was the [Nmake2MsBuild](https://docs.microsoft.com/en-us/windows-hardware/drivers/devtest/converting-a-wdk-sources-file-to-a-visual-studio-project) CLI tool, or its GUI equivalent. The tool is not available in later releases of Visual Studio.
2. Open the VS2012 solution in VS2015 and proceed with the format upgrade.

Step 1 ended up generating all the VS files we see now. Since they are autogenerated, they are much more complicated than the original `dirs` and `sources` files (if you remember, we used to have a lot of build configurations, `Win{7, 8, 8.1} {Debug, Release}`, which were created by the conversion tool). For example, `sources.props` was generated from `sources`. The tool also created the awkwardly-named `dirs-Package` project, and made it the default project of `HaxmDriver.sln`.

I believe we can safely delete the old makefiles (`dirs`, `sources`, `Makefile.inc`, etc.). But then, you may still consider `HaxmDriver.sln` and the `dirs-Package` folder as polluting the root of the source tree. If we somehow make CMake work for Windows, we'll be able to get rid of all the VS files.

raphaelning:
> Include folders give me quite some headaches. There's two issues:

I completely agree. Thanks for the really nice summary.

> Headers of core consumed by core (e.g.: core/include/vmx.h).
>Right now they are in core/include. Maybe they should be next to their sources: that is, moving them to core, e.g. core/vmx.h. There's no reason for having them separate from sources, since there are no external consumers. This is already the case for platforms headers consumed by platforms (see below).

I think `vmx.h` is somewhat similar to `ia32_defs.h`. They define constants and data structures according to Intel SDM, and those definitions are not unique to HAXM. In some sense, they are similar to `hax_types.h`. Therefore, although they seem to be consumed by `core` at the moment, they can be consumed by platform sources as well (or at least we should allow that). Shall we move them to a separate folder under `include/`?

> Headers for platforms consumed by core, (e.g.: include/hax_host_mem.h).
> Right now they are in include... Does that make sense?

No. And I should take the blame. I created that header but wasn't sure where to put it, and ended up making the wrong decision.

linux/hax_entry.c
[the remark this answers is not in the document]
raphaelning: I reached out to colleagues with many years of experience in open source development, and got some important advice from a senior Linux developer/maintainer. A point you and I overlooked earlier is:

> Linux kernel modules should usually be GPL not BSD. It’s highly unlikely the module contains no dependencies on the Linux kernel, so the resulting work is almost certainly GPL (which is fine; BSD to GPL is allowed).

Therefore, this file, as well as probably all other Linux-specific source files, needs to use the GPL license. I know that you wanted a BSD-licensed hypervisor for Linux, but would you compromise on a small number of source files?

In case you do, I've been advised that you should:

* Use the Linux/GPL copyright header with your own copyright notices, so it's clear that this is not an Intel module.
* State that this kernel module is derived from Intel HAXM, and reproduce the Intel copyright as required by the BSD license.
* Call `MODULE_AUTHOR()` only once, using your company name and email. There's no need to include Intel here, since the main purpose is to let people know how to contact the maintainer of the module.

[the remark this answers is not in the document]
AlexAltea: Sure, I will apply those changes.

After checking documentation for `MODULE_LICENSE`, it seems the closest match to what we want is `"Dual BSD/GPL"` (defined as: "*GNU Public License v2 or BSD license choice*").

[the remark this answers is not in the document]
raphaelning: > After checking documentation for MODULE_LICENSE, it seems the closest match to what we want is "Dual BSD/GPL"

Hmm, that's a good point. I thought we would have to use just "GPL" because of the "Linux kernel modules should usually be GPL not BSD" comment, but apparently there's also this dual license option, which is actually adopted by a lot of in-tree modules, e.g.:

https://github.com/torvalds/linux/blob/master/mm/zsmalloc.c
https://github.com/torvalds/linux/blob/master/drivers/video/fbdev/intelfb/intelfbdrv.c

So I agree with your choice.

> Also, I was under the impression that GPL-compatibility was determined automatically at build-time depending on whether GPL symbols were imported and which MODULE_LICENSE was chosen.

That's another good point. IIUC, you are saying that in order to retain the dual license status, we have to avoid using any function marked as `EXPORT_SYMBOL_GPL`. That's beyond my knowledge, so please feel free to carry on with this approach while I reach out for help.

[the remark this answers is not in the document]
raphaelning: According to the aforementioned Linux expert, even dual BSD/GPL licensed modules don't have to avoid functions marked as `EXPORT_SYMBOL_GPL`. So please don't worry too much if you end up having to use one of them.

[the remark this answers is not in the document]
AlexAltea: Great, good to know. Thanks for the information!

AlexAltea:
Brief status update: I've noticed that platform-specific changes on the QEMU-side are minimal. In fact, the *hax-darwin.h* and *hax-darwin.c* files are not specific to Darwin, but to any POSIX-compliant system, and managed to come up with a preliminary patch at https://github.com/kryptoslogic/qemu/commit/48db0b4f0bcd952e09c3355c2b64948bf703da58.

I've renamed `hax-darwin.*` to `hax-posix.*`, which seems to be the standard suffix that QEMU uses for POSIX-specific sources (search on https://github.com/qemu/qemu/find/master). That way, we will reduce even further the maintenance effort required by this PR.

If you agree, I'll submit the corresponding patch to QEMU after this PR has been merged.

raphaelning:
> I've renamed hax-darwin.* to hax-posix.*, [...] If you agree, I'll submit the corresponding patch to QEMU after this PR has been merged.

Sounds great, thanks!

> Unless the issue with the confusing include's causes problems, I'd say we could leave it for a future PR, after this one has been merged.

Agreed.

AlexAltea:
HAXM now builds and runs successfully on Linux, or at least on Ubuntu 18.04 (Linux 4.18.0-rc3+, 64-bit).
After updating the TravisCI script, since they  offer Ubuntu 14.04, I've verified that HAXM links fine against older kernels, specifically 4.4.0. It will probably against any Linux 4.x kernel.

More importantly: **The PR is ready for review now.** Code-wise there's nothing pending. As you see, I've followed the same design choices as in the Windows/Darwin frontends.

PS: Mandatory screenshot: :-)
capture

platforms/linux/hax_mem_alloc.c
```diff
@@ -0,0 +1,168 @@
...
+    if (size == 0)
+        return NULL;
+
+    if (flags == 0)
+        flags = HAX_MEM_NONPAGE;
+
+    // NOTE: Flags ignored. Linux allows only non-pageable memory in kernel.
+    ptr = kmalloc(size, GFP_KERNEL);
+    if (ptr) {
+        memset(ptr, 0, size);
```
raphaelning: nit: Could use `kzalloc()` instead of `kmalloc()`

platforms/linux/hax_mem_alloc.c
```diff
@@ -0,0 +1,168 @@
...
+}
+
+void hax_vfree_flags(void *va, uint32_t size, uint32_t flags)
+{
+    (void)size;
+
+    if (flags == 0)
+        flags = HAX_MEM_NONPAGE;
+
+    // NOTE: Flags ignored. Linux allows only non-pageable memory in kernel.
```
raphaelning: nit: Get rid of the above `if` block

platforms/linux/hax_mem_alloc.c
```diff
@@ -0,0 +1,168 @@
...
+{
+    if (!page)
+        return 0;
+
+    return page->pa >> PAGE_SHIFT;
+}
+
+void hax_clear_page(struct hax_page *page)
+{
+    memset((void*)page->kva, 0, PAGE_SIZE);
```
raphaelning: nit: `void *`

platforms/linux/hax_mem_alloc.c
```diff
@@ -0,0 +1,168 @@
...
+}
+
+void hax_set_page(struct hax_page *page)
+{
+    memset((void*)page->kva, 0xFF, PAGE_SIZE);
```
raphaelning: nit: `void *`

platforms/linux/hax_mm.c
```diff
@@ -0,0 +1,124 @@
...
+
+    if (!mem || !size)
+        return -EINVAL;
+
+    hinfo = kmalloc(sizeof(struct hax_vcpu_mem_hinfo_t), GFP_KERNEL);
+    if (!hinfo)
+        return -ENOMEM;
+
+    nr_pages = ((size - 1) / PAGE_SIZE) + 1;
+    pages = kmalloc(sizeof(struct page*) * nr_pages, GFP_KERNEL);
```
raphaelning: nit: `struct page *`

platforms/linux/hax_wrapper.c
```diff
@@ -0,0 +1,336 @@
...
+
+#include <asm/cmpxchg.h>
+
+int default_hax_log_level = 3;
+int max_cpus;
+hax_cpumap_t cpu_online_map;
+
+int hax_log_level(int level, const char *fmt,  ...)
+{
+    char buffer[256];
```
raphaelning: Hmm, this is not ideal. I'm pretty sure this buffer is not big enough for some existing log messages. Can we just pass the `va_list` to `printk`? If not, let's add a TODO comment in case someone else sees it and comes up with an elegant solution.

platforms/linux/hax_wrapper.c
```diff
@@ -0,0 +1,336 @@
...
+    hax_cpumap_t *cpus;
+    uint32_t cpuid;
+
+    cpus = info->cpus;
+    cpuid = hax_cpuid();
+    if (*cpus & (0x1 << cpuid))
+        info->func(info->param);
+}
+
+int hax_smp_call_function(hax_cpumap_t *cpus, void (*scfunc)(void *), void * param)
```
raphaelning: nit: Wrap this long line

platforms/linux/hax_wrapper.c
```diff
@@ -0,0 +1,336 @@
...
+/* XXX */
+int proc_event_pending(struct vcpu_t *vcpu)
+{
+    return vcpu_event_pending(vcpu);
+}
+
+void hax_disable_preemption(preempt_flag *eflags)
+{
+    mword flags;
+#ifdef __x86_64__
```
raphaelning: Can we use `preempt_disable()` instead? Or you can add a TODO in case someone else wants to try it later.

platforms/linux/hax_wrapper.c
```diff
@@ -0,0 +1,336 @@
...
+        : "=r" (flags)
+    );
+#endif
+    *eflags = flags;
+    hax_disable_irq();
+}
+
+void hax_enable_preemption(preempt_flag *eflags)
+{
+    if (*eflags & X86_EFLAGS_IF)
```
raphaelning: Consider using `preempt_enable()` instead, or add a TODO

platforms/linux/hax_wrapper.c
```diff
@@ -0,0 +1,336 @@
...
+    addr = (unsigned long *)memory;
+    return test_and_clear_bit(bit, addr);
+}
+
+/* Spinlock */
+struct hax_spinlock {
+    spinlock_t lock;
+};
+
+hax_spinlock* hax_spinlock_alloc_init(void)
```
raphaelning: nit: `hax_spinlock *`

raphaelning:
Ah, I just realized that I've only reviewed the first two commits. Will have to leave the other two for tomorrow...

include/linux/hax_interface_linux.h
```diff
@@ -0,0 +1,81 @@
+/*
+ * Copyright (c) 2011 Intel Corporation
```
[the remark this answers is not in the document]
AlexAltea: Sorry, I didn't pay much attention to the file headers, so the authorship is certainly inconsistent. 

Will do the changes you suggested! I've never seen the *Original work* or *Modified work* prefixes either. I think we should drop these prefixes and simply list author copyright notices sorted by date (descending as time increases?). E.g.:
```
/*
 * Copyright (c) 2011 Intel Corporation
 * Copyright (c) 2018 Kryptos Logic
 * [...]
```
So if a third party forks and modifies this file, he can simply add his name below.
(I will apply these changes to all files that contain code originally authored by Intel).

platforms/linux/components.c
```diff
@@ -0,0 +1,496 @@
...
+
+    vcpu->devname = kmalloc(DM_NAME_LEN, GFP_KERNEL);
+    snprintf(vcpu->devname, DM_NAME_LEN, HAX_VCPU_DEVFS_FMT, vm_id, vcpu_id);
+    vcpu->dev.minor = MISC_DYNAMIC_MINOR;
+    vcpu->dev.name = vcpu->devname;
+    vcpu->dev.fops = &hax_vcpu_fops;
+
+    err = misc_register(&vcpu->dev);
+    if (err) {
+        printk(KERN_ERR "Failed to register HAXM-VCPU device\n");
```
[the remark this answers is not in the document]
AlexAltea: Good point. I have chosen `hax_error` to remain consistent with the rest of the code, and will also update similar functions that invoked *printk* for no good reason.

platforms/linux/hax_host_mem.c
```diff
@@ -0,0 +1,190 @@
...
+#include "../../include/hax_host_mem.h"
+#include "../../core/include/paging.h"
+
+#include <linux/mm.h>
+#include <linux/pagemap.h>
+#include <linux/slab.h>
+#include <linux/version.h>
+#include <linux/vmalloc.h>
+
+int hax_pin_user_pages(uint64 start_uva, uint64 size, hax_memdesc_user *memdesc)
```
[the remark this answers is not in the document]
AlexAltea: The issue is that the `linux` branch was created before the "*uintN to uintN_t*" patch, and I forgot to update the integer types after rebasing. As for where it's defined, Very likely in the Linux headers.

Either way, I'll change all these missing types back to *uintN_t*.

platforms/linux/hax_mem_alloc.c
AlexAltea: Good point. Thank you!

platforms/linux/hax_wrapper.c
AlexAltea: I had tried several approaches before this one, and the necessity of adding the `haxm_*:` prefix caused problems in all of them. However, after checking [this article](https://lwn.net/Articles/310194/), I think I have found the perfect solution via the `%v` placeholder:

> `printk("%shaxm_level: %v, level, fmt, args);`

I will apply this patch to all logging-related functions.

include/linux/hax_interface_linux.h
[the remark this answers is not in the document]
raphaelning: > descending as time of first modification increases?

The example looks good to me, thanks! (I think it uses the _ascending_ order of year.)

> So if a third party forks and modifies this file, he can simply add his name below.

Exactly. Also I think it's a common practice to aggregate the notices by author. So if Intel updates the file substantially next year, we'll just update the first record (`2011, 2019`) instead of adding a new one.

platforms/linux/hax_host_mem.c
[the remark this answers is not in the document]
raphaelning: I see, thanks. I think I've seen a `uint8` in one of the new source files as well, so maybe you want to catch all of these with a regex.

platforms/linux/hax_wrapper.c
raphaelning: I see, that's indeed an elegant solution!

platforms/linux/hax_wrapper.c
```diff
@@ -0,0 +1,336 @@
...
+    mutex_init(lock);
+    return lock;
+}
+
+void hax_mutex_lock(hax_mutex lock)
+{
+    if (!lock)
+        return;
+
+    mutex_lock((struct mutex*)lock);
```
raphaelning: nit: `struct mutex *`

platforms/linux/hax_wrapper.c
```diff
@@ -0,0 +1,336 @@
...
+}
+
+void hax_mutex_unlock(hax_mutex lock)
+{
+    if (!lock)
+        return;
+
+    mutex_unlock((struct mutex*)lock);
```
raphaelning: nit: `struct mutex *`

platforms/linux/hax_wrapper.c
```diff
@@ -0,0 +1,336 @@
...
+}
+
+void hax_mutex_free(hax_mutex lock)
+{
+    if (!lock)
+        return;
+
+    mutex_destroy((struct mutex*)lock);
```
raphaelning: nit: `struct mutex *`

platforms/linux/hax_wrapper.c
```diff
@@ -0,0 +1,336 @@
...
+    if (ret == old_val)
+        return true;
+    else
+        return false;
+}
+
+/* Atomics */
+hax_atomic_t hax_atomic_add(volatile hax_atomic_t *atom, uint32_t value)
+{
+    return atomic_add_return(value, (atomic_t*)atom) - value;
```
raphaelning: nit: `atomic_t *`

platforms/linux/hax_wrapper.c
```diff
@@ -0,0 +1,336 @@
...
+}
+
+hax_atomic_t hax_atomic_inc(volatile hax_atomic_t *atom)
+{
+    return atomic_inc_return((atomic_t*)atom) - 1;
```
raphaelning: nit: `atomic_t *`

platforms/linux/components.c
```diff
@@ -0,0 +1,496 @@
...
+        break;
+    }
+    case HAX_VCPU_SET_REGS: {
+        struct vcpu_state_t vc_state;
+        if (copy_from_user(&vc_state, argp, sizeof(vc_state))) {
+            ret = -EFAULT;
+            break;
+        }
+        ret = vcpu_set_regs(cvcpu, &vc_state);
+        // Due to debugging, vcpu_set_regs might modify drN registers
```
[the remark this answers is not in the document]
raphaelning: Just realized that I had referred you to the wrong QEMU-side patch... @junxiaoc will delete it and re-upload the correct one. Sorry!

platforms/linux/hax_wrapper.c
AlexAltea: Brief update: the approach above didn't work as the patch that documented it didn't get merged. However, I have found a very similar approach while browsing the kernel sources that relies on `va_format` struct and the `%pV` placeholder:

```
    struct va_format vaf;
    va_list args;

    // [...]

    vaf.fmt = fmt;
    vaf.va = &args;
    va_start(args, fmt);
    printk("%shaxm_debug: %pV", KERN_DEBUG, &vaf);
    va_end(args);
```

This, along all previous issues, have been fixed in ac265fd.

AlexAltea:
Turns out that Debian `debian-9.5.0-amd64-xfce-CD-1.iso` kpanics right after choosing "Graphical install" in the boot menu. The same crash happens on Windows hosts so it can hardly be a Linux-side issue.

This might be a regression caused by this patch, I'll investigate the issue further and report back.

platforms/linux/components.c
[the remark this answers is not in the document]
raphaelning: @AlexAltea This issue is not addressed by https://github.com/intel/haxm/pull/108/commits/ac265fd651b00f8c9e6ee9a8a808781d5f203d96, so I can't approve the PR yet. What's your take on it?

BTW, @junxiaoc has updated #81 with the correct QEMU patch.

raphaelning:
Thanks, the latest commit looks great, but I'm still waiting for your response to the `HAX_VCPU_SET_REGS` ioctl issue.

mborgerson:
I've tested this on an Ubuntu 18.04 machine and was able to boot a QEMU guest! Great work, @AlexAltea!

Unfortunately however, I did have some issues. First is that the device files are created without any permissions. This makes it a pain when launching QEMU as a regular user. I think the files should probably be created with the group being set to `haxm` or something, and owner/group perms being r/w. This is how KVM does it as well:
```
crw-rw---- 1 root kvm 10, 232 Nov  2 00:46 /dev/kvm
```
Second, I also experienced some system instability when using HAXM which caused a hard kernel hang and forced me to hard-reset my system. This behavior happened a few times, always after trying to launch another program while QEMU was running in the background. Unfortunately, I have not captured a proper trace to help track down the source of the hang. I would not suggest to block a merge for this one however.

platforms/linux/components.c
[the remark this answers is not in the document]
AlexAltea: Sorry, I had posponed this issue and ended up forgetting about it.

The reason I added `copy_to_user` was because I saw. But I agree this is completely counter-intuitive, since `SET_REGS` should never change the userland-provided struct. I'll remove those lines.

AlexAltea:
> I'm still waiting for your response to the HAX_VCPU_SET_REGS ioctl issue.

@raphaelning I have just removed those extra lines, at the `*SET_REGS` handler on 7062cce. That ioctl should **not** modify the `vcpu_state_t` struct. If that behavior breaks userland, then it's an issue on userland.

Before merging this branch, let me know so I can clean up the commit history a bit (if you deem it necessary).

---

>  I think the files should probably be created with the group being set to haxm or something, and owner/group perms being r/w. This is how KVM does it as well [...]

@mborgerson That's a good point, however this does not occur at kernel-level. In fact, on a Linux distro without QEMU, after manually creating */dev/kvm* via its major/minor identifiers, I obtain this information:
```bash
$ ls -l /dev/kvm
crw-r--r-- 1 root root 10, 232 Nov  2 09:54 /dev/kvm
```

It's only after installing QEMU that the appropriate information is created:
```bash
$ cat /etc/group | grep kvm
kvm:x:125:
$ cat /lib/udev/rules.d/40-qemu-system-common.rules 
KERNEL=="kvm", GROUP="kvm", MODE="0660"
$ ls -l /dev/kvm 
crw-rw---- 1 root kvm 10, 232 Nov  2 09:54 /dev/kvm
```

I understand that the QEMU/KVM relationship has "historical background", but I personally find that HAXM should not have any dependency on QEMU, despite it being its primary consumer: There's uses for hypervisors other than full-system emulation.

For now, I suggest that we handle group creation and udev rules with the Makefile.
- `make install`: Will install the *haxm.ko* kernel module, create a *haxm* group, and create an udev rule to assign the */dev/HAX* device to the *haxm* group.
- `make uninstall`; Will return */dev/HAX* to *root*, remove the udev rule, remove the *haxm* group and finally uninstall the *haxm* kernel module.

Note that Linux users won't automatically be added to the *haxm* group. They will need to do so manually via the following command:
```
sudo adduser `id -un` haxm
```
Afterwards, they will be able to launch `qemu-system-x86_64 -accel hax` as a regular user. All this information will be added to `docs/manual-linux.md`.

@raphaelning Sounds good?

---

> Second, I also experienced some system instability when using HAXM which caused a hard kernel hang and forced me to hard-reset my system.

@mborgerson I will investigate this issue further. Thank you!

mborgerson:
@AlexAltea thanks for looking into it! I wasn't aware that those perms on /dev/kvm were handled with a udev rule, but that makes sense. I think your suggestion regarding make [un-]install + documentation sounds good.

maronz:
I'm quite impressed with the progress this project has recently seen. So I was curious, whether I could build it myself.

One of my observations (i.e. that 'make clean' leaves most of the *.o files still in place) looks to have been addressed by 'dbfba2f'.

Another one is that it looks like that no attempt has been made to build this on a 32-bit Linux system. I guess the NASM output file format will have to be dependent on the architecture. Furthermore the compiler ran into heaps of errors related to the 'ASMCALL' and 'EMCALL' macros. Since this is not my area of expertise, I can only report my finding, and hope that a solution is not too difficult to be found.

AlexAltea:
> Another one is that it looks like that no attempt has been made to build this on a 32-bit Linux system.

@maronz Thanks a lot for your feedback! Indeed, 32-bit Linux hosts hadn't been tested yet.
After b6c43ac it has been fixed. As you mentioned, it required generating 32-bit objects with NASM and fixing some calling convention macros.

However, after successful installation of the kernel module, HAXM failed to run a 32-bit guest kernel, `debian-9.5.0-i386-xfce-CD-1.iso`. The issue seems quite different others I've seen so far (error happens inside `asm_vmxrun`, i.e. the CPU itself is reporting the failure), so maybe it's a nested virtualization bug of the underlying hypervisor, in my case VMware's one. These were the logs:
```
[   65.146688] haxm_debug: vm entry!
[   65.146948] haxm_debug: HAX: VM entry failed: result=2 RIP=0000fff0
[   65.146958] haxm_info: HAX: Prev exit: 0 error code: 7
[   65.146960] haxm_info: HAX: VM entry failed
[   65.146961] haxm_info: end of cpu_vmentry_failed
[   65.146963] haxm_debug: cpu_vmx_run error, code:2lx
```

If you are running a 32-bit Linux distribution "bare metal", I would be really helpful if you could re-test this patch at b6c43ac and let me know whether you can replicate the issue (or whether it works).

---

@raphaelning Right now, some logs don't feel helpful enough (see messages above). I often find myself asking what's `result=2`, or `error code: 7`, or `code:2lx`. Ironically, functions that do indeed provide helpful feedback for debugging purposes, are disabled and/or interleaved with unrelated code: 
- https://github.com/intel/haxm/blob/96af3d28/core/cpu.c#L133
- https://github.com/intel/haxm/blob/96af3d28/core/cpu.c#L285

Would it be reasonable to uncomment and move such helper functions to a single file (maybe even the `dump.c` file that already exists)? Additionally, we could add helper functions to transform *enum* constants into strings, to print the meaning of error codes directly in the logs. What are your thoughts?

I'm asking just for the sake of discussion. If accepted, this proposal should be a separate PR. :-)

platforms/linux/haxm-uninstall.sh
```diff
@@ -0,0 +1,19 @@
...
+DEVNAME=HAX
+GRPNAME=haxm
+MODNAME=haxm
+
+# Remove udev rule
+rm -f /lib/udev/rules.d/99-haxm.rules
+
+# Remove group, if necessary
+if [ $(getent group $GRPNAME) ]; then
+    groupdel -f $GRPNAME
```
raphaelning: I'm on Ubuntu 16.04, and my `groupdel` command does not accept the `-f` flag:

```
groupdel: invalid option -- 'f'
Usage: groupdel [options] GROUP
```

raphaelning:
@AlexAltea  The new commits look good, except that I haven't been able to get `qemu-system-x86_64 -accel hax` working on my Ubuntu 16.04:

```
Failed to open vm 0
Failed to create HAX VM
No accelerator found.
qemu-system-x86_64: failed to initialize HAX: Invalid argument
```

I'll look into the "invalid argument" error later. But if it rings a bell with you, please help :-)

I've already added my user to the `haxm` group, logged out and logged back in. Without the re-login step, I'd get a different error:

```
Failed to open the hax module
No accelerator found.
qemu-system-x86_64: failed to initialize HAX: No such device
```

I also noticed an error during `sudo make install`, but I'm not sure if that's related:

```
$ sudo make install
make -C /lib/modules/4.15.0-38-generic/build M=$PWD modules_install
make[1]: Entering directory '/usr/src/linux-headers-4.15.0-38-generic'
  INSTALL /home/sdkadmin/Code/haxm/platforms/linux/haxm.ko
At main.c:160:
- SSL error:02001002:system library:fopen:No such file or directory: bss_file.c:175
- SSL error:2006D080:BIO routines:BIO_new_file:no such file: bss_file.c:178
sign-file: certs/signing_key.pem: No such file or directory
  DEPMOD  4.15.0-38-generic
make[1]: Leaving directory '/usr/src/linux-headers-4.15.0-38-generic'
./haxm-install.sh
```

Lastly, I'm using the latest QEMU master, since it was pretty easy to rebase https://github.com/kryptoslogic/qemu/commit/48db0b4f0bcd952e09c3355c2b64948bf703da58 onto it. I'll try v2.12 later.

raphaelning:
> Afterwards, they will be able to launch qemu-system-x86_64 -accel hax as a regular user. All this information will be added to docs/manual-linux.md.
>
> @raphaelning Sounds good? Proposal at dbfba2f.

Yes, the install/uninstall scripts are very handy. Please update the Linux manual accordingly.

> Would it be reasonable to uncomment and move such helper functions to a single file (maybe even the dump.c file that already exists)? Additionally, we could add helper functions to transform enum constants into strings, to print the meaning of error codes directly in the logs. What are your thoughts?

Indeed there are quite a few problems with the HAXM fatal error dump/log, especially in terms of completeness and readability as you have pointed out. Both of your ideas should be great steps towards putting this right!

platforms/linux/components.c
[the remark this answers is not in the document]
raphaelning: > The reason I added copy_to_user was because I saw vcpu_set_regs modifying the state.

Ah, you must be referring to this part:

```
        ustate->_dr6 = fix_dr6(ustate->_dr6);
        UPDATE_VCPU_STATE(_dr6, dr_dirty);
        ustate->_dr7 = fix_dr7(ustate->_dr7);
        UPDATE_VCPU_STATE(_dr7, dr_dirty);
```

The purpose of `fix_dr6/fix_dr7` is to make sure we correctly set or clear all the reserved bits. Ideally we shouldn't modify the input `ustate`, but ended up doing that so we could keep using the `UPDATE_VCPU_STATE` macro for DR6 and DR7 as before.

maronz:
@AlexAltea thanks a lot for b6c43ac. I can confirm that I was able to successfully build the kernel module on a 32-bit "bare system". I don't actually use a "proper" installation, I just use the "live session" that comes ups when I boot the system (e.g. 'Linux Mint 19 Mate 32-bit') from a USB stick.

The HAXM driver installed, and I then started to "kick the tyres". Somehow, during this a bit unstructured testing I managed to produce a crash (as I could see in the 'dmesg' output). I then had the (stupid) idea to create a script, so that I'd be able to reproduce results this in a repeatable fashion. The script installs any additionally required packages, downloads the HAXM and QEMU sources, patches the latter, builds both and attempts to run a test.

Albeit now, when the script runs I did not see a crash, but rather essentially the same failure as @raphaelning
`mint@mint:~/qemu-3.0.0$ ./i386-softmmu/qemu-system-i386 -accel hax`
`Failed to open vm 5`
`Failed to create HAX VM`
`No accelerator found.`
`qemu-system-i386: failed to initialize HAX: Invalid argument`
`mint@mint:~/qemu-3.0.0$ ./i386-softmmu/qemu-system-i386 -accel hax`
`Failed to open vm 6`
`Failed to create HAX VM`
`No accelerator found.`
`qemu-system-i386: failed to initialize HAX: Invalid argument`
(note the increment in the VM number)

Operating in such a "live session" allows me also to do things that might not be quite "kosher" on a "proper" system. In that spirit I then ran: `sudo ./i386-softmmu/qemu-system-i386 -accel hax`
and that resulted in "crash" output in the 'dmesg', and a "hanging" process,  which showed up in 'ps -efH' output as:
`root     23641 15067  0 22:14 pts/1    00:00:00             sudo ./i386-softmmu/qemu-system-i386 -accel hax`
`root     23642 23641  0 22:14 pts/1    00:00:00               [qemu-system-i38] <defunct>`
and required a `sudo kill -9 23642` to get rid of it.

So, in summary: the HAXM module does not (yet) work for me on a (32-bit) Linux system.

PS: If anyone sees any benefit in it, I still have the script and the 'dmesg' output, and be happy to share it.

raphaelning:
I got the same "Failed to open vm" error after switching to https://github.com/kryptoslogic/qemu/commit/48db0b4f0bcd952e09c3355c2b64948bf703da58 (QEMU v2.12.1) :-( I haven't looked into it yet.

> If you deem it necessary, I can cleanup the commit history before merging this branch.

That would be ideal, if it's not too much burden for you.

raphaelning:
@maronz Ah interesting, so basically I just need to run QEMU with `sudo`? That's not ideal, but I'll give it a try tomorrow.

platforms/linux/haxm-uninstall.sh
AlexAltea: Turns out this flag was recently added, but it's not really necessary for us.
I will remove it in my next commit.

AlexAltea:
> The new commits look good, except that I haven't been able to get `qemu-system-x86_64 -accel hax` working on my Ubuntu 16.04:
> ```
> Failed to open vm 0
> Failed to create HAX VM
> ```

@raphaelning @maronz Sorry, the HAXM VM/VCPU devices hadn't been assigned to the *haxm* group, That's why only running as root worked. I'll patch the driver to update ownership information accordingly.

The incrementing VM number is definitely a consequence of this issue, but I'll still investigate it further to ensure there's no logic issue: VM's are destroyed once the reference counter, decremented by `hax_put_vm`, reaches zero, so lingering VM's are definitely bad news.

> I've already added my user to the haxm group, logged out and logged back in. Without the re-login step, I'd get a different error.

@raphaelning After adding the current user to any group, it's necessary to re-login. Otherwise, it cannot open */dev/HAX* which results in the error you see. I've added the corresponding information in the manual.
Alternatively, you can get a shell with updated groups without re-login via: `exec su -l $USER`.

> If anyone sees any benefit in it, I still have the script and the 'dmesg' output, and be happy to share it.

@maronz It could be useful. If a crash ever occurs, it would be really helpful to share a *dmesg* snippet (filtering out any sensitive information).

---

> I also noticed an error during `sudo make install`, but I'm not sure if that's related:
> ```
> sign-file: certs/signing_key.pem: No such file or directory
> ```

@raphaelning The reason is that the Makefile's `modules_install` target (dependency of `install` target), attempts to sign the kernel module with the mentioned key pair, which despite being automatically generated during kernel build, they are not shipped with Linux distros for obvious reasons.
This has resulted in [several issues in many drivers](https://github.com/search?q=sign-file%3A+certs%2Fsigning_key.pem%3A+No+such+file+or+directory&type=Issues).

If you don't have a key pair already, the solution is creating one manually:
```
cd /lib/modules/`uname -r`/build
```
Then, create the file: `certs/x509.genkey`, with the following contents (based on `certs/Makefile`):
```
[ req ]
default_bits = 4096
distinguished_name = req_distinguished_name
prompt = no
string_mask = utf8only
x509_extensions = myexts

[ req_distinguished_name ]
CN = Build time autogenerated kernel key

[ myexts ]
basicConstraints=critical,CA:FALSE
keyUsage=digitalSignature
subjectKeyIdentifier=hash
authorityKeyIdentifier=keyid
```
Finally, create the key pair with (based on `certs/Makefile` and `/boot/config-*` variables):
```
openssl req -new -nodes -utf8 -sha512 -days 36500 -batch -x509 -config certs/x509.genkey -outform PEM -out certs/signing_key.x509 -keyout certs/signing_key.pem
```
This should suffice to self-sign the kernel module. Of course, you will still get a "*PKCS#7 signature not signed with a trusted key*" warning in dmesg, unless the public key is trusted by the kernel.

I could add these steps to the manual, but right now Ubuntu 18.04 (and most distros?) haven't enabled a strict driver trusted signature enforcement yet. Even if enabled, for testing/debugging it's much simpler to disable it, for release/production it's much easier to give users pre-signed binaries, so I'm not sure if these lengthy instructions should be written in the manual.

raphaelning:
> @raphaelning @maronz Sorry, the HAXM VM/VCPU devices hadn't been assigned to the haxm group, That's why only running as root worked. I'll patch the driver to update ownership information accordingly.

I see, thanks! Indeed the lingering `/dev/hax_vm/vmXX` device nodes are owned by `root:root`. I checked the Darwin implementation, which calls `devfs_make_node()` to create these special files. That function conveniently asks for the owner, group and mode of the new file, and HAXM passes to it the current process's UID, GID and `0600`. This piece of logic is missing from the current Linux implementation, which seems to rely on `misc_register()` to create the device node. I hope you can find a way to specify ownership information, either in the driver or with udev rules. But we could also argue that creating a device node for each VM and vCPU is not the best design in the first place, because those should only be visible to one process (QEMU) and should not consume system-wide resources...

> The incrementing VM number is definitely a consequence of this issue, but I'll still investigate it further to ensure there's no logic issue: VM's are destroyed once the reference counter, decremented by hax_put_vm, reaches zero, so lingering VM's are definitely bad news.

I can confirm that running QEMU as root works :-) However, HAXM fails to clean up the VM and vCPU device nodes after I quit QEMU, so there's indeed a flaw in the reference counting mechanism. I think I've seen lingering device nodes on Mac as well, although very rarely (only when HAXM runs in to a fatal error). If this issue turns out to be outside the scope of this PR, we can merge this PR first.

raphaelning:
> I could add these steps to the manual, but right now Ubuntu 18.04 (and most distros?) haven't enabled a strict driver trusted signature enforcement yet. Even if enabled, for testing/debugging it's much simpler to disable it, for release/production it's much easier to give users pre-signed binaries, so I'm not sure if these lengthy instructions should be written in the manual.

Thanks for the detailed explanation! I see that signing the driver is not very useful at the moment, so let's not add these steps to the manual. However, since people might get confused by the `sign-file` error, how about simply noting in the manual that the error can be safely ignored?

maronz:
I'm back again with another round of observations. And as it is a new day, I've decided to change the Linux distribution for today's testing to be Knoppix v8.2 (with either a 4.16.5 or a 4.16.5-64 kernel). For the "price" of a single ISO file download, one gets the option to boot either into a 32 or a 64-bit system. [ I've now come to the impression, that this distribution is a combination of two different kernels, but always with a 32-bit userland ].

(1) My inability to use '-accel hax' on a 32-bit (bare metal) Linux system continues. But I've finally noticed a message in the 'dmesg' output, when running 'sudo make install', that might help in the problem identification:
`[  648.771322] haxm_error: NX is not enabled in the system, HAXM does not function.`
`[  648.771338] haxm_warning: -------- HAXM v7.3.2 Start --------`

Initially I ran this on a 'i5-520M' CPU, of which I was never quite sure, whether HAXM would be supported. But I repeated the test on a 'i5-4300U', and the result was the same.

(2) I then moved on to repeat my test with the 64-bit kernel. There I ran into the problem that Knoppix v8.2, when booted in 64-bit mode, reports the following slightly unexpected results:
`knoppix@Microknoppix:~$ getconf LONG_BIT ; uname -m`
`32`
`x86_64`

With this in mind, it is clear why 'haxm.ko' failed to be built, as NASM had produced 32-bit objects. I changed the architecture check in the 'Makefile' via:
`sed -i 's#LBITS#MACHINE# ; s#getconf LONG_BIT#uname -m# ; s#,64#,x86_64#' Makefile`

This seems to have done the trick, and maybe that is a more generic method to determine the processor architecture. After the kernel module got built, the 'sudo make install' did not cause another message in the 'dmesg' output, and apart from the (current) need to run QEMU as super-user I was able to use '-accel hax' successfully (even on the old i5-520M).

AlexAltea:
> This piece of logic is missing from the current Linux implementation, which seems to rely on misc_register() to create the device node.

@raphaelning Indeed. I'll find a way around that and report back once I fix the issue.

> But we could also argue that creating a device node for each VM and vCPU is not the best design in the first place, because those should only be visible to one process (QEMU) and should not consume system-wide resources...

Agreed. Probably creating an anonymous file instance and returning the file descriptor to userland would work: it limits VM/VCPU access to QEMU. That's `anon_inode_getfd` on Linux. Hopefully, there's something similar in Windows/macOS. Anyway, this should be left for future patches since it's a big API change.

> I think I've seen lingering device nodes on Mac as well, although very rarely

I might have found (part of?) the root cause while working on these patches since the code came straight from the haxm-darwin frontend. Some ioctl's exit without decrementing the reference counter (via `hax_put_vcpu` and `hax_put_vm`):
- https://github.com/intel/haxm/blob/592f97a4/platforms/darwin/com_intel_hax_ui.c#L166
- https://github.com/intel/haxm/blob/592f97a4/platforms/darwin/com_intel_hax_ui.c#L189
- https://github.com/intel/haxm/blob/592f97a4/platforms/darwin/com_intel_hax_ui.c#L399
- https://github.com/intel/haxm/blob/592f97a4/platforms/darwin/com_intel_hax_ui.c#L421
- https://github.com/intel/haxm/blob/592f97a4/platforms/darwin/com_intel_hax_ui.c#L432

AlexAltea:
> There I ran into the problem that Knoppix v8.2, when booted in 64-bit mode, reports the following slightly unexpected results:

@maronz That's surprising. Thanks for providing the Makefile patch, that seems to be more reliable.

> Quick question that I've missed to ask: which Linux distribution have you used for your testing of a 32-bit host?

I used Ubuntu 14.04, specifically [`ubuntu-14.04.5-desktop-i386.iso`](http://releases.ubuntu.com/14.04/ubuntu-14.04.5-desktop-i386.iso). Let me know you will see any noticeable difference with it. :-)

AlexAltea:
Last issue fixed: Similarly to *haxm-darwin*, newly created VM/VCPU devices will be owned by the UID:GID of the calling process. This is done by looking-up the inode for the device path and modifying its attributes. The solution isn't ideal, but as mentioned earlier, it's the best we can do without API changes.

Current status of HAXM:
- 64-bit Linux hosts: builds and runs fine. No lingering issues.
- 32-bit Linux hosts: builds, but fails at VM-entry. I need extra help/time to fix it (info at https://github.com/intel/haxm/pull/108#issuecomment-435691965).

include/hax.h
```diff
@@ -232,15 +230,16 @@ static inline unsigned char *hax_page_va(struct hax_page *page)
...
 
 #ifdef __cplusplus
 extern "C" {
 #endif
 
-int smp_call_function(cpumap_t *cpus, void(*scfunc)(void *param), void *param);
-extern int cpu_number(void);
+int hax_smp_call_function(hax_cpumap_t *cpus, void(*scfunc)(void *param),
+                          void *param);
+extern int hax_cpu_number(void);
```
raphaelning: We got the following error on Mac when loading the kext:

```
(kernel) kxld[com.intel.kext.intelhaxm]:        _hax_cpu_number
(kernel) Failed to load executable for kext com.intel.kext.intelhaxm.
(kernel) Kext com.intel.kext.intelhaxm failed to load (0xdc008016).
(kernel) Failed to load kext com.intel.kext.intelhaxm (error 0xdc008016).
```

I think `cpu_number()` is probably a function provided by the Darwin kernel. Also, I don't see `hax_cpu_number()` used outside of `platforms/darwin/`, so I'd:

1. Remove this declaration the common header.
2. Keep using `cpu_number()` for Darwin code.

maronz:
The following has been observed (for a change) on a 64-bit Linux Mint 19, with a 4.15.0 kernel.

The '/dev/HAX' device is still getting created with 0600 permissions, and 'root:root' ownership, but correcting those (via: `sudo chmod 660 /dev/HAX && sudo chown root:haxm /dev/HAX`) allows to successfully start VMs with 'accel hax' without the use of 'sudo'.

After I've have started 8 VMs I'm out of luck, and I'm getting:
`Failed to create vm ffffffff`
`Failed to create HAX VM`
`No accelerator found.`
`qemu-system-i386: failed to initialize HAX: Invalid argument`

Even `sudo make uninstall && sudo make install && sudo rm -r /dev/hax_vm*` does not help, and any further attempt to start a VM ends up with:
`sysfs: cannot create duplicate filename '/devices/virtual/misc/hax_vm!vm00'`
 in the 'dmesg' output (followed by some process dump).

AlexAltea:
> The '/dev/HAX' device is still getting created with 0600 permissions, and 'root:root' ownership, but correcting those (via: `sudo chmod 660 /dev/HAX` && `sudo chown root:haxm /dev/HAX`)

@maronz That's weird. Did that happen even after running `sudo make install`? Because the installation script will run those two `chmod` + `chown` commands. I can't replicate this issue.

> After I've have started 8 VMs I'm out of luck, and I'm getting

This issue most likely has to do with the incrementing VM ID problem. I'll fix that one now.

include/hax.h
AlexAltea: Fixed on f153b54.

AlexAltea:
Linking issues on macOS are fixed as of f153b54. Then I rebased on top of *master*, which required fixing the Kbuild script on 12e4821.

> Good catch! All of these happen only in extreme circumstances. I may have run into one of them on Mac--I'm not sure

Should I submit a PR to add those mising `hax_put_vcpu` and `hax_put_vm` calls?

As for the lingering device issues on Linux, I'll investigate that now, as well as the reason why the VM/VCPU's are shown in red: it seems it's detecting the file with an unknown type (note the `?` in `?rw-rw----`), maybe a side-effect of messing with the inode directly.

maronz:
@AlexAltea  Regarding the ownership/permission problem I've now found something of a work-around: after adding to 'haxm-install.sh' a final line of `ls -l /dev/$DEVNAME` the problem seems to "disappear" (at least on the one system, I've now done many iterations of testing).

Another option that appears to work is inserting a `sleep 1` before the 'chown' command. If one puts that delay in between the 'chown' and the 'chmod' the latter gets done, but not the former. This last observation indicates to me, that some weird race condition might be in play here. I'm probably as dumbfounded as yourself, as I can't see anything wrong in the original script,

Anyway, the 'ls -l ...' is probably more useful, as one would get to see right at the end of the install script whether the device creation has worked, or not.

AlexAltea:
@maronz Indeed, it sounds like a race condition. Maybe `modprobe` does return right after loading the module, but does not way for the initialization function to finish, thus `chmod`+`chown` might be called before the HAXM device exists.

I've updated the installation script to wait up to 3 seconds (if necessary), which should be way more than enough time for the device to be created. Let me know if this change fixes your issue.

AlexAltea:
@raphaelning Regarding the 32-bit Linux host issues I found: Now with proper logging I see:
```
[   73.853623] haxm_warning: -------- HAXM v7.3.2 Start --------
[   98.103785] haxm_warning: hax_alloc_pages: HAX_MEM_LOW_4G is ignored
[   98.559302] haxm_error: VM entry failed: RIP=0000fff0
[   98.559311] haxm_error: VMfailValid. Prev exit: 0. Error code: 7 (VMX_ERROR_ENTRY_CTRL_FIELDS_INVALID)
```

The Intel SDM Vol. 3C claims at *30.4 VM Instruction Error Numbers* that this code means:
> VM entry with invalid control field(s) *b*,*c*
> - **b**. VM-entry checks on control fields and host-state fields may be performed in any order. Thus, an indication by error number of one cause does not imply that there are not also other errors. Different processors may give different error numbers for the same VMCS.
> - **c**. Error number 7 is not used for VM entries that return from SMM that fail due to invalid VM-execution control fields in the executive VMCS. Error number 25 is used for these cases.

Any clues as to debug *which* control field(s) might be invalid?

maronz:
@AlexAltea I'm afraid, I owe you an apology. Before my last comment (about the race condition in the install script) I did not read the script carefully enough. [ I guess it being late in the evening at my end did not help things. ]

After having done a quick test of a0b270b, and still (!!) noticing the permission/ownership issue, I've had another look at the script, and have now come to the conclusion that we've got a "cart before the horse" situation. I'm now suggesting to re-arrange the order of the critical steps into the following order:

1. create the group (if necessary),
2. create the 'udev' rule, and
3. then load the kernel module

With this sequence of events I'm now confident that any attempt of "fixing" the permission/ownership afterwards won't be required. The current order of processing is: 3,1,2 (i.e. load the module first, then set up the group and the 'udev' rule). And that seems to explains (in hindsight) what has been observed up to now.

I'm attaching a proposal of a re-worked 'haxm-install.sh' script for your perusal.
[haxm-install.sh.gz](https://github.com/intel/haxm/files/2559158/haxm-install.sh.gz)

AlexAltea:
@maronz Thanks a lot for tracking down the issue! I've added your patched file and amended the commit.

include/linux/hax_interface_linux.h
```diff
@@ -0,0 +1,82 @@
...
+
+/* API 2.0 */
+#define HAX_VM_IOCTL_NOTIFY_QEMU_VERSION _IOW(0, 0x84, struct hax_qemu_version)
+
+#define HAX_IOCTL_VCPU_DEBUG _IOW(0, 0xc9, struct hax_debug_t)
+
+#define HAX_KERNEL64_CS 0x80
+#define HAX_KERNEL32_CS 0x08
+#ifdef __i386__
+#define is_compatible() 1
```
raphaelning: This must be the root cause of the VM entry failure on 32-bit Linux hosts. The code is apparently taken from `hax_interface_mac.h`, but probably we should follow the example of `hax_interface_windows.h` instead:

`#define is_compatible() 0`

I don't know why Darwin i386 needs it to be 1, but it's no longer a valid platform for HAXM. We can actually get rid of `is_compatible()` and all the code paths where `is_compatible() == 1`.

raphaelning:
> After diffing the files, aside from many expected differences in HOST_* fields I've noticed a mismatch in VMX_EXIT_CONTROLS

Thanks for doing all the hard work to track this issue down to this simple diff! Apparently, the Linux driver incorrectly sets bit 9 (Host address-space size) and bit 21 (Load IA32_EFER), both which should be 0 for a 32-bit host OS (especially when the host kernel is 32-bit). With this information, it's very easy to identify the code that sets these bits:

https://github.com/intel/haxm/blob/24fdff5da8a15347463bfe12e35da56fbdd74cef/core/vcpu.c#L1360-L1367

As I've pointed out in the above comment, the solution is to make sure `is_compatible()` evaluates to 0 on Linux hosts.

AlexAltea:
@raphaelning Thank you! At first, I was sceptical about `is_compatible()` being the culprit so I never bother changing it and testing. But that fixed indeed the problem. :-)

Now with 64-bit and 32-bit Linux hosts on the same grounds I'm going to fix the lingering VM/VCPU issues, after some checking I figured out I didn't implement *open* and *release* (i.e. close) file operations for them. This should fix the issue, but during `hax_clear_vcpumem(cv->tunnel_vcpumem);` there's a kernel panic happening, caused by *vm_munmap*.

I'll investigate why this happens and report back.

raphaelning:
The new commits look good to me, but the lingering device nodes issue is somehow still not completely fixed:

(Before the test, I rebooted my computer and confirmed that `/dev/hax*` didn't exist.)

1. The first QEMU VM creates `/dev/hax_vm/vm00` and `/dev/hax_vm00/`, but does not delete them in the end.
1. The second VM somehow manages to "reuse" `vm00` instead of creating `vm01`. In the end, it doesn't delete `vm00` either.
1. The third and fourth VMs are run in parallel. One of them reuses `vm00`, and the other creates `vm01`. In the end, we have both `vm00` and `vm01` nodes lingering around.

BTW, we have tested this PR (in its current state) on Windows and Mac without running into any issue, so we're ready to merge it. But if you want to add any Linux-specific patches, we can wait.

AlexAltea:
@raphaelning Great! None of the remaining issues require changing *core* or Windows/Mac-specific code, so the tests won't be needed again. Once everything is ready from my side, I'll rewrite the commit history as:
* Normalized prefixes of functions/types
* Added support for Linux hosts
* Added Linux to TravisCI build matrix
* Manual for building/testing HAXM on Linux

---

> the lingering device nodes issue is somehow still not completely fixed

This seems caused by the inode modification, which also caused the devices to be shown in red with unknown type `?` (permissions `?rw-rw----`). Reverting ceed27e fixes both issues, but makes `sudo` necessary again. I'll investigate and fix this.

raphaelning:
@AlexAltea Sounds good, thanks! My previous comment coincided with https://github.com/intel/haxm/pull/108/commits/8f5956448c786b579e2a9ed9171444ecb609c47e, but I was testing without this commit. It looks like an important fix though.

raphaelning:
I can confirm that https://github.com/intel/haxm/pull/108/commits/edcd0ef09ebf11687199117f827dc7d9d9f78d0c has eradicated lingering VM/vCPU device nodes, and those device nodes now show up with the correct file type (`c`) in `ls -l` output. Awesome!

BTW, in the final patch set, I'd like to see https://github.com/intel/haxm/pull/108/commits/d673697fbdb342edc4bc9d76d006e51493637872 ("Ensuring all VM/VCPU ioctl exit paths decrement the refcounter") remain a separate commit.

maronz:
@AlexAltea Congratulations!! The last commit (edcd0ef) appears to have really done the trick.

I've been (silently) following the progress over the course of the day, and even on a 32-bit host the '/dev/hax_vm*/vcpu*' entries are appearing, as a new VM gets started, and disappearing (!!), as a VM run comes to an end.

AlexAltea:
@raphaelning Sure thing! I'll insert that one as a separate commit right after "*Added support for Linux hosts*". If you want, I can submit an identical patch for MacOS in another PR.

The only thing pending, is fixing the *vm_munmap* issue mentioned at https://github.com/intel/haxm/pull/108#issuecomment-436857670. However, since it's commented out, it doesn't cause any problems for users: the only drawback is that we cannot unmap the "HAXM tunnel" page from userland. But at that point our main consumer (QEMU) is already finished, so exiting the process will free that page.

With this in mind, I'd say we could proceed to merge (after rewriting the commit history). Sounds good?
Of course, I'll keep providing support for the Linux platform, and fixing *vm_munmap* as well as maintaining the rest of the code will have top priority for me now.

AlexAltea:
All done!

raphaelning:
> If you want, I can submit an identical patch for MacOS in another PR.

That would be great, thanks!

There's a small problem with the latest PR, though:

```
$ ll platforms/linux/*.sh
-rw-rw-r-- 1 myuser   myuser   441 Nov  8 16:11 platforms/linux/haxm-install.sh
-rw-rw-r-- 1 myuser   myuser   375 Nov  8 16:11 platforms/linux/haxm-uninstall.sh
```

As you can see, these files are not executable, so `sudo make install` gives me this error:

```
./haxm-install.sh
make: execvp: ./haxm-install.sh: Permission denied
Makefile:35: recipe for target 'install' failed
make: *** [install] Error 127
```

Could you fix this in https://github.com/intel/haxm/pull/108/commits/a99ba167aa99ba760cc525bfd66f3c6d9fc14ec9 and force push?

krytarowski:
Great work!

PR #110, opened 2018-10-19
Codebase cleanup and portability refactoring

AlexAltea:
This patch allows for easier portability of HAXM to other platforms (see #108).
Builds successfully under Windows (32-bit and 64-bit configurations) and MacOS.

## Changes

Main changes are:

- **Normalized casing**: `s/ASSERT/assert`, `s/TRUE/true`, `s/FALSE/false`, since it was inconsistent (both uppercase and lowercase variants were used across the codebase).
- **Normalized integer types**: The codebase mixes (u)intN_t and (u)intN. This commit picks the former variant and consistently use it. Unnecessary typedef's to the latter variant are removed.
- **Fixed missing declarations**: Defined helper functions and macros to fix errors caused by undeclared non-standard functions, e.g.: `min`, `max`.
- **Generalized page-related constants**: Platform-specific headers, e.g. `hax_windows.h` defined page-related constants such as `page_size` and `page_shift` that are identical in all systems since they are unique to the x86 architecture. They were moved to`hax_types.h`.
- **Custom arch/compiler/platform macros**: Created unique `HAX_ARCH_*`, `HAX_COMPILER_*`, `HAX_PLATFORM_*` macros: The goal is avoiding dependance in compiler/platform-specific macros throughout the codebase by creating custom macros declared at a single location.
- **Fixed function prototypes**: Changing `function()` to `function(void)`, fixing `-Wstrict-prototypes`.

Full information in the messages of the individual commits.

### DIFF
--- a/core/include/vmx.h
+++ b/core/include/vmx.h
@@ -502,7 +502,7 @@ enum {
     GAS_CSTATE      = 4
 };
 
-#ifdef __WINNT__
+#ifdef HAX_COMPILER_MSVC
 #pragma pack(push, 1)
 #endif
 
@@ -592,7 +592,7 @@ struct PACKED info_t {
     uint64_t             _ept_cap;
 };
 
-#ifdef __WINNT__
+#ifdef HAX_COMPILER_MSVC
 #pragma pack(pop)
 #endif
 
@@ -670,7 +670,7 @@ void vmx_vmwrite(struct vcpu_t *vcpu, const char *name,
         ((val).base  = vmread(vcpu, GUEST_##desc##_BASE),          \
          (val).limit = vmread(vcpu, GUEST_##desc##_LIMIT))
 
-#if defined(__WINNT__)
+#if defined(HAX_PLATFORM_WINDOWS)
 #define VMWRITE_SEG(vcpu, seg, val) {                              \
             uint32_t tmp_ar = val.ar;                              \
             if (tmp_ar == 0)                                       \
@@ -681,7 +681,7 @@ void vmx_vmwrite(struct vcpu_t *vcpu, const char *name,
             vmwrite(vcpu, GUEST_##seg##_AR, tmp_ar);               \
         }
 
-#elif defined(__MACH__)
+#elif defined(HAX_PLATFORM_DARWIN)
 #define VMWRITE_SEG(vcpu, seg, val) ({                             \
             uint32_t tmp_ar = val.ar;                              \
             if (tmp_ar == 0)                                       \
--- a/core/include/vmx.h
+++ b/core/include/vmx.h
@@ -213,65 +213,65 @@ typedef enum vmx_error_t {
 
 // Exit qualification 64-bit OK
 union exit_qualification_t {
-    uint64 raw;
-    uint64 address;
+    uint64_t raw;
+    uint64_t address;
     struct {
-        uint32 size        : 3;
-        uint32 direction   : 1;
-        uint32 string      : 1;
-        uint32 rep         : 1;
-        uint32 encoding    : 1;
-        uint32 rsv1        : 9;
-        uint32 port        : 16;
+        uint32_t size        : 3;
+        uint32_t direction   : 1;
+        uint32_t string      : 1;
+        uint32_t rep         : 1;
+        uint32_t encoding    : 1;
+        uint32_t rsv1        : 9;
+        uint32_t port        : 16;
     } io;
     struct {
-        uint32 creg        : 4;
-        uint32 type        : 2;
-        uint32 rsv2        : 2;
-        uint32 gpr         : 4;
-        uint32 rsv3        : 4;
-        uint32 lmsw_source : 16;
+        uint32_t creg        : 4;
+        uint32_t type        : 2;
+        uint32_t rsv2        : 2;
+        uint32_t gpr         : 4;
+        uint32_t rsv3        : 4;
+        uint32_t lmsw_source : 16;
     } cr;
     struct {
-        uint32 dreg        : 3;
-        uint32 rsv4        : 1;
-        uint32 direction   : 1;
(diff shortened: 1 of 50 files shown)


AlexAltea:
I think I've found a bug in [core/memory.c:332-344](https://github.com/intel/haxm/blob/a12f70423b854305cc7a426bef175fb94f8f7fee/core/memory.c#L332-L344):

```c
#if defined(__MACH__)
#ifdef __x86_64__
            hva = (uint64_t)pmem->kva + (cur_va - pmem->uva);
#else
            hva = (uint64_t)(uint32_t)pmem->kva + (cur_va - pmem->uva);
#endif
#else   // __MACH
#if defined(_WIN64)
            hva = (uint64_t)pmem->kva + (cur_va - pmem->uva);
#else
            hva = 0;
#endif
#endif
```

Note how Windows (32-bit) sets `hva` to 0, while ironically MacOS (32-bit! which doesn't even exist) sets it to a non-zero value. Probably these two lines should be swapped. This probably hasn't affected end users since this only triggers in the absence of `CONFIG_HAX_EPT2`.

If it's a bug, I could easily fix it with this (which show the usefulness of the new HAX_ARCH* macros):

```c
#ifdef HAX_ARCH_X86_64
            hva = (uint64_t)pmem->kva + (cur_va - pmem->uva);
#else
            hva = (uint64_t)(uint32_t)pmem->kva + (cur_va - pmem->uva);
#endif
```

include/hax_types.h
```diff
@@ -31,182 +31,52 @@
...
-};
-
-typedef enum component_index_t component_index_t;
+/* Detect architecture */
+// x86 (32-bit)
+#if defined(__i386__) || defined(_M_IX86)
+#define HAX_ARCH_X86_32
+#define ASMCALL __cdecl
+#endif
+// x86 (64-bit)
```
raphaelning: nit: Just in case we run into a compiler that defines things differently, how about making this code more robust:

```
#if defined(__i386__) || defined(_M_IX86)
// 32-bit x86
#elif defined(__x86_64__) || defined(_M_X64)
// 64-bit x86
#else
#error "Unknown target CPU architecture"
#endif
```

include/hax_types.h
```diff
@@ -31,182 +31,52 @@
...
+// Clang
+#ifdef __clang__
+#define HAX_COMPILER_CLANG
+#define PACKED     __attribute__ ((packed))
+#define ALIGNED(x) __attribute__ ((aligned(x)))
+#endif
+// MSVC
+#ifdef _MSC_VER
+#define HAX_COMPILER_MSVC
+#define PACKED
```
raphaelning: Apparently MSVC doesn't have a simple equivalent for PACKED. Could you add a FIXME comment for this?

include/hax_types.h
```diff
@@ -31,182 +31,52 @@
...
+#define ALIGNED(x) __declspec(align(x))
+#endif
 
+/* Detect platform */
+// MacOS
 #ifdef __MACH__
+#define HAX_PLATFORM_DARWIN
```
raphaelning: Now that we have `HAX_PLATFORM_DARWIN`, shall we get rid of all usages of `__MACH__`?

include/hax_types.h
```diff
@@ -31,182 +31,52 @@
...
 #include "darwin/hax_types_mac.h"
 #endif
-
+// Windows
 #ifdef __WINNT__
+#define HAX_PLATFORM_WINDOWS
```
raphaelning: Same here. Shall we get rid of all usages of `__WINNT__`? In addition, there are also `_WIN64` and `_WIN32`, which can be replaced with a combination of `HAX_PLATFORM_WINDOWS` and `HAX_ARCH_X86_*`.

AlexAltea: > Same here. Shall we get rid of all usages of __WINNT__?

Yes.

> In addition, there are also _WIN64 and _WIN32, which can be replaced with a combination of HAX_PLATFORM_WINDOWS and HAX_ARCH_X86_*.

Correct. The latest commit applies this. Note that I have left platform-specific code unchaged, i.e. the `_WIN64` in `windows/hax_mm.h` precisely because those platform-specific macros are always present in that platform.

include/hax_types.h
AlexAltea: Yes, That's the intended purpose. The latest commit applies this.

AlexAltea:
All fixes applied!

>> Regex: "int([0-9]+)(?![_\d])" -> "int$1_t" (Mostly accurate).
>
> Does "mostly" imply that there were exceptions? Just curious.

Yeah. The regex rule correctly replaced everything, but it accidentally hit `SInt32` since the Visual Studio IDE editor does case-insensitive replacements by default (even in regex mode). Toggling case sensitivity fixes it. Additionally, the rule isn't limited to just `intN` and `uintN` but will also replace `foointN` to `foointN_t`.

That note in the commit message is a mere reminder of what I did, and also a reminder to double-check if used ever again. :-)

include/hax_types.h
raphaelning: I see that you've done the same for the compiler and platform checks, thanks!

core/include/vmx.h
```diff
@@ -681,7 +681,7 @@ void vmx_vmwrite(struct vcpu_t *vcpu, const char *name,
             vmwrite(vcpu, GUEST_##seg##_AR, tmp_ar);               \
         }
 
-#elif defined(__MACH__)
+#elif defined(HAX_PLATFORM_DARWIN)
```
raphaelning: I guess this is probably to workaround a compiler issue (MSVC vs. Clang), because the Mac code differs from the Windows code only in an extra pair of parentheses. But since we've finished testing the PR, you can address this in a later patch when you add Linux support.

PR #115, opened 2018-10-24
Added TravisCI scripts

AlexAltea:
Based on the discussion at https://github.com/intel/haxm/pull/108#issuecomment-430210463, I've prepared a TravisCI configuration script for HAXM.

This should be extremely helpful to review pull requests. For now it only checks for successful builds, but we can extend it by checking the emulator unit tests, or even integration tests, such as launching a VM (assuming the hardware/hypervisor used in TravisCI servers allows it).

Windows support is very recent (barely two weeks!), and right now they support only Visual Studio 2017 build tools, which apparently is not enough to build HAXM. However I've tested the script locally, and it should work once they provide a proper Visual Studio installation, which is likely. If you want to remove the Windows script in the meantime, let me know (I can add it as a comment to this PR until its ready).

EDIT: It seems https://github.com/intel is already integrated with TravisCI (see badges below). That really helps, and totally makes this choice even more convenient than alternatives (e.g. AppVeyor).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,29 @@
+language: c
+
+matrix:
+  include:
+    - name: "haxm-darwin"
+      os: osx
+      osx_image: xcode10
+      addons:
+        homebrew:
+          packages:
+            - nasm
+      script:
+        - cd platforms/darwin
+        - xcodebuild -configuration Debug -sdk macosx10.14
+
+  exclude: # TODO: Currently TravisCI does not support full VS/EWDK on Windows
+    - name: "haxm-windows"
+      os: windows
+      install:
+        - choco install -y nuget.commandline
+        - choco install -y windowsdriverkit10
+      script:
+        - cd platforms/windows
+        - nuget restore
+        - export PATH="$PATH:/c/Program Files (x86)/Microsoft Visual Studio/Installer/"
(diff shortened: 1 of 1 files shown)


raphaelning:
Thanks a lot! Indeed this is a great help.

The code looks nice and simple. We can keep the Windows script, but how can we test it locally? It doesn't seem to be written in Batch syntax.

AlexAltea:
> We can keep the Windows script, but how can we test it locally? It doesn't seem to be written in Batch syntax.

According to the [TravisCI Windows reference](https://docs.travis-ci.com/user/reference/windows/), the Windows script is executed within [Git BASH](https://gitforwindows.org/) (which I am thankful for, considering how terrible *cmd.exe* is).

Their current environment is essentially a stock Windows Server 2016 installation, with [Chocolatey](https://chocolatey.org/) and these [preinstalled packages](https://docs.travis-ci.com/user/reference/windows/#pre-installed-packages), which include Git BASH. Testing locally, using Windows 10 instead and Visual Studio 2017, that script works for me.

AlexAltea:
All done!

raphaelning:
Cool! I can see in the Travis build log that the macOS build is working :-)

PR #117, opened 2018-10-31
avoid unnecessary load_vmcs calls in VM exit

junxiaoc:
vmexit handler often needs some VMCS information such as guest
rip, physical address which causes vmexit. To read these
fields from VMCS or to write them back to VMCS, it needs to call
VMREAD and VMWRITE instructions. These two instructions require
the entire VMCS region to be loaded into memory, reading or
writing a VMCS field via vmread() and vmwrite() may incur calls
to load_vmcs() and put_vmcs(), which can have a non-negligible
overhead. If the VM exit handler needs to read/write multiple
VMCS fields, it can save time by merging all the load_vmcs()
calls as well as the put_vmcs() calls.

This change is to optimize vmexit handler performance by moving
frequently used VMCS fields vmread/vmwrite to "vmcs loaded"
context. For VMCS fields which have both read and write access,
a dirty flag is added to mark it should be write back or not
during loading guest state.

In addition, removed unused advance_rip_step function.

Signed-off-by: Junxiao Chang <junxiao.chang@intel.com>

### DIFF
--- a/core/cpu.c
+++ b/core/cpu.c
@@ -480,6 +480,10 @@ int cpu_vmx_execute(struct vcpu_t *vcpu, struct hax_tunnel *htun)
                 vcpu, VM_EXIT_INFO_IDT_VECTORING);
         vmx(vcpu, exit_instr_length) = vmread(
                 vcpu, VM_EXIT_INFO_INSTRUCTION_LENGTH);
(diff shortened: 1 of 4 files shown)


raphaelning:
For the record, this patch had passed our internal review before it was uploaded here.

PR #119, opened 2018-11-06
Refactored dumping and logging code

AlexAltea:
As discussed in https://github.com/intel/haxm/pull/108#issuecomment-435691965, there's few improvements that seemed to me a step in the right direction.
Let me know your thoughts on the matter.

## Changes

- **Refactored VMCS dumping code**: The previous implementation of `dump_vmcs` obtained the VMCS component name by hashing its value with a gperf-generated perfect hash function at *vmcs_names.c*. These compact hashes were used as indices in a manually-initialized table of names. This results in lots of boilerplate code, manual initialization/deinitialization and a dependency in gperf.

   This new approach uses a switch/case to dispatch the component names and compactifies the component encoding with the macro `HASH` to "encourage" the compiler to implement the switch/case via jump tables. The approach works in GCC, Clang, MSVC ([check Godbolt snippet](https://godbolt.org/z/-davh9)). Advantages:

   - Less code.
   - No external dependencies.
   - No memory allocated at runtime.
   - Better performance: No need to query an extra table to compute the perfect hash.
   - Perfect hashing tested at build-time: If the hash macro maps two components to the same value, the file won't compile.

* Centralized all dumping-related functions in *dump.c* (renamed from *dump_vmcs.c*). Not sure if this is the best location, but it probably is better than having inlined snippets (that are either commented out or disabled via preprocessor). More functions could be added to this file, e.g. dumping segment selectors.

* Created *name.c* to handle all enum-to-string conversions via functions prefixed with `name_`. For now I've handled just VMX error and exit codes (aside from the previous VMCS components), but more codes could be added there if required.

* **Improved old logging messages**: It looks as if certain log messages (in particular `hax_debug`/`hax_log` ones), had been written long time ago and thus do not fully comply with most of the codebase. Changes done:
   - Removed unnecessary "HAX: " prefix from certain calls to log functions. This was most likely a leftover from older HAXM revisions, since nowadays log functions add automatically the corresponding prefix.
   - Promoted some `hax_debug`/`hax_log` to `hax_error` in the case of actual errors, e.g. in `cpu_vmentry_failed`.

---

@raphaelning Is there any reason to preserve `hax_log` functions as-is instead of promoting them directly to `hax_info`? At the moment they are macro'ed to `hax_info`: Was this intentional, e.g. to change the macro during as a "debugging hack during development", or can we just remove the macro?

If you think we could get rid of this macro, I can amend the last commit and force-push the change.


core/dump.c
```diff
@@ -340,3 +340,70 @@ void dump_vmcs(struct vcpu_t *vcpu)
...
     }
 }
+
+void dump_vmx_info(struct info_t *info)
+{
+    hax_log("VMCS Rev %d\n", info->_vmcs_revision_id);
+
+    hax_log("VMX basic info       : 0x%016llX\n",
+            info->_basic_info);
+    hax_log("VMX misc info        : 0x%08llX\n",
```
raphaelning: Would `08` truncate the upper 32 bits? According to Intel SDM:

> Bits 63:32 report the 32-bit MSEG revision identifier used by the processor.

Not sure how useful this information is, but `016` would be a safe choice.

core/dump.c
```diff
@@ -340,3 +340,70 @@ void dump_vmcs(struct vcpu_t *vcpu)
...
+            info->_miscellaneous);
+    hax_log("VMX revision control : %d\n",
```
raphaelning: nit: `%d` => `%u`

core/name.c
```diff
@@ -0,0 +1,290 @@
...
+ */
+
+#include "include/name.h"
+#include "include/vmx.h"
+
+const char *name_vmcs_component(int value)
+{
+    // Intel SDM Vol. 3C: Table 24-17. Structure of VMCS Component Encoding
+#define HASH(x) \
+    (((x) & 0x003E) >> 1) /* Index */ | \
```
raphaelning: nit: It's a brilliant idea to omit the redundant Access Type bit as well as the upper 4 bits of Index, but the latter took me some time to confirm... How about adding a brief explanation, something like:

> As of this writing, no VMCS component is encoded with an Index of more than 5 bits (Intel SDM Vol. 3D, Appendix B: Field Encoding in VMCS), even though 9 bits are allocated to Index.

Also nit: Perhaps you could replace the 5 with a macro (e.g. `#define VMCS_ENC_INDEX_BITS 5`). That way, in case we ever need a longer Index, we just need to update a single line of code.

core/name.c
```diff
@@ -0,0 +1,290 @@
...
+    (((x) & 0x0C00) >> 4) /* Type  */ | \
```
raphaelning: nit: This hashing scheme (in particular `>> 4`) seems to waste one bit (bit 5):

```
 bit: 98 7654 3210
hash: ww TT0i iiii
```

(where: w = width, T = type, i = index)

platforms/windows/haxm-core.vcxproj
```diff
@@ -114,7 +114,7 @@
   </ItemDefinitionGroup>
   <ItemGroup>
     <!-- We only add items (e.g. form ClSourceFiles) that do not already exist (e.g in the ClCompile list), this avoids duplication -->
-    <ClCompile Include="@(ClSourceFiles)" Exclude="@(ClCompile)" />
```
raphaelning: nit: If you move this line, you may also want to move the comment above it.

core/cpu.c
```diff
@@ -643,22 +644,22 @@ void load_vmcs_common(struct vcpu_t *vcpu)
...
 
     //dump_vmcs();
 
+    reason = vmread(vcpu, VM_EXIT_INFO_REASON);
     if (result == VMX_FAIL_VALID) {
-        hax_log("HAX: Prev exit: %llx error code: %llx\n",
-                vmread(vcpu, VM_EXIT_INFO_REASON),
-                vmread(vcpu, VMX_INSTRUCTION_ERROR_CODE));
+        error = vmread(vcpu, VMX_INSTRUCTION_ERROR_CODE);
+        hax_error("VMfailValid. Prev exit: %llx. Error code: %lld (%s)\n",
```
raphaelning: nit: `%lld` => `%llu`

core/name.c
AlexAltea: Good observation, thanks! I'll fix it.

core/dump.c
AlexAltea: For the record: No truncation occurs ([check Ideone snippet](https://ideone.com/uKfC8r)), but I still agree that if those upper bytes are used, padding to 32-bits makes no sense and `%016llX` is much more convenient.

raphaelning: That's good to know, thanks!

core/name.c
AlexAltea: Fully agree with the first point.

As for adding the macro: I think it could make the code slightly more unreadable:
```diff
#define VMCS_ENC_INDEX_BITS 5
-(((x) & 0x003E) >> 1) /* Index */ | \
+(((x) & (((1 << VMCS_ENC_INDEX_BITS) - 1) << 1) >> 1
```
Promoting 0x003E (5-bit) to 0x007E (6-bit) or 0x00FE (7-bit) or 0x01FE (8-bit) feels more intuitive to me, since we can directly observe the `component_index_t` values, and simply "see" if our mask is compact enough. In fact, that's how I came up with 0x3E in the first place.

Just thought: If you still stand by the macro, I'll apply the inlined patch above.

raphaelning: > [...] since we can directly observe the component_index_t values, and simply "see" if our mask is compact enough.

Ah, that's an interesting perspective. I was going through all the tables in Appendix B and focusing on the Index column... Sure, please keep the hardcoded bitmasks.

AlexAltea:
All done!

PR #122, opened 2018-11-14
Add BT/BTS/BTR/BTC instructions

Marco01809:
Bit Test is used by TempleOS. With this change QEMU+HAXM are able to boot, install and run TempleOS.

Running the integrated test suite, 170/171 tests work without any problems.  One test crashes with a General Protection Fault, but this happens on KVM too, so I don't think it's haxms fault. (Tested on an i3-380M running Linux Kernel 4.18.10 x86_64)

Signed-off-by: Jan Erik Petersen <JanErikPetersen@web.de>

### DIFF
--- a/core/emulate.c
+++ b/core/emulate.c
@@ -165,6 +165,14 @@ static const struct em_opcode_t opcode_group3[8] = {
     F(em_neg, op_modrm_rm, op_none, op_none, 0),
 };
 
+static const struct em_opcode_t opcode_group8[8] = {
+    X4(N),
+    F(em_bt, op_none, op_none, op_none, 0),
+    F(em_bts, op_none, op_none, op_none, 0),
+    F(em_btr, op_none, op_none, op_none, 0),
+    F(em_btc, op_none, op_none, op_none, 0),
+};
+
 static const struct em_opcode_t opcode_group11[8] = {
     I(em_mov, op_none, op_none, op_none, INSN_MOV),
 };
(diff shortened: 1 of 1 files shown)


core/emulate.c
```diff
@@ -165,6 +165,14 @@ static const struct em_opcode_t opcode_group3[8] = {
     F(em_neg, op_modrm_rm, op_none, op_none, 0),
 };
 
+static const struct em_opcode_t opcode_group8[8] = {
+    X4(N),
+    F(em_bt, op_modrm_rm, op_simm8, op_none, 0),
```
raphaelning: nit: Since we have already specified the same operand types in the opcode table (line 260), there's no need to repeat them here. So this line would become:

`    F(em_bt, op_none, op_none, op_none, 0),`

and the same goes for the next 3 lines. In fact, `opcode_group1` and `opcode_group11` also follow this pattern.

Also nit: Just curious, how are the opcode groups numbered? I haven't been able to make sense of these numbers (1, 3, 8, 11, ...).

core/emulate.c
```diff
@@ -232,16 +240,25 @@ static const struct em_opcode_t opcode_table[256] = {
...
     X4(N),
+    /* 0xB0 - 0xBF */
+    X3(N),
+    F(em_btr, op_modrm_rm, op_modrm_reg, op_none, INSN_MODRM),
     X2(N),
     I(em_movzx, op_modrm_reg, op_modrm_rm8, op_none, INSN_MODRM | INSN_MOV),
     I(em_movzx, op_modrm_reg, op_modrm_rm16, op_none, INSN_MODRM | INSN_MOV),
-    X4(N),
+    X2(N),
+    G(opcode_group8, op_modrm_rm, op_simm8, op_none, 0),
```
raphaelning: I think the last parameter should be `INSN_MODRM` instead of 0. And that seems to be the only way to specify this flag: `em_decode_insn()` only checks this flag once, before it reads the flags of each member opcode. So setting it on each member of `opcode_group8` would have no effect.

raphaelning: Oops, my error, `G` implies `INSN_MODRM`, sorry!

core/emulate.c
AlexAltea: > Just curious, how are the opcode groups numbered? I haven't been able to make sense of these numbers (1, 3, 8, 11, ...).

This comes from the Intel SDM Vol. 2D, specifically:
- Groups 1, 3 and 11: *Table A-2. One-byte Opcode Map: (00H — F7H)*
- Group 8: *Table A-3. Two-byte Opcode Map: 88H — FFH (First Byte is 0FH)*

AlexAltea:
> In case you can also test this patch on Windows, could you write unit tests for the BTx instructions?

@raphaelning His message mentioned being on Linux 4.18.10, so if Windows is unavailable to him, I could take care of the tests: I wanted to add few extra cases this weekend anyway.

core/emulate.c
raphaelning: Wow, thanks! This is the first time I've opened Vol. 2D...

Marco01809:
Thanks for the fast review! I made the requested change and force-pushed it.

> On which host OS did you test this patch?

It was indeed the same Linux machine. Alex' linux port actually is what made me try out haxm.

> In case you can also test this patch on Windows, could you write unit tests for the BTx instructions?

Unfortunately, my experience with x86 is very limited and I'm not sure I could write meaningful tests, so I'd really appreciate either of you implementing them.

AlexAltea:
> I'd really appreciate it if either of you could implement them.
Sure! As soon as this is merged, I'll prepare the tests for it!

raphaelning:
Thank you both! We'll test this PR and get it merged.

PR #123, opened 2018-11-14
Added Windows to TravisCI build matrix

AlexAltea:
Although TravisCI doesn't support regular VS/EWDK installations yet, due to lack of early-access features and the impossibility of fetching huge dependencies at build-time, I found a way of integrating the WDK into Visual Studio 2017 Build Tools. Of course, this is rather a hack, but works surprisingly stable. The issue has been reported as well, and it's being considered for a future WDK:
- https://social.msdn.microsoft.com/Forums/windowshardware/en-US/efe5f9f8-c32d-4a25-87e2-abbe711dadbb/no-wdk-support-for-visual-studio-2017-build-tools?forum=wdk

Not sure what will happen first: (1) TravisCI pre-installing VS/EWDK or (2) WDK officially supporting VS2017 Build Tools, but in the meantime we can finally test HAXM on the 3 primary platforms. :-)

image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,6 @@ matrix:
         - cd platforms/linux
         - make -j$(nproc)
 
-  exclude: # TODO: Currently TravisCI does not support full VS/EWDK on Windows
     - name: "haxm-windows"
       os: windows
       install:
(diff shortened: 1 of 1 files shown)


AlexAltea:
Note that the Windows build is significantly slower than the *haxm-darwin* and *haxm-linux* ones (6-ish min). This is due to the installation of the WDK which takes around 5 minutes. Personally, I don't think it's too bad, since "human reviews" of a pull request take much longer on average.

Nonetheless, this has been reported to TravisCI (it's their CPU time!). I've suggested either adding support for caching Chocolatey packages, or otherwise preinstalling essential dependencies, e.g. WDK/SDK.

- https://travis-ci.community/t/caching-preserving-chocolatey-dependencies-across-builds/904

AlexAltea:
> since HAXM also supports 32-bit Linux and Windows hosts, it would be useful to add build configurations for those as well

Good point. It's definitely possible to include both 32-bit and 64-bit configurations without duplicating the scripts: The easiest option is just build it sequentially (e.g. first 32-bit, then 64-bit) in the same build job. Alternatively, and frequently the preferred solution, is building it in parallel with a separate build job for each configuration, but then configuration script will require heavy changes. I have no preference here.

Aside from 32-bit configurations, I have more things planned for TravisCI:
- [**Adding support for tests**](https://user-images.githubusercontent.com/5306886/48526693-e5005280-e888-11e8-8075-39cef9c59579.png). They require the Windows SDK, which installs fine locally as you see, but fails on TravisCI for some mysterious reason.
- **Build artifacts**: Automatically uploading the (unfortunately unsigned) kernel modules to TravisCI, in case we need someone to quickly test a specific build. This also serves as a source of nightly builds.

These could wind up as another PR, so support for 32-bit configurations could tag along then. :-)

raphaelning:
Sounds great, thanks! I have no idea about the Windows SDK install failure... I guess if we can't retrieve the choco log file from Travis, there's not much we can do about it :-(

PR #126, opened 2018-11-17
Rename typename cpuid_t to hax_cpuid_t

krytarowski:
This removes type clash with NetBSD kernel internals.

### DIFF
--- a/core/include/cpu.h
+++ b/core/include/cpu.h
@@ -44,7 +44,7 @@
 struct vcpu_t;
 struct vcpu_state_t;
 
-typedef uint32_t cpuid_t;  // CPU identifier
+typedef uint32_t hax_cpuid_t;  // CPU identifier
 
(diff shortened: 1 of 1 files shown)


krytarowski:
Done.

AlexAltea:
The *haxm-windows* has failed, but that is due to an update of the [windowsdriverkits10](https://chocolatey.org/packages/windowsdriverkit10) package, which needed updating to sync with the [windows-sdk-10.1](https://chocolatey.org/packages/windows-sdk-10.1), which in turn needed an update to fix an issue on Windows Server 2016 (v1803), which is used by TravisCI.

This branch is safe to be merged. I'll push a fix for the TravisCI script as soon as possible.

raphaelning:
@AlexAltea Thanks for root causing the build failure so quickly! I'll proceed with the merge.

PR #127, opened 2018-11-17
Namespace local FXSAVE functions with haxm_

krytarowski:
This removes type clash with NetBSD kernel internals.


krytarowski:
Done.

Issue #40, opened 2018-04-04
haxm much more slow than kvm?


mifritscher:
PC with haxm:  i7 6820HQ, 16 GB RAM, NVMe SSD - Windows 7, haxm 7.1, qemu 2.11
PC with kvm: i7-5500U, 16 GB RAM, SATA SSH - Ubuntu 18.04, kernel 4.15, qemu 2.11

A simple Ubuntu 16.04 installation lasts 16 minutes on the haxm-PC, but only 7 minutes on the kvm-PC...
Is there a technical reason for that or more a bug / inefficiency somewhere?

mifritscher:
Also I noticed that haxm is much slower than VMWare on the 6820HQ machine. At least in the use case "let 7 java8 programs run with light network activity themself and to outside with Ubuntu 16.04". On VMware, there is about 10% cpu utlisation (if not less), while on haxm I've 60-90% cpu utilization on the 2 virtual cores.

AlexAltea:
Recent patches have changed the instruction emulator (#42) and reduced the number of feature checks while load/saving states while entering/exiting the VM (#63). Although the emulator won't probably improve performance, the feature check patches most certainly will.

Out of sheer curiosity: If you still have the same hardware around, and building+installing the drivers is easy for you, could you rerun the tests? Just to see if we are below 16 minutes now.

 ---

@raphaelning We could really use some set of microkernels to do benchmarks of HAXM. I assume the sources of performance penalties are:
- [VMX exit handlers](https://github.com/intel/haxm/blob/e9f8c89/core/vcpu.c#L369-L395).
- [QEMU exit handlers](https://github.com/qemu/qemu/blob/23c11b0/target/i386/hax-all.c#L531-L586).
- Context switches: both VM-HAXM switches and HAXM-QEMU switches.

These three are always triggered due to a VM-exit event. So we could have a set of microkernels, each repeatedly triggering a VM-exit with a specific exit reasons, and comparing average execution time against other hypervisors.

mifritscher:
Sadly I've no suitable build environment :-( But I've still access to the same hardware, so if you give me the binaries I could test them.

wcwang:
Thanks for your reply. The driver binary based on the latest code has been provided as below. You may follow the steps to test the performance.

[...]

PR #130, opened 2018-11-18
Optimizing VMCS reads/writes with a new cache implementation

AlexAltea:
**Disclaimer: This patch isn't finished, but it's available for early reviews/discussion.**

## Motivation

During development, running HAXM in a virtual machine prevents your "bare-metal" host system from kernel panics if runtime errors occur. The downside of nested virtualization, is that during L2-guest VM-exit events, the L1-hypervisor, i.e. HAXM, might need to perform actions, e.g. vmread/vmwrite, that trigger another exit to the L0-hypervisor, i.e. KVM, VMware, etc. These recursive exits add latency so it's convenient to minimize them.

While booting few kernels using both HAXM and KVM as L1-hypervisor, under the same conditions, HAXM appears to be 4-5 times slower than KVM (related: #40). After measuring the amount of vmread/vmwrite's between guest-exit and guest-enter events, I've noticed HAXM amount is (unsurprisingly) 4-5 times higher. Specifically:

- **KVM**: Does 6-10 reads and 2-6 writes on average.
  ~**KVM**: Does 4-8 reads and 0-4 writes on average (depends on VMX exit handler).~
  
- **HAXM**: Does 32-33 reads and 21-23 writes on average.
  ~**HAXM**: Does 32 reads and 1 write (nearly always!).~

Here's the raw data, corresponds to around the first few seconds of trying to boot a Windows 7 install disc: [vmx-measurements.zip](https://github.com/intel/haxm/files/2600897/vmx-measurements.zip) ~([haxm-vmx.log](https://github.com/intel/haxm/files/2592959/haxm-vmx.log) and [kvm-vmx.log](https://github.com/intel/haxm/files/2592961/kvm-vmx.log))~. Efforts at 96af3d28 (thanks @junxiaoc!) have improved this situation, but there's still room for improvement.

## Changes

Previous efforts, and other hypervisors like KVM, have minimized vmread/vmwrite usage by caching the component values with hand-written helper functions. While this *works*, it doesn't scale well for the 120+ components. Plus, it's not trivial what should be cached: caching unused components is counterproductive. Fine-tuning the set of cached components when in manually-implemented code is too time-consuming.

Instead, this patch relies in preprocessor [X-Macros](https://en.wikipedia.org/wiki/X_Macro) to **automatically** create the cache structures, alongside cached/dirty flags and helper functions to access **all** VMCS components. This [Godbolt snippet](https://godbolt.org/z/_3tuHb) allows to inspect the preprocessor output for clear view of the resulting code (ideally, apply [clang-format](https://zed0.co.uk/clang-format-configurator/)).

Summarizing the changes, this patch:
- Moves some VMX-specific definitions to *vmx.h* (required by other changes).
- Adds VMCS-cache macros to *vmx.h*.
- Optimizes code by using new VMCS-cache macros: Removed most _vmcs_pending*_ fields and unnecessary *vmread*'s at the end of `cpu_vmx_execute`.

## Benchmarks

Measuring time elapsed between booting a virtual machine with a Windows XP SP2 installation disc, until the installer shows the "welcome screen", using an Ubuntu 18.04 (+ Linux 4.17) host. This is a particularly bad scenario that shows up to x10 slowdown. (Disclaimer: time measured manually with my phone).

- **KVM**: 45.4 seconds.
- **HAXM (before)**: 532.5 seconds.
- **HAXM (after)**: *TODO*.

## Pending

- [ ] Using VMCS-cache for VCPU state reads/writes (huge performance impact!).
- [ ] Replacing remaining vmread/vmwrite's.

### DIFF
--- a/core/include/vmx.h
+++ b/core/include/vmx.h
@@ -877,11 +877,6 @@ struct vcpu_vmx_data {
 
     uint64_t cr0_mask, cr0_shadow;
     uint64_t cr4_mask, cr4_shadow;
-    uint32_t entry_exception_vector;
-    uint32_t entry_exception_error_code;
-
-    interruption_info_t entry_intr_info;
-    uint32_t entry_instr_length;
 
     exit_reason_t exit_reason;
     exit_qualification_t exit_qualification;
@@ -914,12 +909,12 @@ void vmx_vmwrite(struct vcpu_t *vcpu, const char *name,
  */
 #define COMP_R_0(width, name)                                                 \
     static inline COMP_TYPE_##width                                           \
-                  vmcs_read_##name(struct vcpu_vmx_data* vcpu_vmx) {          \
+                  vmcs_read_##name(struct vcpu_vmx_data *vcpu_vmx) {          \
         return vmread(vcpu_vmx->parent, name);                                \
     }
 #define COMP_R_1(width, name)                                                 \
     static inline COMP_TYPE_##width                                           \
-                  vmcs_read_##name(struct vcpu_vmx_data* vcpu_vmx) {          \
+                  vmcs_read_##name(struct vcpu_vmx_data *vcpu_vmx) {          \
         if (vcpu_vmx->vmcs_cache_r.name##_cache)                              \
             return vcpu_vmx->vmcs.name##_value;                               \
         vcpu_vmx->vmcs.name##_value = vmread(vcpu_vmx->parent, name);         \
@@ -932,12 +927,12 @@ void vmx_vmwrite(struct vcpu_t *vcpu, const char *name,
   * Writes the VMCS-component <name> to the specified VCPU (cached if enabled).
   */
 #define COMP_W_0(width, name)                                                 \
-    static inline void vmcs_write_##name(struct vcpu_vmx_data* vcpu_vmx,      \
+    static inline void vmcs_write_##name(struct vcpu_vmx_data *vcpu_vmx,      \
                                          COMP_TYPE_##width value) {           \
         vmwrite(vcpu_vmx->parent, name, value);                               \
     }
 #define COMP_W_1(width, name)                                                 \
-    static inline void vmcs_write_##name(struct vcpu_vmx_data* vcpu_vmx,      \
+    static inline void vmcs_write_##name(struct vcpu_vmx_data *vcpu_vmx,      \
                                          COMP_TYPE_##width value) {           \
         vcpu_vmx->vmcs.name##_value = value;                                  \
         /*vcpu_vmx->vmcs_cache_r.name##_cache = 1;*/                          \
--- a/core/include/vmx.h
+++ b/core/include/vmx.h
@@ -338,7 +338,7 @@ typedef enum component_index_t component_index_t;
     COMP(0, 0, W_UL, HOST_IDTR_BASE)                         \
     COMP(0, 0, W_UL, HOST_SYSENTER_ESP)                      \
     COMP(0, 0, W_UL, HOST_SYSENTER_EIP)                      \
-    COMP(0, 0, W_UL, GUEST_RIP)                              \
+    COMP(1, 0, W_UL, GUEST_RIP)                              \
     COMP(0, 0, W_UL, GUEST_RFLAGS)                           \
     COMP(0, 0, W_UL, GUEST_RSP)                              \
     COMP(0, 0, W_UL, GUEST_CR0)                              \
--- a/core/include/vmx.h
+++ b/core/include/vmx.h
@@ -320,7 +320,7 @@ typedef enum component_index_t component_index_t;
     COMP(0, 0, W_UL, VMX_CR0_READ_SHADOW)                    \
     COMP(0, 0, W_UL, VMX_CR4_READ_SHADOW)                    \
     COMP(0, 0, W_UL, VMX_CR3_TARGET_VAL_BASE)                \
-    COMP(0, 0, W_UL, VM_EXIT_INFO_QUALIFICATION)             \
+    COMP(1, 0, W_UL, VM_EXIT_INFO_QUALIFICATION)             \
     COMP(0, 0, W_UL, VM_EXIT_INFO_IO_ECX)                    \
     COMP(0, 0, W_UL, VM_EXIT_INFO_IO_ESI)                    \
     COMP(0, 0, W_UL, VM_EXIT_INFO_IO_EDI)                    \
@@ -377,7 +377,7 @@ typedef enum component_index_t component_index_t;
     COMP(0, 0, W_32, VMX_CR3_TARGET_COUNT)                   \
     COMP(0, 0, W_32, VMX_PREEMPTION_TIMER)                   \
     COMP(0, 0, W_32, VMX_INSTRUCTION_ERROR_CODE)             \
-    COMP(0, 0, W_32, VM_EXIT_INFO_REASON)                    \
+    COMP(1, 0, W_32, VM_EXIT_INFO_REASON)                    \
     COMP(1, 0, W_32, VM_EXIT_INFO_INTERRUPT_INFO)            \
     COMP(1, 0, W_32, VM_EXIT_INFO_EXCEPTION_ERROR_CODE)      \
     COMP(1, 0, W_32, VM_EXIT_INFO_IDT_VECTORING)             \
@@ -405,7 +405,7 @@ typedef enum component_index_t component_index_t;
     COMP(0, 0, W_32, GUEST_IDTR_LIMIT)                       \
     COMP(0, 0, W_32, GUEST_SYSENTER_CS)                      \
     COMP(0, 0, W_32, GUEST_SMBASE)                           \
-    COMP(0, 0, W_32, GUEST_INTERRUPTIBILITY)                 \
+    COMP(1, 0, W_32, GUEST_INTERRUPTIBILITY)                 \
     COMP(0, 0, W_32, GUEST_ACTIVITY_STATE)                   \
     /* 16-bit components */                                  \
     COMP(0, 0, W_16, VMX_VPID)                               \
--- a/core/include/vmx.h
+++ b/core/include/vmx.h
@@ -102,6 +102,7 @@ enum {
     VMX_EXIT_XRSTORS                 = 64
 };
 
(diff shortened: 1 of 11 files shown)


raphaelning (commented):
Thanks! You've identified an important problem, and I like your systematic approach to solving it. I'm anxious to see how much performance improvement this PR will eventually make, and maybe not just for the nested virtualization use case :-)

So far I've only finished reviewing the first commit, and please allow me more time for the rest. Meanwhile, I have a question about the VMREAD/VMWRITE data you collected, which I'll ask in a separate post.

raphaelning:
First of all, let me share the "decoder" you sent me offline, in case other people are wondering how to interpret the logs:

>    $(guest rip) <= $(vmx exit reason)!
>    delta: $(absolute difference between rdtsc after guest exit and before guest enter)
>    vmread_ctr: $(number of vmread’s between guest exit and guest enter)
>    vmwrite_ctr: $(number of vmwrite’s between guest exit and guest enter)

I'm just surprised by how small `vmwrite_ctr` is: indeed, for HAXM it's almost always 1 or 2, although 9 also appears once in `haxm-vmx.log`.

But if you look at `vcpu_save_host_state()` (`core/vcpu.c`) alone, which is called before every VM entry, there are already 14 unconditional VMWRITEs! So is `vcpu_save_host_state()` not covered by `vmwrite_ctr`? When do you increment this counter, and when do you reset it?

For the ease of discussion, let me provide an outline of the current VM entry-exit workflow:

```
cpu_vmx_execute() {
    while (1) {
        ...
        load_vmcs()  // VMXON + VMPTRLD
        VMWRITE compA  // Saves host state, flushes dirty guest state (RIP, RFLAGS, etc.), updates VMX controls, etc.
        VMWRITE compB
        VMWRITE compC
        …  // There are some VMREADs as well
        asm_vmxrun()  // VM entry (VMLAUNCH)
        // VM exit
        VMREAD compX  // Caches VM exit info and some guest state (RIP, RFLAGS, etc.)
        VMREAD compY
        VMREAD compZ
        …  // Are there any VMWRITEs as well?
        put_vmcs()  // VMCLEAR + VMXOFF
        cpu_vmexit_handler()  {  // Handler specific to the VM exit reason
            // Ideally, no VMREAD or VMWRITE should be issued from here, since VMX is off (requiring an implicit load_vmcs()).
            // Instead, VMCS accesses are redirected to our cache.
            // PR #117 should have eliminated most of such expensive VMREAD/VMWRITE calls, if not all.
        }
    }
}
```

include/vcpu_state.h
```diff
@@ -187,7 +173,6 @@ struct vcpu_state_t {
 
     uint32_t _activity_state;
     uint32_t pad;
-    interruptibility_state_t _interruptibility_state;
```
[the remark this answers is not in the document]
AlexAltea: Good point, thank you!

AlexAltea:
> I'm just surprised by how small vmwrite_ctr is [...] So is vcpu_save_host_state() not covered by vmwrite_ctr? When do you increment this counter, and when do you reset it?

Nice catch! On HAXM, the counter is incremented in *vmread*/*vmx_vmwrite*, which is fine since on 64-bit each will generate one single vmread/vmwrite instruction. The KVM counters also increment correctly.

However, on HAXM, I had placed the counter display/reset before and after `cpu_vmx_run`, respectively, without realizing that more accesses ocurred inside this function. I've updated my code, and now we are wrapping `asm_vmxrun` instead. A similar issue happened with KVM, which is fixed too.

I repeated the benchmarks, and these are the updated results:
- **KVM**: Does 6-10 reads and 2-6 writes on average.
- **HAXM**: Does 32-33 reads and 21-23 writes on average.

The updated full logs (and the diff's to replicate them!) are available at: [vmx-measurements.zip](https://github.com/intel/haxm/files/2600897/vmx-measurements.zip)

core/intr_exc.c
```diff
@@ -218,20 +218,16 @@ void hax_inject_exception(struct vcpu_t *vcpu, uint8_t vector, uint32_t error_co
         if (error_code != NO_ERROR_CODE) {
             intr_info |= 1 << 11;
             if (vector == VECTOR_PF) {
-                vcpu->vmcs_pending_entry_error_code = 1;
-                vmx(vcpu, entry_exception_error_code) = error_code;
+                vmcs_write(vcpu, VMX_ENTRY_EXCEPTION_ERROR_CODE, error_code);
             } else {
                 vmwrite(vcpu, VMX_ENTRY_EXCEPTION_ERROR_CODE, error_code);
```
raphaelning: Use `vmcs_write` instead?

core/intr_exc.c
```diff
@@ -218,20 +218,16 @@ void hax_inject_exception(struct vcpu_t *vcpu, uint8_t vector, uint32_t error_co
...
     if (vector == VECTOR_PF) {
-        vcpu->vmcs_pending_entry_instr_length = 1;
-        vmx(vcpu, entry_instr_length) = exit_instr_length;
-        vcpu->vmcs_pending_entry_intr_info = 1;
-        vmx(vcpu, entry_intr_info).raw = intr_info;
-        vcpu->vmcs_pending = 1;
+        vmcs_write(vcpu, VMX_ENTRY_INSTRUCTION_LENGTH, exit_instr_length);
+        vmcs_write(vcpu, VMX_ENTRY_INTERRUPT_INFO, intr_info);
     } else {
         vmwrite(vcpu, VMX_ENTRY_INSTRUCTION_LENGTH, exit_instr_length);
```
raphaelning: Use `vmcs_write` instead? (Also applies to the next line.)

core/include/vmx.h
```diff
@@ -636,6 +855,39 @@ struct invept_desc {
...
+    uint32_t pin_ctls;
+    uint32_t pcpu_ctls;
+    uint32_t scpu_ctls;
+    uint32_t entry_ctls;
+    uint32_t exc_bitmap;
+    uint32_t exit_ctls;
+
+    uint64_t cr0_mask, cr0_shadow;
+    uint64_t cr4_mask, cr4_shadow;
+    uint32_t entry_exception_vector;
```
raphaelning: nit: This is only written but never read, so I'd remove it.

core/include/vmx.h
```diff
@@ -636,6 +855,39 @@ struct invept_desc {
...
+    uint32_t entry_exception_error_code;
```
raphaelning: The new cache should have made this obsolete. Remove?

core/include/vmx.h
```diff
@@ -636,6 +855,39 @@ struct invept_desc {
...
+
+    interruption_info_t entry_intr_info;
```
raphaelning: This old write-cache is now obsolete/redundant. Remove?

core/include/vmx.h
```diff
@@ -636,6 +855,39 @@ struct invept_desc {
...
+    uint32_t entry_instr_length;
```
raphaelning: Same here.

core/vmx.c
```diff
@@ -296,3 +297,25 @@ void get_interruption_info_t(interruption_info_t *info, uint8_t v, uint8_t t)
...
 }
+
+#define COMP_PENDING_0(name)
+#define COMP_PENDING_1(name) \
+    if (vcpu->vmx.vmcs_cache_w.name##_dirty) \
+        vmwrite(vcpu, name, vcpu->vmx.vmcs.name##_value);
+#define COMP_PENDING(cache_r, cache_w, width, name) \
+    COMP_PENDING_##cache_w(name)
+
+void vcpu_handle_vmcs_pending(struct vcpu_t* vcpu)
```
raphaelning: You have changed the semantics of this function, but not where it is called. Now that it has the responsibility of flushing all the dirty VMCS values, we should delay it as much as possible, in case `vmcs_write` gets called again after the flush but before VM entry (`asm_vmxrun`). Probably we need to draw a line somewhere in `cpu_vmx_execute` or in `cpu_vmx_run`:

```
// Cached vmcs_write() is preferred over uncached vmwrite()
vmcs_write(..);
vmcs_write(..);
...
// *** No cached vmcs_write allowed after this point ***
vmwrite(..);
vmwrite(..);
...
// VM entry
```

core/intr_exc.c
AlexAltea: In that case, I should get rid of the `if (vector == VECTOR_PF)` altogether... I wonder why it was placed there to begin with.

core/intr_exc.c
AlexAltea: In that case, I should get rid of the `if (vector == VECTOR_PF)` altogether... I wonder why it was placed there to begin with.

raphaelning:
> The updated full logs (and the diff's to replicate them!) are available at: vmx-measurements.zip.

Thanks, the new data is plausible. Clearly there's a lot of room for improvement for HAXM, to look on the bright side :-)

AlexAltea:
I've fixed some of the issues you mentioned. To make reviewing easier, I'll avoid rewriting the commit history until this is ready to merge. Then, I'll squash and re-sign everything into the first three commits, i.e.:

- Moved VMX-specific definitions to vmx.h
- Added VMCS-cache macros
- Optimize code by using new VMCS-cache macros

AlexAltea:
I'm trying to figure out when is the best time to flush certain VMCS components:

Additionally, I'm thinking about defining several levels of the cache:
- 0 - Uncached
- 1 - Cached, flushed on right after guest-exit.
- 2 - Cached, .

xianxiaoyin:
Build finished.

xianxiaoyin:

yiqisiben:

yiqisiben:

wayne-ma:
OK to verify

nevilad:
What is with this PR? My windows 7 guest boots and runs pretty slow compared to VmWare, I wan't do some enhancements to speed it up. This PR is a good start point.

wcwang:
Thanks for @nevilad comments. I noticed that @AlexAltea added new commits a month ago. And @AlexAltea once declared that the pull request were not completed. @AlexAltea, could you help to rebase the current pull request to the head of master branch first? So we can consider to proceed it and @nevilad can start the enhancement based on it. Thanks for both of you.

nevilad:
> I noticed that @AlexAltea added new commits a month ago.

I don't see new comments, the latest is in november of 2018.

PR #131, opened 2018-11-19
Rename assert to hax_assert to remove conflict with native symbol

krytarowski:
NetBSD defines assert() internally inside the kernel headers and
this causes fatal conflicts. Namespace the local symbol to hax_.


include/darwin/hax_mac.h
```diff
@@ -198,6 +198,8 @@ extern int default_hax_log_level;
             v->panicked = 1;          \
         }
 
+#define hax_assert assert
```
raphaelning: For Linux and Windows, `hax_assert` is defined as/like a function, and I think we should follow that style here:

```suggestion
#define hax_assert(condition) assert(condition)
```

krytarowski:
Done.

PR #132, opened 2018-11-19
Improved test_insn_xN_xN semantics

AlexAltea:
- Removed unnecessary `readonly_dst` parameter and enforced identical destination in each test case (only affects the `test` instruction so far).

- Templated over integer parameters *N* and *M* representing the width of both instruction operands: test_insn_xN_xM. Unless explicitly specified, *M* will be equal to *N* by default.

- Enforced sign-aware truncation of immediate types via the `imm_t` type. To verify this, the second operand in BT/BTC/BTR/BTS has changed from 0x7F to 0xFF, in their very last test case (this operand gets casted to -1).

### DIFF
--- a/tests/test_emulator.cpp
+++ b/tests/test_emulator.cpp
@@ -36,6 +36,17 @@
 
 #include "../core/include/emulate.h"
 
+/* Immediate types */
+template <int N>
+struct imm_t;
+template <> struct imm_t<8>  { int8_t  val; };
+template <> struct imm_t<16> { int16_t val; };
+template <> struct imm_t<32> { int32_t val; };
+template <> struct imm_t<64> { int32_t val; };
+
+template <int N>
+using imm = decltype(imm_t<N>::val);
+
 /* Emulator operations */
 struct test_cpu_t {
     uint64_t gpr[16];
@@ -278,15 +289,14 @@ protected:
         EXPECT_LT(ret, 0);
     }
 
-    template <int N>
-    void test_insn_rN_rN(const char* insn_name,
-                         const std::vector<test_alu_2op_t>& tests,
-                         bool readonly_dst = false) {
+    template <int N, int M=N>
+    void test_insn_rN_rM(const char* insn_name,
+                         const std::vector<test_alu_2op_t>& tests) {
         char insn[256];
         test_cpu_t vcpu_original;
         test_cpu_t vcpu_expected;
         snprintf(insn, sizeof(insn), "%s %s, %s", insn_name,
-            gpr<N>(REG_RDX), gpr<N>(REG_RCX));
+                 gpr<N>(REG_RDX), gpr<M>(REG_RCX));
 
         // Run tests
         for (const auto& test : tests) {
@@ -295,17 +305,15 @@ protected:
             vcpu_original.gpr[REG_RCX] = test.in_src;
             vcpu_original.flags = test.in_flags;
             vcpu_expected = vcpu_original;
-            if (!readonly_dst)
-                vcpu_expected.gpr[REG_RDX] = test.out_dst;
+            vcpu_expected.gpr[REG_RDX] = test.out_dst;
             vcpu_expected.flags = test.out_flags;
             run(insn, vcpu_original, vcpu_expected);
         }
     }
 
-    template <int N>
-    void test_insn_rN_iN(const char* insn_name,
-                         const std::vector<test_alu_2op_t>& tests,
-                         bool readonly_dst = false) {
+    template <int N, int M=N>
+    void test_insn_rN_iM(const char* insn_name,
+                         const std::vector<test_alu_2op_t>& tests) {
         char insn[256];
         test_cpu_t vcpu_original;
         test_cpu_t vcpu_expected;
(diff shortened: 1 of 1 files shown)


AlexAltea:
The main reason, in my opinion, to remove `readonly_dst` is that the *test_insn_xN_xN* functions were supposed to encode the **syntax** of the instructions (number and type of operands), and not their **behavior** (effect on source and destination operands).

tests/test_emulator.cpp
```diff
@@ -295,70 +305,64 @@ class EmulatorTest : public testing::Test {
...
+                         const std::vector<test_alu_2op_t>& tests) {
         char insn[256];
         test_cpu_t vcpu_original;
         test_cpu_t vcpu_expected;
 
         // Run tests
         for (const auto& test : tests) {
             snprintf(insn, sizeof(insn), "%s %s, %d", insn_name,
-                gpr<N>(REG_RAX), (uint32_t)test.in_src);
+                gpr<N>(REG_RAX), (imm<M>)test.in_src);
```
raphaelning: nit: We could take this chance to fix the indentation.

```suggestion
                     gpr<N>(REG_RAX), (imm<M>)test.in_src);
```

tests/test_emulator.cpp
```diff
@@ -295,70 +305,64 @@ class EmulatorTest : public testing::Test {
...
+                         const std::vector<test_alu_2op_t>& tests) {
         char insn[256];
         test_cpu_t vcpu_original;
         test_cpu_t vcpu_expected;
 
         // Run tests
         for (const auto& test : tests) {
             snprintf(insn, sizeof(insn), "%s %s ptr [edx + 2*ecx + 0x10], %d",
-                insn_name, mem<N>(), (uint32_t)test.in_src);
+                insn_name, mem<N>(), (imm<M>)test.in_src);
```
raphaelning: nit:

```suggestion
                     insn_name, mem<N>(), (imm<M>)test.in_src);
```

tests/test_emulator.cpp
```diff
@@ -295,70 +305,64 @@ class EmulatorTest : public testing::Test {
...
-                         bool readonly_dst = false) {
+    template <int N, int M=N>
+    void test_insn_rN_mM(const char* insn_name,
+                         const std::vector<test_alu_2op_t>& tests) {
         char insn[256];
         test_cpu_t vcpu_original;
         test_cpu_t vcpu_expected;
         snprintf(insn, sizeof(insn), "%s %s, %s ptr [edx + 2*ecx + 0x10]",
-            insn_name, gpr<N>(REG_RAX), mem<N>());
+            insn_name, gpr<N>(REG_RAX), mem<M>());
```
raphaelning: nit:
```suggestion
                 insn_name, gpr<N>(REG_RAX), mem<M>());
```

tests/test_emulator.cpp
```diff
@@ -369,22 +373,20 @@ class EmulatorTest : public testing::Test {
...
-                         bool readonly_dst = false) {
+    template <int N, int M=N>
+    void test_insn_mN_rM(const char* insn_name,
+                         const std::vector<test_alu_2op_t>& tests) {
         char insn[256];
         test_cpu_t vcpu_original;
         test_cpu_t vcpu_expected;
         snprintf(insn, sizeof(insn), "%s %s ptr [edx + 2*ecx + 0x10], %s",
-            insn_name, mem<N>(), gpr<N>(REG_RAX));
+            insn_name, mem<N>(), gpr<M>(REG_RAX));
```
raphaelning: nit:

```suggestion
                 insn_name, mem<N>(), gpr<M>(REG_RAX));
```

AlexAltea:
All done. I've also fixed the *snprintf* argument indentation in the functions below *test_insn_xN_xN_xN*, and in *test_insn_rN_rM*, i.e. all remaining cases. Changes are force-pushed on top of the existing commit.

Also, ignore the build issue. This is due to a Chocolatey package update, and is fixed in #133.
Probably that patch should be merged before this one.

---

> I get the part where 0xFF is cast to -1, but Intel SDM doesn't seem to allow this....

I've noticed it, but this seems in contradiction with *Intel SDM Vol. 2A: BT—Bit Test* (and BTC, BTR, BTS). Quoting below:

> If the bit base operand specifies a register, the **instruction takes the modulo 16, 32, or 64 of the bit offset operand** (modulo size depends on the mode and register size; 64-bit operands are available only in 64-bit mode). **This allows any bit position to be selected**.

The section you are quoting *Intel SDM Vol. 2A: Table 3-2: Range of Bit Positions Specified by Bit Offset Operands* mentions:

> If the BitBase is a register, **the BitOffset can be in the range 0 to [15, 31, 63]** depending on the mode and register size.

Probably, rather than a contradiction what the SDM is trying to say (for BT/BTC/BTR/BTS) is that the modulo of the bit offset is taken **before** applying the *Bit* "macro", i,e, **Bit(BitBase, BitOffset % Width)**, so the *Bit* "macro" always receives an operand in range 0 to [15, 31, 63].

raphaelning:
Thanks! I agree with your interpretation of the `BT rN, imm8` case, which you have also verified by experiment. I still wonder what should be the correct behavior of `BT mN, imm8` when `imm8` is negative. Does the CPU take the bit from an address lower than `mN` (BitBase), as Figure 3-2 suggests? Or does it honor the operand size (N) and translate `imm8` (BitOffset) to an offset between [0, N)? My bet is on the latter. Or maybe we already know the answer, since `test_bt<64>` with `0xFF` yields the expected result for `test_insn_mN_iM<64>`. Is the `test_cpu_t::mem` array initialized to all-zero?

raphaelning:
Ah, this time Travis rebuild started instantly, and finished with flying colors.

AlexAltea:
@raphaelning You are right, the scenario `bt* mN, rN` has an edge-case that we are emulating wrong, since it can go out of bounds whenever *rN &lt; 0* or *rN &ge; N*. Thanks for the finding! See:

image

The problem here is that our "fast emulator" works by converting memory-operands into register-operands, which unfortunately have a different behavior. This seems a weird design choice of x86, but we have to comply with it. The best approach will be, I think, adding `INSN_BITOP` to the [instruction flags] and modifying resulting memory address based on the offset operand.

If you agree, I'll implement it that way and prepare more tests for BT/BTC/BTR/BTS.

---

Also, note that this out-of-bounds behaviour does not occur on `bt* rN, rN/i8` (due to modulo operation), or `bt* mN, i8` (as you said, due to constraints of the immediate operand specified in *Table 3-2*, although internally it does modulo, see image below).

image2

PR #135, opened 2018-11-21
Fixed out-of-bounds behavior for btx mN,rN instructions

AlexAltea:
Fixes issue reported at https://github.com/intel/haxm/pull/132#issuecomment-440606304.

The bit-test instructions BT/BTC/BTR/BTS, through the variant `bt* mN,rN`, allow selecting bytes outside the value referenced by the memory operand. This is now checked for all instructions flagged with INSN_BITOP. Corresponding tests have been added as well (they have been cross-checked with tests run on my CPU).

See also: *Intel SDM Vol. 2A: Figure 3-2. Memory Bit Indexing*.

### DIFF
--- a/core/emulate.c
+++ b/core/emulate.c
@@ -47,6 +47,8 @@
 #define INSN_NOFLAGS ((uint64_t)1 <<  6)
 /* Instruction has two memory operands */
 #define INSN_TWOMEM  ((uint64_t)1 <<  7)
+/* Instruction takes bit test operands */
+#define INSN_BITOP   ((uint64_t)1 <<  8)
 /* String instruction */
 #define INSN_STRING  (INSN_REP|INSN_REPX)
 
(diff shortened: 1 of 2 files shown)


raphaelning:
@AlexAltea Thanks. I’m traveling and AFK. May not be able to give you much feedback on these PRs until next Monday...

AlexAltea:
Have a nice trip! Tomorrow I'll head to HITB until Novermber 30th, so I'll be mostly AFK until then as well.

core/emulate.c
```diff
@@ -403,6 +417,8 @@ static em_status_t operand_read(struct em_context_t *ctxt,
         if (op->flags & OP_READ_PENDING) {
             rc = ctxt->ops->read_memory_post(ctxt->vcpu, &op->value, op->size);
         } else {
+            if (ctxt->opcode.flags & INSN_BITOP)
```
raphaelning: nit: This works, but kind of "hacks" into the usual `em_emulate_insn()` workflow: parsing of the destination operand ends up taking care of the next/source operand. I think it hurts readability.

Wouldn't it be better to place this logic somewhere between `decode_operands()` and the first `operand_read()` call?

tests/test_emulator.cpp
```diff
@@ -629,6 +629,24 @@ TEST_F(EmulatorTest, insn_bt) {
...
+    vcpu_original = {};
+    vcpu_original.gpr[REG_RCX] = 0ULL;
+    (uint64_t&)vcpu_original.mem[0x00] = 0x00020000'00000000;
+    (uint64_t&)vcpu_original.mem[0x08] = 0x00000000'00000000;
+    (uint64_t&)vcpu_original.mem[0x10] = 0xFFFFFFFF'FFFFFFFE;
+    
+    vcpu_original.gpr[REG_RAX] = -15LL;
+    vcpu_expected = vcpu_original;
+    vcpu_expected.flags |= RFLAGS_CF;
+    run("bt [ecx + 0x08], eax", vcpu_original, vcpu_expected);
```
raphaelning: I was surprised that this test case passes, due to the flaw I've pointed out above. I inspected `em_ctxt.dst.mem.ea` after `em_emulate_insn()`, and saw the wrong value I had expected (`0x20000006`, instead of `0x07`). Note that this address is well outside the bounds of the `test_cpu_t::mem` array. Then I realized that `test_read_memory()` does `ea &= 0xFF` before validating `ea` and `size`... That has to be fixed.

core/emulate.c
AlexAltea: True, that's much better. I'll apply your suggestion when I'm back next weekend.

core/emulate.c
```diff
@@ -382,6 +384,18 @@ static em_status_t segmented_write(struct em_context_t *ctxt,
     return ctxt->ops->write_memory(ctxt->vcpu, la, data, size, flags);
 }
 
+static void parse_bit_operands(struct em_context_t *ctxt)
+{
+    if (ctxt->dst.type == OP_MEM && ctxt->src1.type == OP_REG) {
+        int64_t offset;
+
+        offset = READ_GPR(ctxt->src1.reg.index, ctxt->src1.size);
```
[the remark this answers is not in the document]
AlexAltea: Fixed at dcf0a41.

For now, I'm just post-processing the offset by casting the value to the corresponding signed type (based on the operand size). Same approach is done by other hypervisors. If we see more instructions that have similar behavior, we can make a helper function instead.

Note that these casts occur only when `ctxt->dst.type == OP_MEM && ctxt->src1.type == OP_REG`. But that's alright since any other situation will mask the bit offset, so signedness can be ignored.

AlexAltea:
All done.

Once this passes your review, I'll squash it into a single commit (the first one) if you want.

core/emulate.c
```diff
@@ -1069,20 +1106,27 @@ em_status_t EMCALL em_emulate_insn(struct em_context_t *ctxt)
...
-    }
-    if (!(opcode->flags & INSN_MOV)) {
-        rc = operand_read(ctxt, &ctxt->dst);
+    if (opcode->flags & INSN_BITOP &&
+        ctxt->dst.type == OP_MEM && ctxt->src1.type == OP_REG) {
+        rc = operand_read_bitop(ctxt);
+        if (rc != EM_CONTINUE)
+            goto exit;
+    } else {
+        if (!(opcode->flags & INSN_NOFLAGS)) {
```
raphaelning: Why move this condition here? Any good reason to skip `read_rflags` for the `BT* mN, rN` case?

AlexAltea: Sorry, this was a typo, I've moved it out of the `else` case in 9a3e342 (and placed it before the `if`).

AlexAltea:
>> Once this passes your review, I'll squash it into a single commit (the first one) if you want.
>
> Thanks, please do. I think the removal of `ea &= 0xFF` is worth mentioning in the commit message.

Done!

PR #142, opened 2018-12-09
Added 32-bit Linux builds to Travis script

AlexAltea:
Added cross-compilation for 32-bit Linux hosts on the Travis configuration file, based on the scripts provided by @maronz (thank you!) on https://github.com/intel/haxm/pull/133#issuecomment-441545820. As you see, there's some extra warnings triggered when compiling the kernel module for 32-bit Linux hosts, but I'll leave that for future PRs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,15 +17,41 @@ matrix:
       os: linux
       dist: trusty
       sudo: false
+      env:
+        - CHROOT=trusty_i386
+        - CHROOT_DIR=/tmp/chroot/${CHROOT}
       before_install:
         - sudo apt-get update
         - sudo apt-get install -y linux-headers-`uname -r`
-        - wget http://mirrors.kernel.org/ubuntu/pool/universe/n/nasm/nasm_2.13.02-0.1_amd64.deb
+        - wget -P /tmp/downloads http://mirrors.kernel.org/ubuntu/pool/universe/n/nasm/nasm_2.13.02-0.1_amd64.deb
         - sudo apt-get install -y dpkg
-        - sudo dpkg -i nasm_2.13.02-0.1_amd64.deb
+        - sudo dpkg -i /tmp/downloads/nasm_2.13.02-0.1_amd64.deb
+        # Cross-compiling to 32-bit Linux
+        - mkdir -p ${CHROOT_DIR}/haxm
+        - cp -r . ${CHROOT_DIR}/haxm
+        - sudo apt-get install -y debootstrap schroot
+        - |
+          ( echo "[${CHROOT}]"
+            echo "type=directory"
+            echo "directory=${CHROOT_DIR}"
+            echo "personality=linux32"
+            echo "users=`id -un`"
+            echo "root-users=`id -un`"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -17,15 +17,41 @@ matrix:
...
+        - CHROOT_DIR=/tmp/chroot/${CHROOT}
       before_install:
         - sudo apt-get update
         - sudo apt-get install -y linux-headers-`uname -r`
         - wget http://mirrors.kernel.org/ubuntu/pool/universe/n/nasm/nasm_2.13.02-0.1_amd64.deb
         - sudo apt-get install -y dpkg
         - sudo dpkg -i nasm_2.13.02-0.1_amd64.deb
+        # Cross-compiling to 32-bit Linux
+        - mkdir -p ${CHROOT_DIR}/haxm
+        - cp -r . ${CHROOT_DIR}/haxm
```
raphaelning: nit: If the current directory is our git repo, maybe it would be better to tell `wget` to use a different location for the downloaded files? E.g. `wget -P /tmp/haxm-downloads/ <url>` (BTW, `-P` seems to automatically create the directory). Otherwise this line would copy an extra `.deb` file, and we would be building from a "dirty" source tree.

krytarowski:
> Is it because struct hax_msr_data is too big?

Actually, it's because the variable of this type is stored more than once on the stack. There are various limits of frame size depending on the context.

krytarowski:
A quick hack is to `struct hax_msr_data msrs;` to define once in `hax_vcpu_ioctl()`. A cleaner solution is to allocate it dynamically with a kernel allocator.

raphaelning:
@krytarowski Thanks! Both solutions sound good to me.

raphaelning:
With this change, the `haxm-linux` Travis job finishes in under 3 minutes, which is still much faster than `haxm-windows` (8m29s). And because these jobs run in parallel, the total build check time should be unchanged.

.travis.yml
AlexAltea: Good point! Fixed as of 2403ad3.

.travis.yml
```diff
@@ -17,15 +17,41 @@ matrix:
...
+            echo "personality=linux32"
+            echo "users=`id -un`"
+            echo "root-users=`id -un`"
+          ) | sudo tee /etc/schroot/chroot.d/${CHROOT}.conf > /dev/null
+        - sudo debootstrap --variant=buildd --arch=i386 trusty ${CHROOT_DIR}
+        - schroot -c ${CHROOT} -u root -- sed -i "1p ; 1s| trusty | trusty-updates |" /etc/apt/sources.list
+        - schroot -c ${CHROOT} -u root -- apt update
+        - schroot -c ${CHROOT} -u root -- apt-cache search linux-header
+        - schroot -c ${CHROOT} -u root -- apt install -y linux-headers-`uname -r`
+        - wget http://mirrors.kernel.org/ubuntu/pool/universe/n/nasm/nasm_2.13.02-0.1_i386.deb
```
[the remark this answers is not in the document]
AlexAltea: Sorry, I overlooked this line. You are right. This is fixed as of 385a087.

PR #158, opened 2019-01-28
Correct the NetBSD hax_test_and_{set,clear}_bit operations

krytarowski:
Add a new version that mimics the Linux semantics more closely.
    
In particular this new version:
 - allows bit argument to be >63
 - adds memory barrier that is implied on Linux

With this change HAXM on NetBSD can boot NetBSD as guest.

### DIFF
--- a/platforms/netbsd/hax_wrapper.c
+++ b/platforms/netbsd/hax_wrapper.c
@@ -251,36 +251,34 @@ hax_atomic_t hax_atomic_dec(volatile hax_atomic_t *atom)
 
 int hax_test_and_set_bit(int bit, uint64_t *memory)
 {
-    volatile uint64_t *val;
-    uint64_t mask, old;
+    const unsigned int units = (sizeof(*memory) * CHAR_BIT);
+    volatile uint64_t *const p = &memory[bit / units];
+    const uint64_t mask = (1ULL << (bit % units));
+    uint64_t v;
 
-    val = (volatile uint64_t *)memory;
-    mask = 1 << bit;
-
+    membar_exit();
     do {
-        old = *val;
-        if ((old & mask) != 0)
-            break;
-    } while (atomic_cas_64(val, old, old | mask) != old);
+        v = *p;
+    } while (atomic_cas_64(p, v, (v | mask)) != v);
+    membar_enter();
 
-    return !!(old & mask);
+    return ((v & mask) != 0);
 }
 
 int hax_test_and_clear_bit(int bit, uint64_t *memory)
 {
-    volatile uint64_t *val;
-    uint64_t mask, old;
+    const unsigned int units = (sizeof(*memory) * CHAR_BIT);
+    volatile uint64_t *const p = &memory[bit / units];
+    const uint64_t mask = (1ULL << (bit % units));
+    uint64_t v;
 
-    val = (volatile uint64_t *)memory;
-    mask = 1 << bit;
-
+    membar_exit();
     do {
-        old = *val;
-        if ((old & mask) != 0)
-            break;
-    } while (atomic_cas_64(val, old, old & ~mask) != old);
+        v = *p;
+    } while (atomic_cas_64(p, v, (v & ~mask)) != v);
+    membar_enter();
 
-    return !!(old & mask);
+    return ((v & mask) == 0);
 }
 
 /* Spinlock */


krytarowski:
On the request by @AlexAltea I've checked if bit can be larger than 63.

The answer is yes, and there are used values in range 0-255

```
[...]
[   322,049030] hax_test_and_set_bit(255, 0xffffda825c559000)
[   322,049030] hax_test_and_set_bit(255, 0xffffda825c559000)
[   322,049030] hax_test_and_set_bit(255, 0xffffda825c559000)
[   322,049030] hax_test_and_set_bit(255, 0xffffda825c559000)
[   322,049030] hax_test_and_set_bit(255, 0xffffda825c559000)
[   338,806489] hax_test_and_set_bit(254, 0xffffda825c559000)
[   338,816493] hax_test_and_set_bit(253, 0xffffda825c559000)
[   338,846506] haxm_warning: Ignored guest CR8 write, val=0x0
[   340,397196] haxm_warning: Ignored guest CR8 write, val=0x0
[   340,477232] hax_test_and_set_bit(65, 0xffffda825c559000)
[   340,477232] hax_test_and_set_bit(66, 0xffffda825c559000)
[   340,477232] hax_test_and_set_bit(67, 0xffffda825c559000)
[   340,477232] hax_test_and_set_bit(68, 0xffffda825c559000)
[...]
```

And later iterating to 255 and using various values in the range at least of 255 bit position.

The original API could be better documented to note this behavior.

AlexAltea:
@krytarowski Good point. I've reviewed the existing implementations for Linux/Darwin/Windows, and they all support bit indices larger than 63, so this was indeed a bug. Thanks for submitting a patch!

platforms/netbsd/hax_wrapper.c
```diff
@@ -251,36 +251,34 @@ hax_atomic_t hax_atomic_dec(volatile hax_atomic_t *atom)
 
 int hax_test_and_set_bit(int bit, uint64_t *memory)
 {
-    volatile uint64_t *val;
-    uint64_t mask, old;
-
-    val = (volatile uint64_t *)memory;
-    mask = 1 << bit;
+    const unsigned int units = (sizeof(*memory) * CHAR_BIT);
+    volatile unsigned long *const p = &memory[bit / units];
```
raphaelning: nit: Just curious, what's the benefit of using `unsigned long` and `atomic_cas_ulong` instead of `uint64_t` and `atomic_cas_64`? I don't think the following code would work correctly if `sizeof(unsigned long) != sizeof(uint64_t)`.

platforms/netbsd/hax_wrapper.c
```diff
@@ -251,36 +251,34 @@ hax_atomic_t hax_atomic_dec(volatile hax_atomic_t *atom)
...
+    const unsigned long mask = (1UL << (bit % units));
+    unsigned long v;
 
+    membar_exit();
```
raphaelning: nit: Could you enlighten me on the use of memory barriers here? Is the atomic operation (CAS) not enough to guarantee the correct behavior?

platforms/netbsd/hax_wrapper.c
krytarowski: `ulong` is more portable and since we need to support >63 bit it's better to stick to it. In future it can work better on other CPUs that don't implement 64-bit atomics or just don't want to use compiler flags to get them to use like on 32-bit CPUs where we need i586/i686 target)

platforms/netbsd/hax_wrapper.c
krytarowski: I've followed Linux that implies memory barrier.

```
/**
 * test_and_set_bit - Set a bit and return its old value
 * @nr: Bit to set
 * @addr: Address to count from
 *
 * This operation is atomic and cannot be reordered.
 * It also implies a memory barrier.
 */
static __always_inline bool test_and_set_bit(long nr, volatile unsigned long *addr)
{
	return GEN_BINARY_RMWcc(LOCK_PREFIX __ASM_SIZE(bts), *addr, c, "Ir", nr);
}
```

krytarowski:
> I think we should declare these functions in a common header file, so we only need to document them once.

Please do, also please document the need (or lack of thereof) for memory barrier.

platforms/netbsd/hax_wrapper.c
raphaelning: That's a good point. But this line does an implicit cast from `uint64_t *` to `ulong *`. Shouldn't we make `units` match `sizeof(ulong)`? Something like:

```c
    unsigned long *p0 = (unsigned long *)memory;
    const unsigned int units = sizeof(*p0) * CHAR_BIT;
    volatile unsigned long *const p = &p0[bit / units];
```

platforms/netbsd/hax_wrapper.c
raphaelning: I see. I guess the underlying x86 implementation for `atomic_cas_ulong()` already implies a memory barrier, but since it's not specified in the documentation, your approach is more portable.

platforms/netbsd/hax_wrapper.c
krytarowski: I've decided to keep for now atomic_cas_64 and don't do trick on bits unless we can test it on a big endian setup.

krytarowski:
Done!

PR #159, opened 2019-01-28
Include hax_types in core/include/config.h

krytarowski:
This allows to the get correct reference of PLATFORM_NETBSD regardless
of include order from other part of the codebase.

### DIFF
--- a/core/include/config.h
+++ b/core/include/config.h
@@ -31,6 +31,8 @@
 #ifndef HAX_CORE_CONFIG_H_
 #define HAX_CORE_CONFIG_H_
 
+#include "../../include/hax_types.h"
+
 struct config_t {
     int memory_pass_through;
     int disable_ept;


raphaelning:
> This fixes the definition of HAX_MAX_VCPUS.

Sorry, what exactly does this fix? You have already included definition of `HAX_MAX_VCPU` (in `config.h`)  before using the macro. Why does the include order matter?

krytarowski:
This was my mistake. It's about `HAX_MAX_VMS`. Without including the file defining PLATFORM_NETBSD we fallback to `#else` clause and get 64 while we intend to support for now just 8.

raphaelning:
Actually, if we think about it, the root cause is that `config.h` uses the `HAX_PLATFORM_NETBSD` macro, but doesn't include the header where it is defined. So how about adding `#include ../../include/hax_types.h` to `config.h`? That's easier than fixing the order of includes for all the files that include `config.h`...

krytarowski:
> Actually, if we think about it, the root cause is that `config.h` uses the `HAX_PLATFORM_NETBSD` macro, but doesn't include the header where it is defined. So how about adding `#include "../../include/hax_types.h"` to `config.h`? That's easier than fixing the order of includes for all the files that include `config.h`...

I will do it.

PR #168, opened 2019-02-12
Fix FPU DNA exception on NetBSD

### LINKED ISSUE #166: HAXM triggers FPU DNA exception
Summary: 
HAXM from HEAD (as of 2019-02-11) triggers FPU DNA (Device Not Available) exception.

FPU DNA is triggered with calling FPU instructions (fxsave) inside kernel without FPU context available.

**Host Environment**
NetBSD/amd64 8.0 and HEAD (probably any supported version)

* HAXM version: 
HAXM from HEAD (as of 2019-02-11)

* Host OS version: 
NetBSD chieftec 8.99.34 NetBSD 8.99.34 (GENERIC) #12: Wed Feb  6 17:38:03 CET 2019  root@chieftec:/public/netbsd-root/sys/arch/amd64/compile/GENERIC amd64

x86_64 (amd64)

* Host CPU model: 
N/A

* Host RAM size: 
N/A

N/A

**Guest Environment**
Any non-trivial guest such as NetBSD.

* Android Emulator or QEMU version: 
qemu used directly, no Android Emulator.

* Guest OS version: 
Any.

* Guest OS architecture: 
x86_64, i386 untested

**To Reproduce**
`qemu-system-x86_64 -accel hax`

1. build qemu with NetBSD host support patch
2. build haxm from original github sources (unmodified)
3. insert haxm.kmod into the kernel
4. qemu-system-x86_64 -accel hax `

**Expected Behavior**

Stop triggering the FPU DNA exception.

FreeBSD code says that FPU DNA can be triggered by "some" VM engines.

**Reproducibility**

Always

**Diagnostic Information**

Host crash dump: 

Not provided, I don't like to panic my host again to get the backtrace.

HAXM log: 

Not provided.

Android Emulator or QEMU log: 

N/A

Screenshots:
 
Not provided.

**Additional context**

Please see the comment in `%cr0.TS`:

```
/*
 * Device Not Available (DNA, #NM) exception handler.
 *
 * It would be better to switch FP context here (if curthread !=
 * fpcurthread) and not necessarily for every context switch, but it
 * is too hard to access foreign pcb's.
 */
void
fpudna(void)
{
	struct thread *td;

	td = curthread;
	/*
	 * This handler is entered with interrupts enabled, so context
	 * switches may occur before critical_enter() is executed.  If
	 * a context switch occurs, then when we regain control, our
	 * state will have been completely restored.  The CPU may
	 * change underneath us, but the only part of our context that
	 * lives in the CPU is CR0.TS and that will be "restored" by
	 * setting it on the new CPU.
	 */
	critical_enter();

	KASSERT((curpcb->pcb_flags & PCB_FPUNOSAVE) == 0,
	    ("fpudna while in fpu_kern_enter(FPU_KERN_NOCTX)"));
	if (__predict_false(PCPU_GET(fpcurthread) == td)) {
		/*
		 * Some virtual machines seems to set %cr0.TS at
		 * arbitrary moments.  Silently clear the TS bit
		 * regardless of the eager/lazy FPU context switch
		 * mode.
		 */
		stop_emulating();
	} else {
		if (__predict_false(PCPU_GET(fpcurthread) != NULL)) {
			panic(
		    "fpudna: fpcurthread = %p (%d), curthread = %p (%d)\n",
			    PCPU_GET(fpcurthread),
			    PCPU_GET(fpcurthread)->td_tid, td, td->td_tid);
		}
		restore_fpu_curthread(td);
	}
	critical_exit();
}
```

https://github.com/freebsd/freebsd/blob/master/sys/amd64/amd64/fpu.c#L699

`stop_emulating()` translates to `clts()`

Cherry-picking this to the NetBSD kernel workarounds the issue.

http://www.netbsd.org/~kamil/patch-00083-fpu-dna-netbsd8.txt

However this workaround has been rejected by the x86 NetBSD port maintainer and we are expected to address this in HAXM. Additionally fixing this in HAXM will allow to use older released versions as they are.

NVMM/NetBSD works without this workaround.

This workaround is the only patch needed for the NetBSD kernel in order to run HAXM. Likely all other mainstream OSes workaround this (Windows, Darwin, Linux, FreeBSD, OpenBSD), but it's still likely a generic bug.

krytarowski:
Call clts (Clear Task-Switch) bit before calling FPU functions.
This prevents FPU Device Not Available exceptions.

Store the original CR0_TS of host before entering FPU state and
restore it on FPU exit.

Closes #166

### DIFF
--- a/core/vcpu.c
+++ b/core/vcpu.c
@@ -2041,6 +2041,13 @@ static void vcpu_enter_fpu_state(struct vcpu_t *vcpu)
     struct fx_layout *hfx = (struct fx_layout *)hax_page_va(hstate->hfxpage);
     struct fx_layout *gfx = (struct fx_layout *)hax_page_va(gstate->gfxpage);
 
+    hstate->cr0_ts = !!(get_cr0() & CR0_TS);
+
+    // Before executing any FPU instruction (e.g. FXSAVE) in host kernel
+    // context, make sure host CR0.TS = 0, so as to prevent a Device Not Found
+    // (#NM) exception, which can result in a host kernel panic on NetBSD.
+    hax_clts();
+
     hax_fxsave((mword *)hfx);
     hax_fxrstor((mword *)gfx);
 }
@@ -2052,8 +2059,14 @@ static void vcpu_exit_fpu_state(struct vcpu_t *vcpu)
     struct fx_layout *hfx = (struct fx_layout *)hax_page_va(hstate->hfxpage);
     struct fx_layout *gfx = (struct fx_layout *)hax_page_va(gstate->gfxpage);
 
+    hax_clts();
+
     hax_fxsave((mword *)gfx);
     hax_fxrstor((mword *)hfx);
+
+    if (hstate->cr0_ts) {
+        set_cr0(get_cr0() | CR0_TS);
+    }
 }
 
 // Instructions are never longer than 15 bytes:
--- a/core/ia32.c
+++ b/core/ia32.c
@@ -81,6 +81,11 @@ uint64_t ia32_rdtsc(void)
 #endif
 }
 
+void hax_clts(void)
+{
+    asm_clts();
+}
+
 void hax_fxinit(void)
 {
     asm_fxinit();
--- a/core/ia32_ops.asm
+++ b/core/ia32_ops.asm
@@ -213,6 +213,10 @@ function asm_enable_irq, 0
     sti
     ret
 
+function asm_clts, 0
+    clts
+    ret
+
 function asm_fxinit, 0
     finit
     ret
--- a/core/include/cpu.h
+++ b/core/include/cpu.h
@@ -84,6 +84,8 @@ struct hstate {
     uint64_t dr3;
     uint64_t dr6;
     uint64_t dr7;
+    // CR0
+    bool cr0_ts;
 };
 
 struct hstate_compare {
--- a/core/include/ia32.h
+++ b/core/include/ia32.h
@@ -72,6 +72,7 @@ void ASMCALL set_kernel_fs(uint16_t val);
 
 void ASMCALL asm_btr(uint8_t *addr, uint bit);
 void ASMCALL asm_bts(uint8_t *addr, uint bit);
+void ASMCALL asm_clts(void);
 void ASMCALL asm_fxinit(void);
 void ASMCALL asm_fxsave(mword *addr);
 void ASMCALL asm_fxrstor(mword *addr);
@@ -84,6 +85,8 @@ uint64_t ia32_rdmsr(uint32_t reg);
 void ia32_wrmsr(uint32_t reg, uint64_t val);
 
 uint64_t ia32_rdtsc(void);
+
+void hax_clts(void);
 
 void hax_fxinit(void);
 void hax_fxsave(mword *addr);


raphaelning:
Thanks for fixing the issue! IIUC, the NetBSD guest runs into \#NM because guest CR0.TS is set, and in this PR you are making sure host CR0.TS is clear right before every VM entry.

So my question is: what's the relationship between host CR0.TS and guest CR0.TS? Why does clearing the former result in the latter being cleared?

Note that there is a `GUEST_CR0` field in VMCS. According to Intel SDM 26.3.2.1 (_Loading Guest Control Registers, Debug Registers, and MSRs_):

> CR0 is loaded from the CR0 field with the exception of the following bits, which are never modified on VM entry: ET (bit 4); reserved bits 15:6, 17, and 28:19; NW (bit 29) and CD (bit 30). The values of these bits in the CR0 field are ignored.

So have you looked into how HAXM updates the TS bit (bit 3) of `GUEST_CR0`? In addition, there are also the `VMX_CR0_MASK` and `VMX_CR0_READ_SHADOW` fields in VMCS (cf. `vmwrite_cr()` in `core/vcpu.c`), whose exact values seem to depend on the host CPU model. Could you try to print them?

krytarowski:
According to my understanding the NetBSD kernel runs with CR0_TS set and we must explicitly disable this bit before any FPU operation in order to prevent FPU DNA. This is done in other parts of the NetBSD kernel whenever we do anything with FPU.

A similar magic is done in NVMM:

```
   1093 static void
   1094 svm_vcpu_guest_fpu_enter(struct nvmm_cpu *vcpu)
   1095 {
   1096 	struct svm_cpudata *cpudata = vcpu->cpudata;
   1097 
   1098 	cpudata->ts_set = (rcr0() & CR0_TS) != 0;
   1099 
   1100 	fpu_area_save(&cpudata->hfpu, svm_xcr0_mask); // implicit clts()
   1101 	fpu_area_restore(&cpudata->gfpu, svm_xcr0_mask); // implicit clts()
   1102 
   1103 	if (svm_xcr0_mask != 0) {
   1104 		cpudata->hxcr0 = rdxcr(0);
   1105 		wrxcr(0, cpudata->gxcr0);
   1106 	}
   1107 }
   1108 
   1109 static void
   1110 svm_vcpu_guest_fpu_leave(struct nvmm_cpu *vcpu)
   1111 {
   1112 	struct svm_cpudata *cpudata = vcpu->cpudata;
   1113 
   1114 	if (svm_xcr0_mask != 0) {
   1115 		cpudata->gxcr0 = rdxcr(0);
   1116 		wrxcr(0, cpudata->hxcr0);
   1117 	}
   1118 
   1119 	fpu_area_save(&cpudata->gfpu, svm_xcr0_mask); // implicit clts()
   1120 	fpu_area_restore(&cpudata->hfpu, svm_xcr0_mask); // implicit clts()
   1121 
   1122 	if (cpudata->ts_set) {
   1123 		stts();
   1124 	}
   1125 }
```

https://nxr.netbsd.org/xref/src/sys/dev/nvmm/x86/nvmm_x86_svm.c#1093

raphaelning:
Oh I see! Sorry, my earlier interpretation was completely off. It's not about the guest at all--this is just to make sure HAXM's own FXSAVE/FXRSTOR instructions don't cause a host \#NM. Now I can review the details of the code :-)

core/include/vcpu.h
```diff
@@ -238,6 +238,7 @@ struct vcpu_t {
 #ifdef CONFIG_HAX_EPT2
     struct mmio_fetch_cache mmio_fetch;
 #endif  // CONFIG_HAX_EPT2
+    int cr0_ts;
```
raphaelning: `struct vcpu_t` stores the guest CPU state, so putting the host CR0.TS state here can cause confusion. How about moving it to `struct hstate` (`core/include/cpu.h`)? Also, since this is a Boolean value, I'd use `bool` instead of `int`.

core/vcpu.c
```diff
@@ -2041,6 +2041,10 @@ static void vcpu_enter_fpu_state(struct vcpu_t *vcpu)
     struct fx_layout *hfx = (struct fx_layout *)hax_page_va(hstate->hfxpage);
     struct fx_layout *gfx = (struct fx_layout *)hax_page_va(gstate->gfxpage);
 
+    vcpu->cr0_ts = !!(get_cr0() & CR0_TS);
+
```
raphaelning: nit: Add a comment to explain why CLTS is necessary.

```suggestion

    // Before executing any FPU instruction (e.g. FXSAVE) in host kernel
    // context, make sure host CR0.TS = 0, so as to prevent a Device Not Found
    // (#NM) exception, which can result in a host kernel panic on NetBSD.
```

raphaelning:
BTW, just to answer my own question:

> So my question is: what's the relationship between host CR0.TS and guest CR0.TS?

The two are completely separate. At least on my Haswell host, VMX allows the guest full control over CR0.TS, so this bit is mapped to bit 3 of `GUEST_CR0` and has nothing to do with host CR0.TS.

core/include/vcpu.h
krytarowski: Done

krytarowski:
Done!

raphaelning:
Thanks! We'll test this on Windows and macOS.

Issue #175, opened 2019-02-14
QNX 6 guest - VCPU shutdown on boot


ensra:
**Describe the Bug**

Summary: QNX 6 fails to boot on NetBSD.

**Host Environment**

* HAXM version: c072ad9b68e1d558a9fb791511468d10a1a9b319

* Host OS version: NetBSD 8.99.34

* Host OS architecture: x86\_64

* Host CPU model: Intel(R) Core(TM) i5-7200U

* Host RAM size: 8 GB 

* (Optional) Host computer model: 

**Guest Environment**

* Android Emulator or QEMU version: QEMU 3.1.0

* Guest OS version: 6.5.0

* Guest OS architecture: x86

**To Reproduce**

1. `qemu-system-x86_64 -accel hax -m 1.5G -cdrom qnx-6.5.0-x86-201007091524.iso`

**Expected Behavior**

Expectation: QNX should boot.

**Reproducibility**

100%

**Diagnostic Information**

Host crash dump: n/a

HAXM log: 

~~~
[ 35108.592366] haxm_panic: em_decode_insn() failed: vcpu_id=0, len=2, CS:IP=0x0:0x40c761, instr[0..5]=0xf3 0xa6 0x75 0x2d 0xc7 0x44
[ 35108.592366] haxm_warning: 4000 VMX_PIN_CONTROLS: 1f
[ 35108.592366] haxm_warning: 4002 VMX_PRIMARY_PROCESSOR_CONTROLS: 969861fa
[ 35108.592366] haxm_warning: 401e VMX_SECONDARY_PROCESSOR_CONTROLS: aa
[ 35108.592366] haxm_warning: 4004 VMX_EXCEPTION_BITMAP: 40000
[ 35108.592366] haxm_warning: 4006 VMX_PAGE_FAULT_ERROR_CODE_MASK: 0
[ 35108.592366] haxm_warning: 4008 VMX_PAGE_FAULT_ERROR_CODE_MATCH: 0
[ 35108.592366] haxm_warning: 400c VMX_EXIT_CONTROLS: 236fff
[ 35108.592366] haxm_warning: 400e VMX_EXIT_MSR_STORE_COUNT: 0
[ 35108.592366] haxm_warning: 4010 VMX_EXIT_MSR_LOAD_COUNT: 0
[ 35108.592366] haxm_warning: 4012 VMX_ENTRY_CONTROLS: 11ff
[...]

PR #176, opened 2019-02-16
Added support for CMPS{B,W,D,Q} instructions

AlexAltea:
Changes:
- Renamed `INSN_MOV` to `INSN_DST_NR` for instructions that do not read the destination operand: Instruction flags should signal behavioral quirks of the instruction, not just an alternative way of identifying a particular instruction (*mov* in this case).
- Created `INSN_DST_NW` flag for instructions that do not read the destination operand. Name is consistent with the change mentioned above.
- Added `INSN_DST_NW` flag to *cmp*/*test* instructions as a small optimization. It will prevent unnecessary (and possibly illegal?) memory writes from occurring during these instructions.
- Implemented `cmps` in the instruction emulator, fixing #175.
- Added tests for `cmps` for both regular and REPx variants.

### DIFF
--- a/core/emulate.c
+++ b/core/emulate.c
@@ -300,7 +300,7 @@ static bool is_translation_required(struct em_context_t *ctxt)
     if (opcode->flags & INSN_TWOMEM) {
         return true;
     }
-    if (opcode->flags & INSN_STRING && ctxt->rep) {
+    if (ctxt->rep) {
         return true;
     }
     return false;
@@ -1090,6 +1090,20 @@ em_status_t EMCALL em_decode_insn(struct em_context_t *ctxt, const uint8_t *insn
     if ((opcode->flags & INSN_NOTIMPL) || !opcode->handler) {
         return EM_ERROR;
     }
+
+    if (ctxt->rep) {
+        if (!(opcode->flags & INSN_STRING)) {
+            /* Instruction does not support any REP* prefix */
+            // TODO: Should throw #UD
+            return EM_ERROR;
+        }
+        if (ctxt->rep == PREFIX_REPNE && !(opcode->flags & INSN_REPX)) {
+            /* Instruction supports REP (== REPE) but not REPNE */
+            // TODO: Should throw #UD
+            return EM_ERROR;
+        }
+    }
+
     return decode_operands(ctxt);
 }
 
@@ -1101,7 +1115,7 @@ em_status_t EMCALL em_emulate_insn(struct em_context_t *ctxt)
 
 restart:
     // TODO: Permissions, exceptions, etc.
-    if ((opcode->flags & INSN_STRING) && ctxt->rep) {
+    if (ctxt->rep) {
         if (READ_GPR(REG_RCX, ctxt->address_size) == 0) {
             goto done;
         }
@@ -1167,17 +1181,28 @@ restart:
         register_add(ctxt, REG_RSI, ctxt->operand_size *
             ((ctxt->rflags & RFLAGS_DF) ? -1LL : +1LL));
     }
-    if ((opcode->flags & INSN_STRING) && ctxt->rep) {
+    if (ctxt->rep) {
         register_add(ctxt, REG_RCX, -1LL);
-        if (opcode->flags & INSN_REP) {
-            decode_operands(ctxt);
-            goto restart;
+
+        if (opcode->flags & INSN_REPX) {
+            if ((ctxt->rep == PREFIX_REPNE && (ctxt->rflags & RFLAGS_ZF)) ||
+                (ctxt->rep == PREFIX_REPE && !(ctxt->rflags & RFLAGS_ZF))) {
+                goto done;
+            }
         }
-        if ((ctxt->rep == PREFIX_REPNE && (ctxt->rflags & RFLAGS_ZF)) ||
-            (ctxt->rep == PREFIX_REPE && !(ctxt->rflags & RFLAGS_ZF))) {
-            decode_operands(ctxt);
-            goto restart;
-        }
+
+        /* Continue to emulate the next iteration of this REP*-prefixed string
+         * instruction by jumping to the beginning of em_emulate_insn(). This
+         * avoids exiting to user space and decoding the entire instruction
+         * again. Nevertheless, we still need to rerun the operand decoders:
+         *
+         * a) To recompute the effective address for *SI and *DI operands.
+         * b) To reset the flags for each operand (OP_READ_FINISHED, etc.).
+         */
+        rc = decode_operands(ctxt);
+        if (rc != EM_CONTINUE)
+            goto exit;
+        goto restart;
     }
 
 done:
--- a/core/emulate.c
+++ b/core/emulate.c
@@ -31,24 +31,26 @@
 #include "include/emulate.h"
 
 /* Instruction flags */
-/* Instruction ignores destination original value */
-#define INSN_MOV     ((uint64_t)1 <<  0)
+/* Instruction does not read from destination */
+#define INSN_DST_NR  ((uint64_t)1 <<  0)
+/* Instruction does not write to destination */
+#define INSN_DST_NW  ((uint64_t)1 <<  1)
 /* Instruction expects ModRM byte */
-#define INSN_MODRM   ((uint64_t)1 <<  1)
+#define INSN_MODRM   ((uint64_t)1 <<  2)
 /* Instruction accesses 1-byte registers */
-#define INSN_BYTEOP  ((uint64_t)1 <<  2)
+#define INSN_BYTEOP  ((uint64_t)1 <<  3)
 /* Instruction opcode is extended via ModRM byte */
-#define INSN_GROUP   ((uint64_t)1 <<  3)
+#define INSN_GROUP   ((uint64_t)1 <<  4)
 /* Instruction supports REP prefixes */
-#define INSN_REP     ((uint64_t)1 <<  4)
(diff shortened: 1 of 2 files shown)


AlexAltea:
TravisCI build for macOS has failed, but that's unrelated to this patch.
Their servers have been lately (several weeks now!) suffering from issues while cloning remote repositories.

core/emulate.c
```diff
@@ -1169,8 +1173,8 @@ em_status_t EMCALL em_emulate_insn(struct em_context_t *ctxt)
             decode_operands(ctxt);
             goto restart;
         }
-        if ((ctxt->rep == PREFIX_REPNE && (ctxt->rflags & RFLAGS_ZF)) ||
-            (ctxt->rep == PREFIX_REPE && !(ctxt->rflags & RFLAGS_ZF))) {
+        if ((ctxt->rep == PREFIX_REPNE && !(ctxt->rflags & RFLAGS_ZF)) ||
```
raphaelning: nit: Good catch! I think it's worth mentioning in the cover page, and maybe even moving into a separate commit.

With this change, we're closer to the spec, but still not 100% compliant. Here are the condition checks given in the pseudocode for REP in Intel SDM (Vol. 2B, Section 4.3: _Instructions (M-U)_, _REP/REPE/REPZ/REPNE/REPNZ—Repeat String Operation Prefix_):

```
WHILE CountReg ≠ 0
  DO
      Service pending interrupts (if any);
      Execute associated string instruction;
      CountReg ← (CountReg – 1);
      IF CountReg = 0
          THEN exit WHILE loop; FI;
      IF (Repeat prefix is REPZ or REPE) and (ZF = 0)
      or (Repeat prefix is REPNZ or REPNE) and (ZF = 1)
          THEN exit WHILE loop; FI;
  OD;
```

Note that it checks RCX right after decrementing it. If RCX == 0, probably we should skip `decode_operands()`? If you agree, please implement it in a separate commit (which can be combined with the `ZF` condition fix).

core/emulate.c
```diff
@@ -31,24 +31,26 @@
 #include "include/emulate.h"
 
 /* Instruction flags */
-/* Instruction ignores destination original value */
-#define INSN_MOV     ((uint64_t)1 <<  0)
+/* Instruction does not read from destination */
+#define INSN_DST_NR  ((uint64_t)1 <<  0)
```
raphaelning: nit: Since it wouldn't make sense to use `DST_NW` and `DST_NR` at the same time, how about allocating 2 bits for this flag (or property)?

```c
enum {
    DST_READWRITE = 0,  /* default */
    DST_WRITEONLY,
    DST_READONLY,
    /* DST_NONE is invalid */
}
```

This is not a required change, but if you are willing to do it, please also create macros for getting and setting this 2-bit flag.

raphaelning:
> TravisCI build for macOS has failed, but that's unrelated to this patch.
> Their servers have been lately (several weeks now!) suffering from issues while cloning remote repositories.

Are they aware of this issue? It looks like a git config problem. Quoting the [haxm-darwin build log](https://travis-ci.com/intel/haxm/jobs/178285153):

```
Cloning into '/usr/local/Homebrew/Library/Taps/homebrew/homebrew-bundle'...
fatal: unknown value for config 'protocol.version': 2
```

See also https://stackoverflow.com/questions/53207973/fatal-unknown-value-for-config-protocol-version-2

I don't know who sets `protocol.version` to 2 (probably in a git config file), but if we can override it (e.g. using an environment variable), we should be able to work around the error.

core/emulate.c
AlexAltea: My reasoning is that instruction flags are used to signal quirks and other less-common behaviors. Many x86 instructions read and write the destination register (e.g. `add eax, ecx`), consequently the less-common behavior is **not** reading/writing it, otherwise we would be making everything unnecessarily verbose by flagging most instructions with `DST_READWRITE`.

Note that by having "not-read" and "not-write" flags we also save ourselves from having to merge these two under a single "not_readwrite" flag, because no instruction satisfies that property.

core/emulate.c
AlexAltea: I've fixed this in 3ee3b48. Note that `decode_operands` does not access memory: it's called just to update the `ea` field for `op_di` and `op_si` operands, so that the correct memory address is read later on. Then after `goto restart` we just check whether RCX != 0, so I'd say the current implementation complies with the specification (unless I'm missing something else?).

Nonetheless I've moved the `decode_operands` stage right after the RCX != 0 check.

AlexAltea:
> Are they aware of this issue?

No idea. I'll check it, and report that if necessary.

core/emulate.c
raphaelning: Ah, you're right! I just blindly assumed that `decode_operands` would access memory, sorry. But the new patch does look cleaner :-)

core/emulate.c
raphaelning: Sure, that works, thanks for explaining it in detail! I also wanted to avoid flagging most instructions with `DST_READWRITE`, which is why I assigned it the zero value... But I agree that we should be consistent and stick to the one-bit-per-flag design.

raphaelning:
The [`git-config` manual](https://git-scm.com/docs/git-config#git-config-protocolversion) doesn't mention any environment variable that can override `protocol.version`. Moreover, I'm not sure if the Travis `homebrew` addon allows us to customize the runtime environment for `brew`.

raphaelning:
Hmm, with this PR, `qemu-system-x86_64 -accel hax` is stuck at a black screen. I'm trying to figure out why.

core/emulate.c
```diff
@@ -1105,6 +1105,8 @@ em_status_t EMCALL em_emulate_insn(struct em_context_t *ctxt)
         if (READ_GPR(REG_RCX, ctxt->address_size) == 0) {
             goto done;
         }
+        // Update memory operands
+        decode_operands(ctxt);
```
raphaelning: One way to fix the black screen issue is to move this line back to where it was (before every `goto restart`), but that's suboptimal. There must be something in `decode_operands` that makes it unsafe to be called at the first iteration of REP...

raphaelning: I think the problem is that `decode_operands` clears `em_operand_t::flags` for destination and source operands, which can be nonzero (e.g. `OP_READ_PENDING`) when we re-enter `em_emulate_insn` for the same iteration of the same REP instruction, due to a previous `operand_read/write` attempt that incurred an exit to QEMU. In the case of re-entrance, there's no need to update `operand_mem_t::ea`, since DI/SI has not changed.

raphaelning: But it's okay to clear those flags toward the end of `em_emulate_insn`, because at that point we have completed emulating the instruction (or one iteration of it).

raphaelning: In fact, we must make sure those operand flags are reset before jumping to `restart`, or `operand_read/write` will end up doing nothing. So we still want to call `decode_operands()` right before `goto restart`.

Sorry for the fuss! I've made a patch to fix it and will push it shortly. I hope it still offers a slight improvement over the original logic.

raphaelning:
@AlexAltea Let me know what you think of 23ca2a0 :-) It's intended to be merged into 3ee3b48.

AlexAltea:
@raphaelning I fully agree with your changes. In fact, I had just started modifying my PR on the same lines before realizing you had submitted it already yesterday.

Note that you commit also fixes another minor inconvenience with my approach: That `decode_operands` could be called twice on the before the first iteration: once on `em_decode_insn` and once at the beginning of `em_emulate_insn`. So yeah, looks good to me!

raphaelning:
Cool! Let me rebase and merge the last two commits.

AlexAltea:
Also, it's a shame that the tests didn't catch this bug (otherwise I wouldn't have submitted my last commit).

Probably the emulator tests should be improved to handle memory accesses similarly to how HAXM performs MMIO-accesses, i.e. returning *OP_READ/WRITE_PENDING* first, doing the query, and then returning the result alongside *OP_READ/WRITE_FINISHED*.

raphaelning:
Good point! Yes, looking at `test_read_memory()` vs. `vcpu_read_memory()`, it's clear that the former never returns `EM_EXIT_MMIO`.

raphaelning:
Regarding the Homebrew failure, I noticed that our Travis CI config uses `osx_image: xcode10`, whereas the default is `xcode9.4` and the latest is `xcode10.1`:

https://docs.travis-ci.com/user/reference/osx/#macos-version

I wonder if switching to either can fix the issue. @AlexAltea Is there any reason not to use the default `osx_image`?

AlexAltea:
> Is there any reason not to use the default `osx_image`?

@raphaelning Not that I know of. We can remove that line, and simply use Travis' default setting, although I don't think that will fix the bug. I'll contact Travis support today to see if they can have a look at it.

raphaelning:
Thanks, let me test the default setting.

raphaelning:
> Issue submitted @ ` [4399] - Issues while fetching Homebrew packages`.

Thanks! Even though #184 has worked around the issue for us, it would be nice if Travis CI could fix it properly for everybody.

AlexAltea:
Just for the record: The issue has been addressed by the Travis team:

> We made an attempt to enable Git Wire Protocol v2 for more efficient git cloning/fetching on Travis CI but we unfortunately ran into some unsupported edge cases (like yours) with our first implementation. We have proceeded with reverting the change and your builds should be back on their feet with a restart. Sorry for the inconvenience and please let me know how it goes.

Anyway, I prefer building with default Travis settings, so better keeping our build scripts as in #184.

PR #178, opened 2019-02-19
[netbsd] Correct retrieving the CPU index

krytarowski:
curcpu()->ci_cpuid is not guaranteed to return incremental indices 0..N-1.

This caused crashes on some NetBSD machines.

Patch and report by Andreas Gustafsson (NetBSD).


krytarowski:
Andreas Gustafsson wanted me to submit this myself.

PR #185, opened 2019-02-22
 Added support for PUSH/POP instructions

### LINKED ISSUE #182: push dword [0xd096c080] raises decoder failure
**Describe the Bug**
Summary: 

Attempting to run OpenBSD/i386 6.4 raises a decoder failure:

[ 20026,827747] haxm_panic: em_decode_insn() failed: vcpu_id=0, len=6, CS:IP=0x0:0xd03a4930, instr[0..5]=0xff 0x35 0x80 0xc0 0x96 0xd0

```
$ rasm2 -d '0xff 0x35 0x80 0xc0 0x96 0xd0'
push dword [0xd096c080]
```

**Host Environment**

NetBSD chieftec 8.99.34 NetBSD 8.99.34 (GENERIC) #1: Sun Feb 17 11:51:34 CET 2019  root@chieftec:/public/netbsd-root/sys/arch/amd64/compile/GENERIC amd64

* HAXM version: 

5fdebb50322b9fe0d801a8d0c80013a1e455cfa2 + cherry-picked patch https://github.com/intel/haxm/issues/172#issuecomment-465471085

* Host OS version: 
NetBSD 8.99.34 amd64

x86_64

* Host CPU model: 
Intel(R) Core(TM) i7-3770 CPU @ 3.40GHz

* Host RAM size: 
32GB RAM
desktop

**Guest Environment**

* Android Emulator or QEMU version: 

qemu-3.0.0nb5 from pkgsrc

* Guest OS version: 
6.4

* Guest OS architecture: 
i386

**To Reproduce**

qemu-system-x86_64 -cdrom cd64.iso  -accel hax -curses

**Expected Behavior**
Booting OpenBSD 6.4 to a shell prompt

Expectation: 
Add support for missing instruction in the simulator.

**Reproducibility**
Always.

**Diagnostic Information**

```
[ 19998,997087] haxm_warning: hax_alloc_pages: HAX_MEM_LOW_4G is ignored
[ 20026,827747] haxm_panic: em_decode_insn() failed: vcpu_id=0, len=6, CS:IP=0x0:0xd03a4930, instr[0..5]=0xff 0x35 0x80 0xc0 0x96 0xd0
[ 20026,827747] haxm_warning: 4000 VMX_PIN_CONTROLS: 1f
[ 20026,827747] haxm_warning: 4002 VMX_PRIMARY_PROCESSOR_CONTROLS: 969861fa
[ 20026,827747] haxm_warning: 401e VMX_SECONDARY_PROCESSOR_CONTROLS: aa
[ 20026,827747] haxm_warning: 4004 VMX_EXCEPTION_BITMAP: 40000
[ 20026,827747] haxm_warning: 4006 VMX_PAGE_FAULT_ERROR_CODE_MASK: 0
[ 20026,827747] haxm_warning: 4008 VMX_PAGE_FAULT_ERROR_CODE_MATCH: 0
[ 20026,827747] haxm_warning: 400c VMX_EXIT_CONTROLS: 236fff
[ 20026,827747] haxm_warning: 400e VMX_EXIT_MSR_STORE_COUNT: 0
[ 20026,827747] haxm_warning: 4010 VMX_EXIT_MSR_LOAD_COUNT: 0
[ 20026,827747] haxm_warning: 4012 VMX_ENTRY_CONTROLS: 11ff
[ 20026,827747] haxm_warning: 4014 VMX_ENTRY_MSR_LOAD_COUNT: 0
[ 20026,827747] haxm_warning: 4016 VMX_ENTRY_INTERRUPT_INFO: b0
[ 20026,827747] haxm_warning: 4018 VMX_ENTRY_EXCEPTION_ERROR_CODE: 0
[ 20026,827747] haxm_warning: 401a VMX_ENTRY_INSTRUCTION_LENGTH: 0
[ 20026,827747] haxm_warning: 401c VMX_TPR_THRESHOLD: 0
[ 20026,827747] haxm_warning: 6000 VMX_CR0_MASK: ffffffffe0000020
[ 20026,827747] haxm_warning: 6002 VMX_CR4_MASK: ffffffffffe8f860
[ 20026,827747] haxm_warning: 6004 VMX_CR0_READ_SHADOW: 8000003f
[ 20026,827747] haxm_warning: 6006 VMX_CR4_READ_SHADOW: a0
[ 20026,827747] haxm_warning: 400a VMX_CR3_TARGET_COUNT: 0
[ 20026,827747] haxm_warning: 6008 VMX_CR3_TARGET_VAL_BASE: 0
[ 20026,827747] haxm_warning: 0000 VMX_VPID: 1
[ 20026,827747] haxm_warning: 2000 VMX_IO_BITMAP_A: 13ad59000
[ 20026,827747] haxm_warning: 2002 VMX_IO_BITMAP_B: 13ad5a000
[ 20026,827747] haxm_warning: 2004 VMX_MSR_BITMAP: 13ad5d000
[ 20026,827747] haxm_warning: 2006 VMX_EXIT_MSR_STORE_ADDRESS: 0
[ 20026,827747] haxm_warning: 2008 VMX_EXIT_MSR_LOAD_ADDRESS: 0
[ 20026,827747] haxm_warning: 200a VMX_ENTRY_MSR_LOAD_ADDRESS: 0
[ 20026,827747] haxm_warning: 2010 VMX_TSC_OFFSET: ffffc23b68fc2894
[ 20026,827747] haxm_warning: 2012 VMX_VAPIC_PAGE: 0
[ 20026,827747] haxm_warning: 2014 VMX_APIC_ACCESS_PAGE: 0
[ 20026,827747] haxm_warning: 201a VMX_EPTP: 44257e01e
[ 20026,827747] haxm_warning: 482e VMX_PREEMPTION_TIMER: 0
[ 20026,827747] haxm_warning: 4400 VMX_INSTRUCTION_ERROR_CODE: 0
[ 20026,827747] haxm_warning: 4402 VM_EXIT_INFO_REASON: 30
[ 20026,827747] haxm_warning: 4404 VM_EXIT_INFO_INTERRUPT_INFO: 0
[ 20026,827747] haxm_warning: 4406 VM_EXIT_INFO_EXCEPTION_ERROR_CODE: 0
[ 20026,827747] haxm_warning: 4408 VM_EXIT_INFO_IDT_VECTORING: 0
[ 20026,827747] haxm_warning: 440a VM_EXIT_INFO_IDT_VECTORING_ERROR_CODE: 0
[ 20026,827747] haxm_warning: 440c VM_EXIT_INFO_INSTRUCTION_LENGTH: 6
[ 20026,827747] haxm_warning: 440e VM_EXIT_INFO_INSTRUCTION_INFO: 0
[ 20026,827747] haxm_warning: 6400 VM_EXIT_INFO_QUALIFICATION: 181
[ 20026,827747] haxm_warning: 6402 VM_EXIT_INFO_IO_ECX: 4d0
[ 20026,827747] haxm_warning: 6404 VM_EXIT_INFO_IO_ESI: ffffa502587c6e5c
[ 20026,827747] haxm_warning: 6406 VM_EXIT_INFO_IO_EDI: 408
[ 20026,827747] haxm_warning: 6408 VM_EXIT_INFO_IO_EIP: d039d2c7
[ 20026,827747] haxm_warning: 640a VM_EXIT_INFO_GUEST_LINEAR_ADDRESS: d096c080
[ 20026,827747] haxm_warning: 2400 VM_EXIT_INFO_GUEST_PHYSICAL_ADDRESS: fee00080
[ 20026,827747] haxm_warning: 6c16 HOST_RIP: ffffffff81cc0233
[ 20026,827747] haxm_warning: 6c14 HOST_RSP: ffffa5025d9e4be0
[ 20026,827747] haxm_warning: 6c00 HOST_CR0: 80050033
[ 20026,827747] haxm_warning: 6c02 HOST_CR3: 7f78f0000
[ 20026,827747] haxm_warning: 6c04 HOST_CR4: 1426b0
[ 20026,827747] haxm_warning: 0c02 HOST_CS_SELECTOR: 8
[ 20026,827747] haxm_warning: 0c06 HOST_DS_SELECTOR: 20
[ 20026,827747] haxm_warning: 0c00 HOST_ES_SELECTOR: 20
[ 20026,827747] haxm_warning: 0c08 HOST_FS_SELECTOR: 0
[ 20026,827747] haxm_warning: 0c0a HOST_GS_SELECTOR: 0
[ 20026,827747] haxm_warning: 0c04 HOST_SS_SELECTOR: 10
[ 20026,827747] haxm_warning: 0c0c HOST_TR_SELECTOR: e0
[ 20026,827747] haxm_warning: 6c06 HOST_FS_BASE: 7f7ff7a2ba28
[ 20026,827747] haxm_warning: 6c08 HOST_GS_BASE: ffffa502587c2000
[ 20026,827747] haxm_warning: 6c0a HOST_TR_BASE: ffffff000006f000
[ 20026,827747] haxm_warning: 6c0c HOST_GDTR_BASE: ffffff000005f000
[ 20026,827747] haxm_warning: 6c0e HOST_IDTR_BASE: ffffff0000001000
[ 20026,827747] haxm_warning: 4c00 HOST_SYSENTER_CS: 0
[ 20026,827747] haxm_warning: 6c10 HOST_SYSENTER_ESP: 0
[ 20026,827747] haxm_warning: 6c12 HOST_SYSENTER_EIP: 0
[ 20026,827747] haxm_warning: 681e GUEST_RIP: d03a4930
[ 20026,827747] haxm_warning: 6820 GUEST_RFLAGS: 10046
[ 20026,827747] haxm_warning: 681c GUEST_RSP: d4d01978
[ 20026,827747] haxm_warning: 6800 GUEST_CR0: 8001003b
[ 20026,827747] haxm_warning: 6802 GUEST_CR3: 7d3eb40
[ 20026,827747] haxm_warning: 6804 GUEST_CR4: 26e0
[ 20026,827747] haxm_warning: 0800 GUEST_ES_SELECTOR: 10
[ 20026,827747] haxm_warning: 0802 GUEST_CS_SELECTOR: 50
[ 20026,827747] haxm_warning: 0804 GUEST_SS_SELECTOR: 10
[ 20026,827747] haxm_warning: 0806 GUEST_DS_SELECTOR: 10
[ 20026,827747] haxm_warning: 0808 GUEST_FS_SELECTOR: 20
[ 20026,827747] haxm_warning: 080a GUEST_GS_SELECTOR: 0
[ 20026,827747] haxm_warning: 080c GUEST_LDTR_SELECTOR: 0
[ 20026,827747] haxm_warning: 080e GUEST_TR_SELECTOR: 68
[ 20026,827747] haxm_warning: 4814 GUEST_ES_AR: c093
[ 20026,827747] haxm_warning: 4816 GUEST_CS_AR: c09b
[ 20026,827747] haxm_warning: 4818 GUEST_SS_AR: c093
[ 20026,827747] haxm_warning: 481a GUEST_DS_AR: c093
[ 20026,827747] haxm_warning: 481c GUEST_FS_AR: 93
[ 20026,827747] haxm_warning: 481e GUEST_GS_AR: 1c000
[ 20026,827747] haxm_warning: 4820 GUEST_LDTR_AR: 1c000
[ 20026,827747] haxm_warning: 4822 GUEST_TR_AR: 8b
[ 20026,827747] haxm_warning: 6806 GUEST_ES_BASE: 0
[ 20026,827747] haxm_warning: 6808 GUEST_CS_BASE: 0
[ 20026,827747] haxm_warning: 680a GUEST_SS_BASE: 0
[ 20026,827747] haxm_warning: 680c GUEST_DS_BASE: 0
[ 20026,827747] haxm_warning: 680e GUEST_FS_BASE: d0973ff8
[ 20026,827747] haxm_warning: 6810 GUEST_GS_BASE: 0
[ 20026,827747] haxm_warning: 6812 GUEST_LDTR_BASE: 0
[ 20026,827747] haxm_warning: 6814 GUEST_TR_BASE: d0972000
[ 20026,827747] haxm_warning: 6816 GUEST_GDTR_BASE: d09720d0
[ 20026,827747] haxm_warning: 6818 GUEST_IDTR_BASE: d09e5000
[ 20026,827747] haxm_warning: 4800 GUEST_ES_LIMIT: ffffffff
[ 20026,827747] haxm_warning: 4802 GUEST_CS_LIMIT: ffffffff
[ 20026,827747] haxm_warning: 4804 GUEST_SS_LIMIT: ffffffff
[ 20026,827747] haxm_warning: 4806 GUEST_DS_LIMIT: ffffffff
[ 20026,827747] haxm_warning: 4808 GUEST_FS_LIMIT: 453
[ 20026,827747] haxm_warning: 480a GUEST_GS_LIMIT: ffffffff
[ 20026,827747] haxm_warning: 480c GUEST_LDTR_LIMIT: ffffffff
[ 20026,827747] haxm_warning: 480e GUEST_TR_LIMIT: 67
[ 20026,827747] haxm_warning: 4810 GUEST_GDTR_LIMIT: 7f
[ 20026,827747] haxm_warning: 4812 GUEST_IDTR_LIMIT: 7ff
[ 20026,827747] haxm_warning: 2800 GUEST_VMCS_LINK_PTR: ffffffffffffffff
[ 20026,827747] haxm_warning: 2802 GUEST_DEBUGCTL: 0
[ 20026,827747] haxm_warning: 2804 GUEST_PAT: 0
[ 20026,827747] haxm_warning: 2806 GUEST_EFER: 800
[ 20026,827747] haxm_warning: 2808 GUEST_PERF_GLOBAL_CTRL: 0
[ 20026,827747] haxm_warning: 280a GUEST_PDPTE0: 7d3d001
[ 20026,827747] haxm_warning: 280c GUEST_PDPTE1: 7d3c001
[ 20026,827747] haxm_warning: 280e GUEST_PDPTE2: 7d3b001
[ 20026,827747] haxm_warning: 2810 GUEST_PDPTE3: 7d3a001
[ 20026,827747] haxm_warning: 681a GUEST_DR7: 400
[ 20026,827747] haxm_warning: 6822 GUEST_PENDING_DBE: 0
[ 20026,827747] haxm_warning: 482a GUEST_SYSENTER_CS: 0
[ 20026,827747] haxm_warning: 6824 GUEST_SYSENTER_ESP: 0
[ 20026,827747] haxm_warning: 6826 GUEST_SYSENTER_EIP: 0
[ 20026,827747] haxm_warning: 4828 GUEST_SMBASE: 0
[ 20026,827747] haxm_warning: 4824 GUEST_INTERRUPTIBILITY: 0
[ 20026,827747] haxm_warning: 4826 GUEST_ACTIVITY_STATE: 0
[ 20026,827747] haxm_error: vcpu has panicked, id:0
[ 20026,827747] haxm_error: log_host_cr4_vmxe: 0
[ 20026,827747] haxm_error: log_host_cr4 0
[ 20026,827747] haxm_error: log_vmxon_res 0
[ 20026,827747] haxm_error: log_vmxon_addr 13ad62000
[ 20026,827747] haxm_error: log_vmxon_err_type1 0
[ 20026,827747] haxm_error: log_vmxon_err_type2 0
[ 20026,827747] haxm_error: log_vmxon_err_type3 0
[ 20026,827747] haxm_error: log_vmclear_err 0
[ 20026,827747] haxm_error: log_vmptrld_err 0
[ 20026,827747] haxm_error: log_vmoff_no 0
[ 20026,827747] haxm_error: log_vmxoff_res 0
[ 20026,827747] haxm_error: vcpu has panicked, id:0
[ 20026,827747] haxm_error: log_host_cr4_vmxe: 0
[ 20026,827747] haxm_error: log_host_cr4 0
[ 20026,827747] haxm_error: log_vmxon_res 0
[ 20026,827747] haxm_error: log_vmxon_addr 13ad62000
[ 20026,827747] haxm_error: log_vmxon_err_type1 0
[ 20026,827747] haxm_error: log_vmxon_err_type2 0
[ 20026,827747] haxm_error: log_vmxon_err_type3 0
[ 20026,827747] haxm_error: log_vmclear_err 0
[ 20026,827747] haxm_error: log_vmptrld_err 0
[ 20026,827747] haxm_error: log_vmoff_no 0
[ 20026,827747] haxm_error: log_vmxoff_res 0
[ 20026,827747] haxm_error: vcpu has panicked, id:0
[ 20026,827747] haxm_error: log_host_cr4_vmxe: 0
[ 20026,827747] haxm_error: log_host_cr4 0
[ 20026,827747] haxm_error: log_vmxon_res 0
[ 20026,827747] haxm_error: log_vmxon_addr 13ad62000
[ 20026,827747] haxm_error: log_vmxon_err_type1 0
[ 20026,827747] haxm_error: log_vmxon_err_type2 0
[ 20026,827747] haxm_error: log_vmxon_err_type3 0
[ 20026,827747] haxm_error: log_vmclear_err 0
[ 20026,827747] haxm_error: log_vmptrld_err 0
[ 20026,827747] haxm_error: log_vmoff_no 0
[ 20026,827747] haxm_error: log_vmxoff_res 0
[ 20026,827747] haxm_error: vcpu has panicked, id:0
[ 20026,827747] haxm_error: log_host_cr4_vmxe: 0
[ 20026,827747] haxm_error: log_host_cr4 0
[ 20026,827747] haxm_error: log_vmxon_res 0
[ 20026,827747] haxm_error: log_vmxon_addr 13ad62000
[ 20026,827747] haxm_error: log_vmxon_err_type1 0
[ 20026,827747] haxm_error: log_vmxon_err_type2 0
[ 20026,827747] haxm_error: log_vmxon_err_type3 0
[ 20026,827747] haxm_error: log_vmclear_err 0
[ 20026,827747] haxm_error: log_vmptrld_err 0
[ 20026,827747] haxm_error: log_vmoff_no 0
[ 20026,827747] haxm_error: log_vmxoff_res 0
[ 20026,827747] haxm_error: 
[ 20026,827747] ...........hax_teardown_vm
```

Host crash dump: 
N/A

AlexAltea:
This pull request fixes #182 (ping @krytarowski) and includes several changes:

### Added a register cache to the instruction emulator (c799252)

This implements a cache for reading/writing GPRs in the instruction emulator: Register reads will be stored in ctxt->gpr_cache, and writes will be stored in ctxt->gpr_cache until they are explicitly flushed by `gpr_cache_flush` so that the actual register access, as defined in `em_vcpu_ops_t`, is performed.

Aside from minor performance gains, this change will make easier to introduce instructions with "side effects", e.g. push/pop modifying the stack pointer. The reason is that due to our model of PENDING/FINISHED flags to handle memory accesses, the logic of `em_emulate_insn` is executed few times, but we don't want to apply these "side effects" few times, e.g. decrementing/incrementing the stack pointer at every execution.

By placing a `gpr_cache_flush`, we "commit" the changes on the "done" label, and ensure that side effects occur only once. Note that for REP/REPE/REPNE instructions we also need to flush the GPR cache on each iteration, as such instructions are meant to execute the emulation logic multiple times, e.g. to repeatedly decrement the ~accumulator~ counter.

### Verbose error descriptions in emulator tests (25478d7)

Previously, we added `EXPECT_EQ(memcmp(&vcpu, &vcpu_expected, sizeof(test_cpu_t)), 0);` as a simple check at the end of EmulatorTest::run. While this was enough to catch implementation issues in the emulator, it did specify where the exactly was the mismatch happening. Consequently, we often turn to the debugger and manually diff both `test_cpu_t` structures which is time consuming.

This patch adds a dedicated `EmulatorTest::verify` function that checks each register individually and reports any issues to the stderr, before comparing and asserting the equality of both structures. Note that the previous `EXPECT_EQ` has been replaced with `ASSERT_EQ`, as there is no point in keep trying other instruction variants if there's one broken already.

### Allow emulator tests in 16-bit and 32-bit protected mode (7860a02)

Now we can customize in the `em_context_t::mode` in each test, and provide the `run_prot16`, `run_prot32`, `run_prot64` methods to assemble, decode and emulate instructions. The previous `run` method now defaults to `run_prot64`.

### Added support for PUSH/POP instructions (6d962ac)

- Implemented push/pop instructions in the emulator.
- Added `INSN_STACK` instruction flag to signal push/pop instructions, since the operand size behaves differently than in other instructions.
- Returning `em_status_t` from software-emulated instructions to allow calling operand_read/operand_write within them.
- Adding `ctxt->b` to the emulation context members, since certain variants of push/pop require reading the opcode byte to get the register index.
- Added tests for push/pop instructions.

### DIFF
--- a/core/emulate.c
+++ b/core/emulate.c
@@ -1182,13 +1182,6 @@ em_status_t EMCALL em_decode_insn(struct em_context_t *ctxt, const uint8_t *insn
     }
 
     /* Apply flags */
-    if (flags & INSN_BYTEOP) {
-        ctxt->operand_size = 1;
-    }
-    if (flags & INSN_STACK) {
-        if (ctxt->operand_size == 4 && ctxt->mode == EM_MODE_PROT64)
-            ctxt->operand_size = 8;
-    }
     if (flags & INSN_GROUP) {
         const struct em_opcode_t *group_opcode;
 
@@ -1199,6 +1192,7 @@ em_status_t EMCALL em_decode_insn(struct em_context_t *ctxt, const uint8_t *insn
         group_opcode = &opcode->group[ctxt->modrm.opc];
         opcode->handler = group_opcode->handler;
         opcode->flags |= group_opcode->flags;
+        flags = opcode->flags;
         if (group_opcode->decode_dst != decode_op_none) {
             opcode->decode_dst = group_opcode->decode_dst;
         }
@@ -1208,6 +1202,13 @@ em_status_t EMCALL em_decode_insn(struct em_context_t *ctxt, const uint8_t *insn
         if (group_opcode->decode_src2 != decode_op_none) {
             opcode->decode_src2 = group_opcode->decode_src2;
         }
+    }
+    if (flags & INSN_BYTEOP) {
+        ctxt->operand_size = 1;
+    }
+    if (flags & INSN_STACK) {
+        if (ctxt->operand_size == 4 && ctxt->mode == EM_MODE_PROT64)
+            ctxt->operand_size = 8;
     }
 
     /* Some unimplemented opcodes have an all-zero em_opcode_t */
--- a/core/emulate.c
+++ b/core/emulate.c
@@ -81,14 +81,14 @@
 
 /* Emulator ops */
 #define READ_GPR(idx, size) \
-    ctxt->ops->read_gpr(ctxt->vcpu, idx, size)
+    gpr_read(ctxt, idx, size)
 #define WRITE_GPR(idx, value, size) \
-    ctxt->ops->write_gpr(ctxt->vcpu, idx, value, size)
+    gpr_write(ctxt, idx, size, value)
 
-#define BX (uint16_t)(ctxt->ops->read_gpr(ctxt->vcpu, REG_RBX, 2))
-#define BP (uint16_t)(ctxt->ops->read_gpr(ctxt->vcpu, REG_RBP, 2))
-#define SI (uint16_t)(ctxt->ops->read_gpr(ctxt->vcpu, REG_RSI, 2))
-#define DI (uint16_t)(ctxt->ops->read_gpr(ctxt->vcpu, REG_RDI, 2))
+#define BX (uint16_t)(gpr_read(ctxt, REG_RBX, 2))
+#define BP (uint16_t)(gpr_read(ctxt, REG_RBP, 2))
+#define SI (uint16_t)(gpr_read(ctxt, REG_RSI, 2))
+#define DI (uint16_t)(gpr_read(ctxt, REG_RDI, 2))
 
 /* Operand decoders */
 #define DECL_DECODER(name) \
@@ -291,6 +291,41 @@ static const struct em_opcode_t opcode_table_0F3A[256] = {
     X16(N), X16(N), X16(N), X16(N),
     X16(N), X16(N), X16(N), X16(N),
 };
+
+static uint64_t gpr_read(struct em_context_t *ctxt,
+                         unsigned index, size_t size)
+{
+    uint64_t value = 0;
+
+    if (!(ctxt->gpr_cache_r & (1 << index))) {
+        ctxt->gpr_cache[index] = ctxt->ops->read_gpr(ctxt->vcpu, index);
+        ctxt->gpr_cache_r |= (1 << index);
+    }
+    memcpy(&value, &ctxt->gpr_cache[index], size);
+    return value;
+}
+
+static void gpr_write(struct em_context_t *ctxt,
+                      unsigned index, size_t size, uint64_t value)
+{
+    if (!(ctxt->gpr_cache_r & (1 << index))) {
+        ctxt->gpr_cache[index] = ctxt->ops->read_gpr(ctxt->vcpu, index);
+    }
+    ctxt->gpr_cache_r |= (1 << index);
+    ctxt->gpr_cache_w |= (1 << index);
+    memcpy(&ctxt->gpr_cache[index], &value, size);
+}
+
+static void gpr_cache_flush(struct em_context_t *ctxt)
+{
+    unsigned i;
+
+    for (i = 0; i < 16; i++) {
+        if (ctxt->gpr_cache_w & (1 << i))
+            ctxt->ops->write_gpr(ctxt->vcpu, i, ctxt->gpr_cache[i]);
+    }
+    ctxt->gpr_cache_w = 0;
+}
 
 /* Emulate accesses to guest memory */
 static bool is_translation_required(struct em_context_t *ctxt)
@@ -956,6 +991,8 @@ em_status_t EMCALL em_decode_insn(struct em_context_t *ctxt, const uint8_t *insn
     default:
         return EM_ERROR;
     }
+    ctxt->gpr_cache_r = 0;
+    ctxt->gpr_cache_w = 0;
     ctxt->override_segment = SEG_NONE;
     ctxt->override_operand_size = 0;
     ctxt->override_address_size = 0;
@@ -1202,6 +1239,7 @@ restart:
         rc = decode_operands(ctxt);
         if (rc != EM_CONTINUE)
             goto exit;
+        gpr_cache_flush(ctxt);
         goto restart;
     }
 
@@ -1209,6 +1247,7 @@ done:
     rc = EM_CONTINUE;
     ctxt->finished = true;
     ctxt->ops->advance_rip(ctxt->vcpu, ctxt->len);
+    gpr_cache_flush(ctxt);
 
 exit:
     return rc;
--- a/core/emulate.c
+++ b/core/emulate.c
@@ -51,6 +51,8 @@
 #define INSN_TWOMEM  ((uint64_t)1 <<  8)
 /* Instruction takes bit test operands */
 #define INSN_BITOP   ((uint64_t)1 <<  9)
+/* Instruction accesses the stack */
+#define INSN_STACK   ((uint64_t)1 << 10)
 /* String instruction */
 #define INSN_STRING  (INSN_REP|INSN_REPX)
 
@@ -102,6 +104,7 @@ DECL_DECODER(op_vex_reg);
 DECL_DECODER(op_moffs);
 DECL_DECODER(op_simm);
 DECL_DECODER(op_simm8);
+DECL_DECODER(op_reg);
 DECL_DECODER(op_acc);
 DECL_DECODER(op_di);
 DECL_DECODER(op_si);
@@ -144,12 +147,14 @@ DECL_DECODER(op_si);
     F2_BV(_handler, op_acc, op_simm, op_none, (_flags))
 
 /* Soft-emulation */
-static void em_andn_soft(struct em_context_t *ctxt);
-static void em_bextr_soft(struct em_context_t *ctxt);
-static void em_mov(struct em_context_t *ctxt);
-static void em_movzx(struct em_context_t *ctxt);
-static void em_movsx(struct em_context_t *ctxt);
-static void em_xchg(struct em_context_t *ctxt);
+static em_status_t em_andn_soft(struct em_context_t *ctxt);
+static em_status_t em_bextr_soft(struct em_context_t *ctxt);
+static em_status_t em_mov(struct em_context_t *ctxt);
+static em_status_t em_movzx(struct em_context_t *ctxt);
+static em_status_t em_movsx(struct em_context_t *ctxt);
+static em_status_t em_push(struct em_context_t *ctxt);
+static em_status_t em_pop(struct em_context_t *ctxt);
+static em_status_t em_xchg(struct em_context_t *ctxt);
 
 static const struct em_opcode_t opcode_group1[8] = {
     F(em_add, op_none, op_none, op_none, 0),
@@ -169,6 +174,13 @@ static const struct em_opcode_t opcode_group3[8] = {
     F(em_neg, op_modrm_rm, op_none, op_none, 0),
 };
 
+static const struct em_opcode_t opcode_group5[8] = {
+    X4(N),
+    X2(N),
+    I(em_push, op_none, op_modrm_rm, op_none, INSN_TWOMEM | INSN_STACK),
+    X1(N),
+};
+
 static const struct em_opcode_t opcode_group8[8] = {
     X4(N),
     F(em_bt, op_none, op_none, op_none, 0),
@@ -179,6 +191,10 @@ static const struct em_opcode_t opcode_group8[8] = {
 
 static const struct em_opcode_t opcode_group11[8] = {
     I(em_mov, op_none, op_none, op_none, INSN_DST_NR),
+};
+
+static const struct em_opcode_t opcode_group1A[8] = {
+    I(em_pop, op_modrm_rm, op_none, op_none, INSN_DST_NR | INSN_TWOMEM | INSN_STACK),
 };
 
 static const struct em_opcode_t opcode_table[256] = {
@@ -202,8 +218,11 @@ static const struct em_opcode_t opcode_table[256] = {
     X8(F(em_inc, op_modrm_reg, op_none, op_none, 0)),
     /* 0x48 - 0x4F */
     X8(F(em_dec, op_modrm_reg, op_none, op_none, 0)),
-    /* 0x50 - 0x7F */
-    X16(N), X16(N), X16(N),
+    /* 0x50 - 0x5F */
+    X8(I(em_push, op_none, op_reg, op_none, INSN_STACK)),
+    X8(I(em_pop,  op_reg, op_none, op_none, INSN_STACK)),
+    /* 0x60 - 0x7F */
+    X16(N), X16(N),
     /* 0x80 - 0x8F */
     G(opcode_group1, op_modrm_rm, op_simm, op_none, INSN_BYTEOP),
     G(opcode_group1, op_modrm_rm, op_simm, op_none, 0),
@@ -213,7 +232,8 @@ static const struct em_opcode_t opcode_table[256] = {
     X2(N),  /* TODO: 0x86 & 0x87 (XCHG) */
     I2_BV(em_mov, op_modrm_rm, op_modrm_reg, op_none, INSN_MODRM | INSN_DST_NR),
     I2_BV(em_mov, op_modrm_reg, op_modrm_rm, op_none, INSN_MODRM | INSN_DST_NR),
-    X4(N),
+    N, N, N,
+    G(opcode_group1A, op_none, op_none, op_none, 0),
     /* 0x90 - 0x9F */
     X16(N),
     /* 0xA0 - 0xAF */
@@ -240,7 +260,10 @@ static const struct em_opcode_t opcode_table[256] = {
     X2(N),
     G(opcode_group3, op_none, op_none, op_none, INSN_BYTEOP),
     G(opcode_group3, op_none, op_none, op_none, 0),
-    X8(N),
+    X4(N),
+    X2(N),
+    X1(N),
+    G(opcode_group5, op_none, op_none, op_none, 0),
 };
 
 static const struct em_opcode_t opcode_table_0F[256] = {
@@ -869,6 +892,15 @@ static em_status_t decode_op_simm8(em_context_t *ctxt,
     return EM_CONTINUE;
 }
 
+static em_status_t decode_op_reg(em_context_t *ctxt,
+    em_operand_t *op)
+{
+    op->type = OP_REG;
+    op->size = ctxt->operand_size;
+    op->reg.index = ctxt->b & 0x7;
+    return EM_CONTINUE;
+}
+
 static em_status_t decode_op_acc(em_context_t *ctxt,
                                  em_operand_t *op)
 {
@@ -902,7 +934,7 @@ static em_status_t decode_op_si(em_context_t *ctxt,
 }
 
 /* Soft-emulation */
-static void em_andn_soft(struct em_context_t *ctxt)
+static em_status_t em_andn_soft(struct em_context_t *ctxt)
 {
     uint64_t temp;
     int signbit = 63;
@@ -915,9 +947,10 @@ static void em_andn_soft(struct em_context_t *ctxt)
     ctxt->rflags &= ~RFLAGS_MASK_OSZAPC;
     ctxt->rflags |= (temp == 0) ? RFLAGS_ZF : 0;
     ctxt->rflags |= (temp >> signbit) ? RFLAGS_SF : 0;
+    return EM_CONTINUE;
 }
 
-static void em_bextr_soft(struct em_context_t *ctxt)
+static em_status_t em_bextr_soft(struct em_context_t *ctxt)
 {
     uint8_t start, len;
     uint64_t temp;
@@ -933,20 +966,23 @@ static void em_bextr_soft(struct em_context_t *ctxt)
     ctxt->dst.value = temp;
     ctxt->rflags &= ~RFLAGS_MASK_OSZAPC;
     ctxt->rflags |= (temp == 0) ? RFLAGS_ZF : 0;
+    return EM_CONTINUE;
 }
 
-static void em_mov(struct em_context_t *ctxt)
+static em_status_t em_mov(struct em_context_t *ctxt)
 {
     memcpy(&ctxt->dst.value, &ctxt->src1.value, ctxt->operand_size);
+    return EM_CONTINUE;
 }
 
-static void em_movzx(struct em_context_t *ctxt)
+static em_status_t em_movzx(struct em_context_t *ctxt)
 {
     ctxt->dst.value = 0;
     memcpy(&ctxt->dst.value, &ctxt->src1.value, ctxt->src1.size);
+    return EM_CONTINUE;
 }
 
-static void em_movsx(struct em_context_t *ctxt)
+static em_status_t em_movsx(struct em_context_t *ctxt)
 {
     uint64_t extension = 0;
     if (ctxt->src1.value & (1ULL << ((8 * ctxt->src1.size) - 1))) {
@@ -955,17 +991,63 @@ static void em_movsx(struct em_context_t *ctxt)
     ctxt->dst.value = 0;
     memcpy(&ctxt->dst.value, &extension, ctxt->dst.size);
     memcpy(&ctxt->dst.value, &ctxt->src1.value, ctxt->src1.size);
+    return EM_CONTINUE;
 }
 
-static void em_xchg(struct em_context_t *ctxt)
+static em_status_t em_push(struct em_context_t *ctxt)
 {
+    em_status_t rc;
+    em_operand_t sp;
+
+    // Pre-decrement stack pointer
+    register_add(ctxt, REG_RSP, -(int64_t)ctxt->operand_size);
+
+    memset(&sp, 0, sizeof(sp));
+    sp.type = OP_MEM;
+    sp.size = ctxt->operand_size;
+    sp.mem.ea = READ_GPR(REG_RSP, ctxt->address_size);
+    sp.mem.seg = SEG_SS;
+    sp.value = ctxt->src1.value;
+
+    rc = operand_write(ctxt, &sp);
+    return rc;
+}
+
+static em_status_t em_pop(struct em_context_t *ctxt)
+{
+    em_status_t rc;
+    em_operand_t sp;
+
+    memset(&sp, 0, sizeof(sp));
+    sp.type = OP_MEM;
+    sp.size = ctxt->operand_size;
+    sp.mem.ea = READ_GPR(REG_RSP, ctxt->address_size);
+    sp.mem.seg = SEG_SS;
+    sp.value = ctxt->src1.value;
+
+    // Post-increment stack pointer
+    register_add(ctxt, REG_RSP, +(int64_t)ctxt->operand_size);
+
+    rc = operand_read(ctxt, &sp);
+    if (rc != EM_CONTINUE)
+        return rc;
+    ctxt->dst.value = sp.value;
+    return rc;
+}
+
+static em_status_t em_xchg(struct em_context_t *ctxt)
+{
+    em_status_t rc;
     uint64_t src1, src2;
+
     src1 = ctxt->src1.value;
     src2 = ctxt->src2.value;
     ctxt->src1.value = src2;
     ctxt->src2.value = src1;
-    operand_write(ctxt, &ctxt->src1);
-    operand_write(ctxt, &ctxt->src2);
+    rc = operand_write(ctxt, &ctxt->src1);
+    if (rc != EM_CONTINUE)
+        return rc;
+    return operand_write(ctxt, &ctxt->src2);
 }
 
 em_status_t EMCALL em_decode_insn(struct em_context_t *ctxt, const uint8_t *insn)
@@ -1091,6 +1173,7 @@ em_status_t EMCALL em_decode_insn(struct em_context_t *ctxt, const uint8_t *insn
         }
         break;
     }
+    ctxt->b = b;
 
     /* Intel SDM Vol. 2A: 2.1.3 ModR/M and SIB Bytes */
     flags = opcode->flags;
@@ -1101,6 +1184,10 @@ em_status_t EMCALL em_decode_insn(struct em_context_t *ctxt, const uint8_t *insn
     /* Apply flags */
     if (flags & INSN_BYTEOP) {
         ctxt->operand_size = 1;
+    }
+    if (flags & INSN_STACK) {
+        if (ctxt->operand_size == 4 && ctxt->mode == EM_MODE_PROT64)
+            ctxt->operand_size = 8;
     }
     if (flags & INSN_GROUP) {
         const struct em_opcode_t *group_opcode;
@@ -1195,9 +1282,11 @@ restart:
         ctxt->rflags &= ~RFLAGS_MASK_OSZAPC;
         ctxt->rflags |= eflags & RFLAGS_MASK_OSZAPC;
     } else {
-        void (*soft_handler)(em_context_t*);
+        em_status_t (*soft_handler)(em_context_t*);
         soft_handler = opcode->handler;
-        soft_handler(ctxt);
+        rc = soft_handler(ctxt);
+        if (rc != EM_CONTINUE)
+            goto exit;
     }
 
     // Output operands
--- a/core/emulate.c
+++ b/core/emulate.c
@@ -82,11 +82,6 @@
 #define X16(...)  X8(__VA_ARGS__), X8(__VA_ARGS__)
 
 /* Emulator ops */
-#define READ_GPR(idx, size) \
-    gpr_read(ctxt, idx, size)
-#define WRITE_GPR(idx, value, size) \
-    gpr_write(ctxt, idx, size, value)
-
 #define BX (uint16_t)(gpr_read(ctxt, REG_RBX, 2))
 #define BP (uint16_t)(gpr_read(ctxt, REG_RBP, 2))
 #define SI (uint16_t)(gpr_read(ctxt, REG_RSI, 2))
@@ -333,8 +328,8 @@ static void gpr_write(struct em_context_t *ctxt,
 {
     if (!(ctxt->gpr_cache_r & (1 << index))) {
         ctxt->gpr_cache[index] = ctxt->ops->read_gpr(ctxt->vcpu, index);
+        ctxt->gpr_cache_r |= (1 << index);
     }
-    ctxt->gpr_cache_r |= (1 << index);
     ctxt->gpr_cache_w |= (1 << index);
     memcpy(&ctxt->gpr_cache[index], &value, size);
 }
@@ -456,7 +451,7 @@ static em_status_t operand_read_bitop(struct em_context_t *ctxt)
         return rc;
     }
 
-    offset = READ_GPR(offs->reg.index, offs->size);
+    offset = gpr_read(ctxt, offs->reg.index, offs->size);
     switch (offs->size) {
     case 2:
         offset = (int16_t)offset;
@@ -493,7 +488,7 @@ static em_status_t operand_read(struct em_context_t *ctxt,
         rc = EM_CONTINUE;
         break;
     case OP_REG:
-        op->value = READ_GPR(op->reg.index, op->size);
+        op->value = gpr_read(ctxt, op->reg.index, op->size);
         rc = EM_CONTINUE;
         break;
     case OP_MEM:
@@ -532,7 +527,7 @@ static em_status_t operand_write(struct em_context_t *ctxt,
         break;
     case OP_REG:
         size = (op->size == 4) ? 8 : op->size;
-        WRITE_GPR(op->reg.index, op->value, size);
+        gpr_write(ctxt, op->reg.index, size, op->value);
         rc = EM_CONTINUE;
         break;
     case OP_MEM:
@@ -556,9 +551,10 @@ static em_status_t operand_write(struct em_context_t *ctxt,
 }
 
 static void register_add(struct em_context_t *ctxt,
-                         int reg_index, uint64_t value)
+                         int reg_index, uint64_t addend)
 {
-    WRITE_GPR(reg_index, READ_GPR(reg_index, 8) + value, 8);
+    uint64_t value = gpr_read(ctxt, reg_index, 8) + addend;
+    gpr_write(ctxt, reg_index, 8, value);
 }
 
 static uint8_t insn_fetch_u8(struct em_context_t *ctxt)
@@ -777,19 +773,19 @@ static em_status_t decode_op_modrm_rm(em_context_t *ctxt,
             if ((reg_base & 7) == 5 && ctxt->modrm.mod == 0) {
                 op->mem.ea += insn_fetch_s32(ctxt);
             } else {
-                op->mem.ea += READ_GPR(reg_base, ctxt->address_size);
+                op->mem.ea += gpr_read(ctxt, reg_base, ctxt->address_size);
             }
             /* Added scaled index register unless register is RSP/ESP/SP.
              * Note that we avoid masking with 0x7, as that would prevent
              * R12/R12D/R12W from being a valid index register. */
             if (reg_index != 4) {
                 uint8_t scale = 1 << ctxt->sib.scale;
-                op->mem.ea += READ_GPR(reg_index, ctxt->address_size) * scale;
+                op->mem.ea += gpr_read(ctxt, reg_index, ctxt->address_size) * scale;
             }
         } else if (ctxt->modrm.mod == 0 && ctxt->modrm.rm == 5) {
             op->mem.ea += insn_fetch_s32(ctxt);
         } else {
-            op->mem.ea += READ_GPR(ctxt->modrm.rm, ctxt->address_size);
+            op->mem.ea += gpr_read(ctxt, ctxt->modrm.rm, ctxt->address_size);
         }
 
         // Displacement
@@ -915,7 +911,7 @@ static em_status_t decode_op_di(em_context_t *ctxt,
 {
     op->type = OP_MEM;
     op->size = ctxt->operand_size;
-    op->mem.ea = READ_GPR(REG_RDI, ctxt->address_size);
+    op->mem.ea = gpr_read(ctxt, REG_RDI, ctxt->address_size);
     op->mem.seg = SEG_ES;
     return EM_CONTINUE;
 }
@@ -925,7 +921,7 @@ static em_status_t decode_op_si(em_context_t *ctxt,
 {
     op->type = OP_MEM;
     op->size = ctxt->operand_size;
-    op->mem.ea = READ_GPR(REG_RSI, ctxt->address_size);
+    op->mem.ea = gpr_read(ctxt, REG_RSI, ctxt->address_size);
     op->mem.seg = SEG_DS;
     if (ctxt->override_segment != SEG_NONE) {
         op->mem.seg = ctxt->override_segment;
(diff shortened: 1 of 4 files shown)


AlexAltea:
Also, I've just found a small bug. It didn't affect any previous instruction, but it did affect `pop` instructions in the current implementation. It has been fixed here:

### Fixed missing flags in grouped instructions (9d7a927)

After finding the opcode structure corresponding to the current instruction, we cache `opcode->flags` into the `flags` variable and perform several checks for particular flags such as `INSN_BYTEOP` and `INSN_STACK`.

The issue here is that for instructions that belong in a group, see `INSN_GROUP`, we need to merge the group flags and the flags of the actual opcode before performing such queries. Consequently we need to reposition all `(flags & INSN_*)` statements after we have dealt with groups.

The only exception here is `(flags & INSN_MODRM)` since we need to fetch the ModRM byte in order to figure out which instruction we are dealing with within that group.

krytarowski:
This pushes booting OpenBSD/i386 6.4 further but another instruction is missing in the emulator and I will send a dedicated report for it.

tests/test_emulator.cpp
```diff
@@ -200,13 +200,58 @@ class EmulatorTest : public testing::Test {
...
+        VERIFY_GPR(REG_R9);
+        VERIFY_GPR(REG_R10);
+        VERIFY_GPR(REG_R11);
+        VERIFY_GPR(REG_R12);
+        VERIFY_GPR(REG_R13);
+        VERIFY_GPR(REG_R14);
+        VERIFY_GPR(REG_R15);
+
+        // Verify RFLAGS
+        if (vcpu_obtained.flags != vcpu_expected.flags) \
```
raphaelning: nit: Remove the trailing ` \` from these lines.

tests/test_emulator.cpp
```diff
@@ -200,13 +200,58 @@ class EmulatorTest : public testing::Test {
...
+            std::cerr << "Flags mismatch on: " << insn << "\n" \
+                << "vcpu_obtained.flags: 0x" << PRINT_U64(vcpu_obtained.flags) << "\n" \
+                << "vcpu_expected.flags: 0x" << PRINT_U64(vcpu_expected.flags) << "\n";
+
+#undef DEBUG_U64
```
raphaelning: => `PRINT_U64`

tests/test_emulator.cpp
```diff
@@ -200,13 +200,58 @@ class EmulatorTest : public testing::Test {
         em_status_t ret = EM_ERROR;
 
         vcpu = vcpu_original;
-        assemble_decode(insn, vcpu.rip, &size, &count, &ret);
-        ASSERT_TRUE(ret != EM_ERROR);
+        assemble_decode(mode, insn, vcpu.rip, &size, &count, &ret);
+        ASSERT_TRUE(ret != EM_ERROR)
```
raphaelning: nit: I probably abused `ASSERT_TRUE` and `EXPECT_TRUE` when I last modified this function. Later I realized that the `_EQ` macros would be a better fit for value comparisons, because they would give us more helpful information in case of assertion failure. Please feel free to address this problem in this patch.

tests/test_emulator.cpp
```diff
@@ -168,33 +171,47 @@ class EmulatorTest : public testing::Test {
...
                          size_t* size,
                          size_t* count,
                          em_status_t* decode_status) {
         uint8_t* code;
         int err;
 
-        err = ks_asm(ks, insn, 0, &code, size, count);
+        switch (mode) {
+        case EM_MODE_PROT16:
+            err = ks_asm(ks_x86_16, insn, 0, &code, size, count);
```
raphaelning: nit: Instead of calling `ks_asm()` once per case, create a local `ks_engine*` and make it the focus of the switch-case.

core/emulate.c
```diff
@@ -81,14 +81,14 @@
 
 /* Emulator ops */
 #define READ_GPR(idx, size) \
```
PatrickvL: The need for `READ_GPR` is very much deminished by this commit - consider removing it entirely.

core/emulate.c
```diff
@@ -81,14 +81,14 @@
...
-    ctxt->ops->read_gpr(ctxt->vcpu, idx, size)
+    gpr_read(ctxt, idx, size)
 #define WRITE_GPR(idx, value, size) \
```
PatrickvL: The need for `WRITE_GPR` is very much deminished by this commit - consider removing it entirely.

core/emulate.c
```diff
@@ -292,6 +292,41 @@ static const struct em_opcode_t opcode_table_0F3A[256] = {
...
+
+static void gpr_cache_flush(struct em_context_t *ctxt)
+{
+    unsigned i;
+
+    for (i = 0; i < 16; i++) {
+        if (ctxt->gpr_cache_w & (1 << i))
+            ctxt->ops->write_gpr(ctxt->vcpu, i, ctxt->gpr_cache[i]);
+    }
+    ctxt->gpr_cache_w = 0;
```
PatrickvL: Because `ctxt->gpr_cache_r` isn't reset here, wouldn't that cause potential registers vs. cache mismatches, on re-entry of `em_emulate_insn` when a prior call returned `OP_WRITE_PENDING`?
Same question for `OP_READ_PENDING`?

AlexAltea:
By the way, I forgot to mention how the current verbose errors (during VCPU checks) look like:

```
Register mismatch on: btr [ecx + 0x08], eax
vcpu_obtained.gpr[REG_RAX]: 0xFFFFFFFFFFFFFFF1
vcpu_expected.gpr[REG_RAX]: 0x0000000000000000
Register mismatch on: btr [ecx + 0x08], eax
vcpu_obtained.gpr[REG_RDX]: 0x0000000000000000
vcpu_expected.gpr[REG_RDX]: 0x000000000000FFFF
Expected: memcmp(&vcpu, &vcpu_expected, sizeof(test_cpu_t))
      Which is: 1
To be equal to: 0
```

where previously it was simply:

```
Expected: memcmp(&vcpu, &vcpu_expected, sizeof(test_cpu_t))
      Which is: 1
To be equal to: 0
```

which was not very informative.

tests/test_emulator.cpp
AlexAltea: > Instead of calling `ks_asm()` once per case

But we need to assemble each test case, hence calling `ks_asm` once per case is unavoidable. Or am I misunderstanding something?

> defer the call until after the switch case by creating a local `ks_engine*`.

Wouldn't that hurt performance? That way we'd have to create, configure and destroy a new Keystone engine at every instruction we want to test, whereas currently we simply create 3 engines and switch between them based on the emulator mode.

PatrickvL: @raphaelning Do you mean this switch could just select a `ks_engine*`, to be used in one call to `ks_asm()` below the switch/case statement block?

core/emulate.c
AlexAltea: No, because *OP_WRITE_PENDING* and *OP_READ_PENDING* will make the corresponding `operand_read` and `operand_write` functions return with `EM_EXIT_MMIO` status, which then triggers a `goto exit`. Note that `gpr_cache_flush` is never called after the "exit" label.

AlexAltea:
All issues fixed @ 6c27fc2.

tests/test_emulator.cpp
raphaelning: @PatrickvL Exactly, thanks for clarifying that for me. By "case" I mean a case in the switch-case statement, not a GTest test case... To be absolutely clear:

```c
        ks_engine* ks = nullptr;

        switch (mode) {
        case EM_MODE_PROT16:
            ks = ks_x86_16;
            break;
        // 3 more cases
        }
        ASSERT_EQ(KS_ERR_OK, ks_asm(ks_x86_16, insn, 0, &code, size, count));
```

core/emulate.c
raphaelning: At first I also frowned upon the lack of explicit invalidation/reset of the read cache, besides the "initialization" at the beginning of `em_decode_insn()`. One possible option is to invalidate it also in `vcpu.c` right before the re-entry of `em_emulate_insn()`, in case user space (QEMU) updates any vCPU GPR behind our back. But then I scratched the idea, because QEMU's handler for `HAX_EXIT_FAST_MMIO` (`hax_handle_fastmmio()`) doesn't modify vCPU state. So invalidating the read cache only before decoding every instruction should be good enough.

tests/test_emulator.cpp
raphaelning: Sorry, by "these lines" I mean this line and the 2 lines that follow it :)

tests/test_emulator.cpp
PatrickvL: Surely you mean `ASSERT_EQ(KS_ERR_OK, ks_asm(ks, insn, 0, &code, size, count));`

AlexAltea: Oh, I fully agree with that. I've modified my last commit to include this change.

Now it's at 43d5fde.

AlexAltea:
@raphaelning I've updated my last commit to reflect what you said at: https://github.com/intel/haxm/pull/185#discussion_r259390717.

tests/test_emulator.cpp
```diff
@@ -181,25 +181,26 @@ class EmulatorTest : public testing::Test {
                          size_t* size,
                          size_t* count,
                          em_status_t* decode_status) {
+        ks_engine* ks = nullptr;
         uint8_t* code;
         int err;
```
PatrickvL: With the last commit, `err` isn't used anymore and should be removed...

AlexAltea:
@PatrickvL Good catch, thank you. It's fixed on [`3105e83`](/intel/haxm/commit/3105e8396d9b399ca5f6804aca40e74045a9f9db).

krytarowski:
What host platforms are supported?

Hardware and software breakpoints don't work on NetBSD so far.

AlexAltea:
@krytarowski What do you mean? Maybe you are referring to a different issue/pull request?

krytarowski:
> @krytarowski What do you mean? Maybe you are referring to a different issue/pull request?

Wrong haxm tab! My older message is deleted.

core/emulate.c
```diff
@@ -920,17 +987,63 @@ static void em_movsx(struct em_context_t *ctxt)
...
+
+static em_status_t em_pop(struct em_context_t *ctxt)
+{
+    em_status_t rc;
+    em_operand_t sp;
+
+    memset(&sp, 0, sizeof(sp));
+    sp.type = OP_MEM;
+    sp.size = ctxt->operand_size;
+    sp.mem.ea = gpr_read(ctxt, REG_RSP, ctxt->address_size);
```
raphaelning: There's a detail in the PUSH/POP spec about which \*SP register (SP, ESP or RSP) should be used, i.e. the stack-address size, which can be different from `ctxt->address_size`:

> Stack-address size. Outside of 64-bit mode, the B flag in the current stack-segment descriptor determines the
size of the stack pointer (16 or 32 bits); in 64-bit mode, the size of the stack pointer is always 64 bits.
> The stack-address size determines the width of the stack pointer when reading from the stack in memory and
when incrementing the stack pointer. (As stated above, the amount by which the stack pointer is incremented
is determined by the operand size.)

To be fully compliant, I believe we need to check the B (Big) bit of the VMCS `GUEST_SS_AR` field (cf. Intel SDM Vol. 3A, Section 3.4.5: _Segment Descriptors_). If that requires too much additional effort, please leave a TODO comment instead.

core/emulate.c
```diff
@@ -920,17 +987,63 @@ static void em_movsx(struct em_context_t *ctxt)
...
+    em_status_t rc;
+    em_operand_t sp;
+
+    // Pre-decrement stack pointer
+    register_add(ctxt, REG_RSP, -(int64_t)ctxt->operand_size);
+
+    memset(&sp, 0, sizeof(sp));
+    sp.type = OP_MEM;
+    sp.size = ctxt->operand_size;
+    sp.mem.ea = gpr_read(ctxt, REG_RSP, ctxt->address_size);
```
raphaelning: See my other comment about using the stack-address size instead of address size.

core/emulate.c
```diff
@@ -920,17 +987,63 @@ static void em_movsx(struct em_context_t *ctxt)
...
+    sp.mem.seg = SEG_SS;
+    sp.value = ctxt->src1.value;
+
+    rc = operand_write(ctxt, &sp);
```
raphaelning: Reusing `operand_write()` to emulate the memory copy is really neat! There's just a little problem: the temporary `sp` operand is not preserved across re-entry of `em_emulate_insn()`, so we'll run into an infinite loop of pending MMIO writes if we are to emulate a push to MMIO (although that's very unlikely). How about just adding the implicit destination operand (`sp`) to `em_context_t`?

core/emulate.c
```diff
@@ -920,17 +987,63 @@ static void em_movsx(struct em_context_t *ctxt)
...
+    sp.type = OP_MEM;
+    sp.size = ctxt->operand_size;
+    sp.mem.ea = gpr_read(ctxt, REG_RSP, ctxt->address_size);
+    sp.mem.seg = SEG_SS;
+    sp.value = ctxt->src1.value;
+
+    // Post-increment stack pointer
+    register_add(ctxt, REG_RSP, +(int64_t)ctxt->operand_size);
+
+    rc = operand_read(ctxt, &sp);
```
raphaelning: See my other comment about adding `sp` to `em_context_t`.

core/emulate.c
```diff
@@ -920,17 +987,63 @@ static void em_movsx(struct em_context_t *ctxt)
...
+
+    memset(&sp, 0, sizeof(sp));
+    sp.type = OP_MEM;
+    sp.size = ctxt->operand_size;
+    sp.mem.ea = gpr_read(ctxt, REG_RSP, ctxt->address_size);
+    sp.mem.seg = SEG_SS;
+    sp.value = ctxt->src1.value;
+
+    // Post-increment stack pointer
+    register_add(ctxt, REG_RSP, +(int64_t)ctxt->operand_size);
```
raphaelning: The spec requires us to increment after copy. But IIUC, it doesn't really matter if we call `register_add()` before or after `operand_read()`, because the actual register is not updated until `gpr_cache_flush()`. That's cool!

core/emulate.c
```diff
@@ -920,17 +987,63 @@ static void em_movsx(struct em_context_t *ctxt)
...
+{
+    em_status_t rc;
+    em_operand_t sp;
+
+    memset(&sp, 0, sizeof(sp));
+    sp.type = OP_MEM;
+    sp.size = ctxt->operand_size;
+    sp.mem.ea = gpr_read(ctxt, REG_RSP, ctxt->address_size);
+    sp.mem.seg = SEG_SS;
+    sp.value = ctxt->src1.value;
```
raphaelning: Delete this line, because POP doesn't use a source operand ;-)

tests/test_emulator.cpp
```diff
@@ -171,45 +171,132 @@ class EmulatorTest : public testing::Test {
...
+        VERIFY_GPR(REG_R10);
+        VERIFY_GPR(REG_R11);
+        VERIFY_GPR(REG_R12);
+        VERIFY_GPR(REG_R13);
+        VERIFY_GPR(REG_R14);
+        VERIFY_GPR(REG_R15);
+
+        // Verify RFLAGS
+        if (vcpu_obtained.flags != vcpu_expected.flags)
+            std::cerr << "Flags mismatch on: " << insn << "\n" \
```
raphaelning: nit: Remove the trailing ` \`

tests/test_emulator.cpp
```diff
@@ -171,45 +171,132 @@ class EmulatorTest : public testing::Test {
...
+                << "vcpu_obtained.flags: 0x" << PRINT_U64(vcpu_obtained.flags) << "\n" \
```
raphaelning: nit: Remove the trailing ` \`

core/emulate.c
AlexAltea: Good idea, thank you! I've done that, although I've reused existing slots, i.e. `ctxt->dst` for *em_push*, and `ctxt->src1` for *em_pop*.

core/emulate.c
AlexAltea: Yeah, after setting the appropriate `sp->mem.ea`, we can change anything we want.

core/emulate.c
AlexAltea: Sorry, that was a typo from copy-pasting the *em_push* above. I've deleted it.

core/emulate.c
```diff
@@ -1085,6 +1199,13 @@ em_status_t EMCALL em_decode_insn(struct em_context_t *ctxt, const uint8_t *insn
             opcode->decode_src2 = group_opcode->decode_src2;
         }
     }
+    if (flags & INSN_BYTEOP) {
+        ctxt->operand_size = 1;
+    }
+    if (flags & INSN_STACK) {
+        if (ctxt->operand_size == 4 && ctxt->mode == EM_MODE_PROT64)
+            ctxt->operand_size = 8;
```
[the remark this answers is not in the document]
AlexAltea: Because in `EM_MODE_PROT64` mode, no `INSN_STACK` instruction (i.e. push/pop) can be done with 32-bit operands. The rules that apply to all other instructions default to 32-bit in this case, see:
https://github.com/intel/haxm/blob/1bb4c59/core/emulate.c#L952-L955

Consequently we need to override the operand size in this specific case.

core/emulate.c
AlexAltea: Nice catch, thanks a lot for your feedback! I wasn't aware of this. I'll place a *TODO* for now, since it requires rather big changes to the vcpu-emulator interface.

AlexAltea:
@raphaelning All fixed!

Since my fixes to `decode_op_reg` (now called `decode_op_opc_reg`) involved adding support for AH/CH/DH/BH registers, I've decided to cherry pick the patch from #191 into this branch, and add all requested changes on top of it.

core/include/emulate.h
```diff
@@ -142,6 +140,7 @@ typedef struct em_operand_t {
         } mem;
         struct operand_reg_t {
             uint32_t index;
+            uint32_t shift;
```
raphaelning: nit: Be consistent with the type of `shift`. `gpr_read_shifted()` uses `size_t`.

core/emulate.c
```diff
@@ -292,6 +310,58 @@ static const struct em_opcode_t opcode_table_0F3A[256] = {
     X16(N), X16(N), X16(N), X16(N),
 };
 
+static uint64_t gpr_read_shifted(struct em_context_t *ctxt,
+                                 unsigned index, size_t size, size_t shift)
+{
+    uint64_t value = 0;
+    uint8_t *value_ptr;
+
```
raphaelning: nit: Add `assert()` statements to detect potential buffer overflows:

* `index < 16`
* `size + shift <= 8`

core/emulate.c
```diff
@@ -292,6 +310,58 @@ static const struct em_opcode_t opcode_table_0F3A[256] = {
...
+    value_ptr = (uint8_t*)&ctxt->gpr_cache[index];
+    memcpy(&value, &value_ptr[shift], size);
+    return value;
+}
+
+static void gpr_write_shifted(struct em_context_t *ctxt,
+                              unsigned index, size_t size, size_t value, uint64_t shift)
+{
+    uint8_t *value_ptr;
+
```
raphaelning: nit: Add `assert()` statements to detect potential buffer overflows:

* `index < 16`
* `size + shift <= 8`

core/emulate.c
```diff
@@ -772,6 +859,22 @@ static em_status_t decode_op_vex_reg(em_context_t *ctxt,
...
+    op->reg.shift = 0;
+    return EM_CONTINUE;
+}
+
+static em_status_t decode_op_opc_reg(em_context_t *ctxt,
+                                     em_operand_t *op)
+{
+    op->type = OP_REG;
+    op->size = ctxt->operand_size;
+    op->reg.index = ctxt->b & 0x7 | (ctxt->rex.b << 3);
```
raphaelning: This line fails the `haxm-darwin` Travis CI build. Obviously, clang requires us to write `(ctxt->b & 0x7)`:

```
/Users/travis/build/intel/haxm/core/emulate.c:871:29: error: '&' within '|' [-Werror,-Wbitwise-op-parentheses]
    op->reg.index = ctxt->b & 0x7 | (ctxt->rex.b << 3);
                    ~~~~~~~~^~~~~ ~
/Users/travis/build/intel/haxm/core/emulate.c:871:29: note: place parentheses around the '&' expression to silence this warning
    op->reg.index = ctxt->b & 0x7 | (ctxt->rex.b << 3);
                            ^
                    (            )
1 error generated.
```

core/emulate.c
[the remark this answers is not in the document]
raphaelning: I see! For the record, this exception is specified in Intel SDM Vol. 2A, Section 2.2.1.7: _Default 64-Bit Operand Size_. Also, https://wiki.osdev.org/X86-64_Instruction_Encoding#Usage gives a summary of all instructions that default to 64-bit operand size in 64-bit mode.

core/include/emulate.h
AlexAltea: Instead, I've modified the previous function definitions to make their `shift` argument `uint32_t`.

core/emulate.c
```diff
@@ -292,6 +310,58 @@ static const struct em_opcode_t opcode_table_0F3A[256] = {
     X16(N), X16(N), X16(N), X16(N),
 };
 
+static uint64_t gpr_read_shifted(struct em_context_t *ctxt,
+                                 unsigned index, size_t size, size_t shift)
```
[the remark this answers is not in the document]
AlexAltea: This line is still under 80 characters, what should I do here? For now I've just modify it just a bit.

AlexAltea:
@raphaelning All fixed!

AlexAltea:
I had to remove the asserts from the emulator. Since `emulate.c` is shared both in kernel and user tests, managing dependencies there is a nightmare. In any case, we already do the index checks in *ctxt->ops->read_gpr* and *ctxt->ops->write_gpr*, and the size+shift issue is realistically not going to happen.

core/emulate.c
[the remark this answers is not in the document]
raphaelning: Sorry, my bad!

core/include/emulate.h
raphaelning: OK, that works too.

raphaelning:
Thanks!

AlexAltea:
> Could you split it? I think the cosmetic changes can stay in this PR, and the rest go into a separate PR (like the way you did at first)

@raphaelning I was reluctant to split the commit, because the last commit contained changes without which PUSH/POP emulation was essentially broken. I've investigated the cause of this issue, and turns out I didn't to invalidate the read-flags from the GPR caches after a MMIO exit.

krytarowski:
Hopefully it's now all fine. And right we have now quite many things depending on this PR, better to not create extra work of splitting it up.

raphaelning:
@AlexAltea Thanks. The new change (`gpr_cache_invalidate()`) looks good, although I wonder why it is necessary (more on this later). But first of all, are you sure 8f64789 has fixed the issue? I've pulled the latest PR, but I still get the same black screen when booting Ubuntu ISO with virtio-net... The problem goes away as soon as I revert the last commit (`git revert 8f67489`).

@krytarowski I asked @AlexAltea to split the last commit because I thought that would allow us to merge the first 6 (good) commits now, as well as make it easier for us to identify the offending change that causes the Ubuntu ISO regression. Of course, if someone can fix this PR without splitting it, I'll be more than happy to merge the fixed PR.

core/emulate.c
```diff
@@ -1202,14 +1373,17 @@ em_status_t EMCALL em_emulate_insn(struct em_context_t *ctxt)
...
     }
 
 done:
     rc = EM_CONTINUE;
     ctxt->finished = true;
     ctxt->ops->advance_rip(ctxt->vcpu, ctxt->len);
+    gpr_cache_flush(ctxt);
 
 exit:
+    gpr_cache_invalidate(ctxt);
```
raphaelning: I suppose this is mainly to reset the GPR cache for the `EM_EXIT_MMIO` (re-entry to `em_emulate_insn()` for the same instruction) case, because the decoder already resets the GPR cache for every new instruction. But do you have an example that shows QEMU may indeed touch a guest GPR between `HAX_EXIT_FAST_MMIO` and the subsequent `HAX_VCPU_IOCTL_RUN`?

yiqisiben:

yiqisiben:

AlexAltea:
@raphaelning Sorry for the late reply. Last month was incredibly busy, and had no time to go thoroughly debug this issue. Turns out this was a subtle bug introduced while modifying the `decode_op_modrm_reg` and `decode_op_modrm_rm` operand decoders.

This happened probably because I misunderstood the specification. If you look at the table "*Table 3-1.  Register Codes Associated With +rb, +rw, +rd, +ro*", in the REX.B column, and AH/BH/CH/DH rows it's said that these are not "*Not Encodable*". However this actually means that no REX prefix is allowed at all, and not just that REX.B is not allowed (otherwise it would say just "*None*", I suppose).

One of the guest instructions that were executed in your Ubuntu image shows this issue:
```asm
40 88 3E        mov  byte ptr [rsi], dil
```

Here the REX prefix is 0x40, meaning rex.b=0, rex.x=0, rex.r=0, rex.w=0. With the previous implementation this meant that the instruction was decoded as: `mov  byte ptr [rsi], bh` instead. Now the entire REX value is tested, not just the relevant bit, as you can see in the [diff](https://github.com/intel/haxm/compare/8f6478982f410328351c6cfa9632df719330f5f6..bd7439f7d2db8cfbc63b7d0f2b2c98da01cc027b).

polprog:
With this diff applied to intel/master, OpenBSD/i386 is failing to boot up on NetBSD for me

The guest system fails in a later stage in kernel bootup (after disk initialization etc.)
The error in the guest console is:
```uvm_fault(0xd0e87e04, 0x5d26000, 0, 1) -> e
uvm_fault(0xd09bbad8, 0x4d92000, 0, 1) -> e
```,
then many of those messages:
```fatal page fault (6)trap [swapper]0/0 type 4: sp d02e2397 not inside 0-0```
after a second or so, the guest system resets back into bootloader and the same thing happens again, but without the page fault messages visible (it would initialize up to the same point, then reset to bootloader again)

There aren't any messages in dmesg.
I have checked the addresses in the guest kernel log against the kernel executable with gdb, `0xd0e87e04` is not a valid address:
```(gdb) disas 0xd0e87e04,+10
Dump of assembler code from 0xd0e87e04 to 0xd0e87e0e:
   0xd0e87e04:	Cannot access memory at address 0xd0e87e04```
and `0xd09bbad8` is in a range full of `add    %al,(%eax)`, ie. zero bytes:
```$ rasm2 add    '%al,(%eax)'
00```

**Comment by AlexAltea** [comment]
@polprog Is OpenBSD/i386 working before **without** this patch on intel/master? Just asking to figure out whether the patch is ready to be merged, or whether you found a regression and further work is necessary.

**Comment by krytarowski** [comment]
> @polprog Is OpenBSD/i386 working before **without** this patch on intel/master? Just asking to figure out whether the patch is ready to be merged, or whether you found a regression and further work is necessary.

It didn't work earlier.. however it's not possible to state whether this patch is helping or not. Assuming that there are no setup mistakes by polprog, at least we no longer see undefined instruction fault.

**Comment by krytarowski** [comment]
@AlexAltea I propose to check it yourself with OpenBSD/i386. Changing unspecified instruction to vm fault is suspicious.

**Comment by polprog** [comment]
@AlexAltea With this patch, openbsd/i386 boots farther than ever

**Comment by AlexAltea** [comment]
@krytarowski According to @polprog this patch causes no known regressions, and it allows OpenBSD/i386 to boot further. If there's any other issue beyond that point I'd suggest to simply create another issue. This patch was simply meant to fix #182, and should be ready to be merged.

**Comment by krytarowski** [comment]
OK, so as @polprog claims that it breaks now much later we can stop blocking it.

**Comment by raphaelning** [comment]
All: Sorry about my late reply! I was first on vacation and then in an internal workshop. Now that I'm back to clear my backlog of HAXM work, I'm glad to see the great progress made on this PR. Thanks @AlexAltea for finally fixing the regression. I'll try to finish the review ASAP, but it's been a while, so please give me a couple of days.

**Comment by raphaelning** [comment]
> This happened probably because I misunderstood the specification. If you look in the Intel SDM Vol. 2A at the table "Table 3-1. Register Codes Associated With +rb, +rw, +rd, +ro", in the REX.B column, and AH/BH/CH/DH rows it's said that these are not "Not Encodable". However this actually means that no REX prefix is allowed at all, and not just that REX.B is not allowed (otherwise it would say just "None", I suppose).

Yes, I also find this table very confusing. For me, the register encoding table from https://wiki.osdev.org/X86-64_Instruction_Encoding#Registers is a much better reference. I believe it's applicable to all instruction fields that encode a GPR: modrm.reg, opcode.reg, and modrm.rm (when modrm.mod = 11b).

Also, the following paragraph from Intel SDM Vol. 1, Section 3.4.1.1 (_General-Purpose Registers in 64-Bit Mode_) gives further explanation as to why AH/BH/CH/DH are sometimes not encodable in 64-bit mode:

> In 64-bit mode, there are limitations on accessing byte registers. An instruction cannot reference legacy highbytes (for example: AH, BH, CH, DH) and one of the new byte registers at the same time (for example: the low byte of the RAX register). However, instructions may reference legacy low-bytes (for example: AL, BL, CL or DL) and new byte registers at the same time (for example: the low byte of the R8 register, or RBP). The architecture enforces this limitation by changing high-byte references (AH, BH, CH, DH) to low byte references (BPL, SPL, DIL, SIL: the low 8 bits for RBP, RSP, RDI and RSI) for instructions using a REX prefix.

At first glance, I thought the new `insn_mov` test added by bd7439f violated the above rule, because it references both CH and RDX in 64-bit mode (`run_prot64()`):

```asm
mov ch, [rdx]
```

But that's actually a valid instruction, because the encoding of RDX does not require the REX prefix at all. In fact, Keystone assembles the instruction into `0x8a 0x2a`, where:

* opcode = 0x8a  // MOV r8, r/m8
* modrm.mod = 00b
* modrm.reg = 101b  // r8 = CH (cf. Intel SDM Vol. 2A, Table 2-2)
* modrm.rm = 010b  // r/m8 = \[EDX\] (cf. Intel SDM Vol. 2A, Table 2-2)

Since this is a 64-bit instruction, the default address size is 64 bits, thus \[EDX\] becomes \[RDX\]. As soon as I replace `[rdx]` with `[r8]` (which requires the use of the REX prefix), the test fails:

`LLVM ERROR: Cannot encode high byte register in REX-prefixed instruction`

AlexAltea:
@raphaelning Good to know! Thanks for investigating this in detail! :-)

Issue #186, opened 2019-02-22
'repne stosw word es:[edi], ax' raises emulator error


krytarowski:
**Describe the Bug**

Booting OpenBSD/i386 6.4 doesn't work.

Summary: 

$ rasm2  -d '0xf2 0x66 0xab'                
repne stosw word es:[edi], ax

**Host Environment**

NetBSD chieftec 8.99.34 NetBSD 8.99.34 (GENERIC) #1: Sun Feb 17 11:51:34 CET 2019 root@chieftec:/public/netbsd-root/sys/arch/amd64/compile/GENERIC amd64

* HAXM version: 

5df99973178384f8f812d9f7c8644341f43559fc + cherry-picked patch https://github.com/intel/haxm/issues/172#issuecomment-465471085 + https://github.com/intel/haxm/pull/185

* Host OS version: 
NetBSD 8.99.34 amd64
* Host OS architecture: x86\_64
x86_64
* Host CPU model: 
Intel(R) Core(TM) i7-3770 CPU @ 3.40GHz
* Host RAM size: 
32GB RAM
* (Optional) Host computer model: 
desktop

**Guest Environment**

* Android Emulator or QEMU version: 
qemu-3.0.0nb5 from pkgsrc

* Guest OS version: 
6.4
* Guest OS architecture: 
i386

**To Reproduce**

qemu-system-x86_64 -cdrom cd64.iso  -accel hax -curses

**Expected Behavior**
Booting OpenBSD 6.4 to a shell prompt

Expectation: 

Add support for missing instruction in the simulator.

**Reproducibility**

Always.

**Diagnostic Information**

```
[ 289625.798177] haxm_panic: em_decode_insn() failed: vcpu_id=0, len=3, CS:IP=0x0:0xd039d6d3, instr[0..5]=0xf2 0x66 0xab 0x5f 0x5d 0xc3
[ 289625.798177] haxm_warning: 4000 VMX_PIN_CONTROLS: 1f
[ 289625.798177] haxm_warning: 4002 VMX_PRIMARY_PROCESSOR_CONTROLS: 969861fe
[ 289625.798177] haxm_warning: 401e VMX_SECONDARY_PROCESSOR_CONTROLS: aa
[ 289625.798177] haxm_warning: 4004 VMX_EXCEPTION_BITMAP: 40000
[ 289625.798177] haxm_warning: 4006 VMX_PAGE_FAULT_ERROR_CODE_MASK: 0
[...]

PR #187, opened 2019-02-22
Allow REPNE prefixes in REP-only instructions

AlexAltea:
It seems that REP-only instructions also accept REPNE prefixes as a synonym of REP, so we need to remove the corresponding check in `em_decode_insn`.

Unfortunately, I can't justify this decision with the Intel SDM because this seems an undocumented feature, but such instructions do appear in OpenBSD guests (see #186), and I've also managed to replicate this behavior in my machine.

### DIFF
--- a/core/emulate.c
+++ b/core/emulate.c
@@ -1097,11 +1097,13 @@ em_status_t EMCALL em_decode_insn(struct em_context_t *ctxt, const uint8_t *insn
             // TODO: Should throw #UD
             return EM_ERROR;
         }
-        if (ctxt->rep == PREFIX_REPNE && !(opcode->flags & INSN_REPX)) {
-            /* Instruction supports REP (== REPE) but not REPNE */
-            // TODO: Should throw #UD
-            return EM_ERROR;
-        }
+        // NOTE: Specifications at Intel SDM Vol. 2B: 2.1.1 Instruction Prefixes
+        // mention that REP prefixes are encoded via 0xF3 (PREFIX_REPE), however
+        // they remain ambiguous as to whether 0xF2 (PREFIX_REPNE) is also valid.
+        // Hardware tests reveal both values act as REP in REP-only instructions,
+        // consequently we mimic this undocumented hardware behavior by skipping
+        // the following check:
+        // assert((opcode->flags & INSN_REPX) || (ctxt->rep != PREFIX_REPNE));
     }
 
     return decode_operands(ctxt);


AlexAltea:
@krytarowski Let us know about the status of OpenBSD guests after merging this patch! :-)

krytarowski:
It seems to go further.. but it still crashes.

http://netbsd.org/~kamil/haxm/openbsd-i386-trap-page-fault-2019-02-22-haxm.png

I'm expecting to get feedback from @polprog for the IA32_EFER patch and whether it boots Solaris.. if so we can marge it and sequentially these emulator fixups. Once done we can continue in a new issue ticket the porting process of OpenBSD/i386.

raphaelning:
Sorry for the delay. Usually we don't want to implement hacks that violate the spec just to work around some weird guest behavior, unless it is indeed an undocumented hardware feature--like @krytarowski said, the hypervisor should emulate hardware behavior faithfully.

> I've tested also managed to replicate this behavior in my machine.

Could you elaborate on that? I've only verified that NASM would compile `repne stosw` and retain the `0xf2` (REPNE) prefix, but the ultimate test would be to use that instruction in a test program and run it on bare metal (Intel CPU). Have you done that?

core/emulate.c
```diff
@@ -1097,11 +1097,6 @@ em_status_t EMCALL em_decode_insn(struct em_context_t *ctxt, const uint8_t *insn
             // TODO: Should throw #UD
             return EM_ERROR;
         }
-        if (ctxt->rep == PREFIX_REPNE && !(opcode->flags & INSN_REPX)) {
```
raphaelning: I'm okay with removing this check, but I'd like to see it replaced by a comment that explains why it is not done. The best argument would be "to mimic undocumented hardware behavior"--again, we need some kind of proof of that.

AlexAltea:
> I've only verified that NASM would compile `repne stosw` and retain the `0xf2` (REPNE) prefix, but the ultimate test would be to use that instruction in a test program and run it on bare metal (Intel CPU). Have you done that?

@raphaelning Yes, that's precisely what I did. I successfully assembled such instructions from [x64dbg](https://x64dbg.com), by using both the [XEDParse](https://github.com/x64dbg/XEDParse) and [asmjit](https://github.com/asmjit/asmjit) assembler backends. Its disassembler [Zydis](https://zydis.re/) also displays such instructions correctly. Ironically, it never displays *rep*, instead it displays *repe*/*repne* according to the prefix byte, in all REP-only instructions).

More importantly, I've manually verified that all variants of `repne stos*` and `repe stos*` behave the same, i.e. the ZF is never modified, and its original value is completely ignored. These tests have been done on my *Intel Core i7-7700HQ* processor. I haven't tested other REP-only instructions aside from `stos*`.

AlexAltea:
@raphaelning I've just added the comment.

core/emulate.c
```diff
@@ -1097,11 +1097,13 @@ em_status_t EMCALL em_decode_insn(struct em_context_t *ctxt, const uint8_t *insn
...
-            /* Instruction supports REP (== REPE) but not REPNE */
-            // TODO: Should throw #UD
-            return EM_ERROR;
-        }
+        // NOTE: Specifications at Intel SDM Vol. 2B: 2.1.1 Instruction Prefixes
+        // mention that REP prefixes are encoded via 0xF3 (REFIX_REPE), however
```
raphaelning: nit: => PREFIX_REPE

core/emulate.c
```diff
@@ -1097,11 +1097,13 @@ em_status_t EMCALL em_decode_insn(struct em_context_t *ctxt, const uint8_t *insn
...
+        // they remain ambiguous as to whether 0xF2 (REFIX_REPNE) is also valid.
```
raphaelning: nit: => PREFIX_REPNE

core/emulate.c
```diff
@@ -1097,11 +1097,13 @@ em_status_t EMCALL em_decode_insn(struct em_context_t *ctxt, const uint8_t *insn
...
+        // Hardware tests reveal both values act as REP in REP-only instructions,
+        // consequently we mimic this undocumented hardware behavior by skipping
+        // the following check:
+        // assert((opcode->flags & INSN_REP) && (ctxt->rep != PREFIX_REPNE));
```
raphaelning: nit: Technically, this `assert` will fail for `INSN_REPX` instructions. Probably you mean:

`assert((opcode->flags & INSN_REPX) || (ctxt->rep != PREFIX_REPNE));`

AlexAltea:
@raphaelning All done!

PR #188, opened 2019-02-22
Allow read and write of IA32_EFER in the !PAE && PG mode

krytarowski:
This is used at least by OpenBSD/i386 and Solaris 10 64-bit.

### DIFF
--- a/core/vcpu.c
+++ b/core/vcpu.c
@@ -3284,11 +3284,7 @@ static int handle_msr_read(struct vcpu_t *vcpu, uint32_t msr, uint64_t *val)
             break;
         }
         case IA32_EFER: {
-            if (!(state->_cr4 & CR4_PAE) && (state->_cr0 & CR0_PG)) {
-                r = 1;
-            } else {
-                *val = state->_efer;
-            }
+            *val = state->_efer;
             break;
         }
         case IA32_STAR:
@@ -3548,11 +3544,9 @@ static int handle_msr_write(struct vcpu_t *vcpu, uint32_t msr, uint64_t val)
             hax_info("Guest writing to EFER[%u]: 0x%x -> 0x%llx, _cr0=0x%llx,"
                      " _cr4=0x%llx\n", vcpu->vcpu_id, state->_efer, val,
                      state->_cr0, state->_cr4);
-            if ((state->_cr0 & CR0_PG) && !(state->_cr4 & CR4_PAE)) {
-                state->_efer = 0;
-            } else {
-                state->_efer = val;
-            }
+
+            state->_efer = val;
+
             if (!(ia32_rdmsr(IA32_EFER) & IA32_EFER_LMA) &&
                 (state->_efer & IA32_EFER_LME)) {
                 hax_panic_vcpu(
--- a/core/vcpu.c
+++ b/core/vcpu.c
@@ -3544,7 +3544,31 @@ static int handle_msr_write(struct vcpu_t *vcpu, uint32_t msr, uint64_t val)
             hax_info("Guest writing to EFER[%u]: 0x%x -> 0x%llx, _cr0=0x%llx,"
                      " _cr4=0x%llx\n", vcpu->vcpu_id, state->_efer, val,
                      state->_cr0, state->_cr4);
-
+            
+            /* val - "new" EFER, state->_efer - "old" EFER.*/
+            if ((val &
+                 ~((uint64_t)(IA32_EFER_SCE | IA32_EFER_LME |
+                              IA32_EFER_LMA | IA32_EFER_XD)))) {
+                hax_error("Illegal value 0x%llx written to EFER. "
+                          "Reserved bits were set. EFER was 0x%llx\n",
+                          val, (uint64_t) state->_efer);
+                r = 1;
+                break;
+            }
+            if (((val & IA32_EFER_LMA) ^
+                 (state->_efer & IA32_EFER_LMA))) {
+                hax_warning("IA32_EFER.LMA changed. EFER: 0x%llx -> 0x%llx\n",
+                            (uint64_t) state->_efer,val);
+            }
+            if ((state->_cr0 & CR0_PG) &&
+                ((val & IA32_EFER_LME) ^
+                 (state->_efer & IA32_EFER_LME))) {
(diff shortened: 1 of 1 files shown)


polprog:
This allows Solaris 10 to boot the kernel up to a point - it panics due to unhandled cr8 reads and writes

core/vcpu.c
```diff
@@ -3548,11 +3544,9 @@ static int handle_msr_write(struct vcpu_t *vcpu, uint32_t msr, uint64_t val)
             hax_info("Guest writing to EFER[%u]: 0x%x -> 0x%llx, _cr0=0x%llx,"
                      " _cr4=0x%llx\n", vcpu->vcpu_id, state->_efer, val,
                      state->_cr0, state->_cr4);
-            if ((state->_cr0 & CR0_PG) && !(state->_cr4 & CR4_PAE)) {
-                state->_efer = 0;
-            } else {
-                state->_efer = val;
-            }
+
```
raphaelning: I agree with getting rid of the old check, but I'd also like to see the following proper checks implemented in order to prevent invalid writes to `IA32_EFER`:

1. All reserved bits (1-7, 9, 12-63) must remain 0. (Vol. 3A, Table 2-1: _IA32_EFER MSR Information_; Vol. 2C, Section 5.2: _WRMSR—Write to Model Specific Register_)
1. The read-only IA32_EFER_LMA bit must not be modified. (Vol. 3A, Table 2-1: _IA32_EFER MSR Information_)
1. The IA32_EFER_LME bit must not be modified while CR0.PG = 1. (Vol. 3A, Section 4.1.2: _Paging-Mode Enabling_)

polprog: What's the expected behavior when such a write is detected? VCPU panic?

raphaelning: Good question. It's actually a little more complicated than I initially thought:

* For violations to 1) and 3), we should inject a \#GP exception to the vCPU. Simply setting `r = 1;` here and `break;` will do the trick.
* For violations to 2), it turns out that we should silently ignore the `IA32_EFER_LMA` bit in `val` (cf. the quote from Intel SDM below). In fact, we could skip this check altogether, because the subsequent `vmwrite_efer()` call always overwrites the `IA32_EFER_LMA` bit in `state->_efer`.

Reference: Intel SDM Vol. 3A, 4.1.1 _Three Paging Modes_, footnote:

> 2. The LMA flag in the IA32_EFER MSR (bit 10) is a status bit that indicates whether the logical processor is in IA-32e mode (and thus using 4-level paging). The processor always sets IA32_EFER.LMA to CR0.PG & IA32_EFER.LME. Software cannot directly modify IA32_EFER.LMA; an execution of WRMSR to the IA32_EFER MSR ignores bit 10 of its source operand.

polprog: So far I have the following (dirty) diff. If this sort of an approach is fine, I'll clean up the code and push it to this branch. 
```
diff --git a/core/vcpu.c b/core/vcpu.c
index b8a835c..c739245 100644
--- a/core/vcpu.c
+++ b/core/vcpu.c
@@ -3545,18 +3545,43 @@ static int handle_msr_write(struct vcpu_t *vcpu, uint32_t msr, uint64_t val)
                      " _cr4=0x%llx\n", vcpu->vcpu_id, state->_efer, val,
                      state->_cr0, state->_cr4);
 
+
+
+            /* val - "new" EFER, state->_efer - "old" EFER. moved state->_efer = val; to after checks*/
+            if ((val &
+                 ~(IA32_EFER_SCE | IA32_EFER_LME |
+                   IA32_EFER_LMA | IA32_EFER_XD   ))){
+                hax_info("Illegal value 0x%llx written to EFER. "
+                         "Reserved bits were set. EFER was 0x%llx\n", val, state->_efer);
+                r = 1;
+                break;
+            } else if (((val & IA32_EFER_LMA) ^
+                        (state->_efer & IA32_EFER_LMA)) ) {
+                hax_info("IA32_EFER.LMA changed. EFER: 0x%llx -> 0x%llx\n", state->_efer,val);
+            } else if ( !(state->_cr0 & CR0_PG) &&
+                        ((val & IA32_EFER_LME) ^
+                         (state->_efer & IA32_EFER_LME))) {
+                hax_info("Illegal value 0x%llx written to EFER. "
+                         "Attempted to enable Long Mode with paging enabled. EFER was 0x%llx\n", val, state->_efer);
+                r = 1;
+                break;
+            }
             state->_efer = val;
 
             if (!(ia32_rdmsr(IA32_EFER) & IA32_EFER_LMA) &&
                 (state->_efer & IA32_EFER_LME)) {
                 hax_panic_vcpu(
-                        vcpu, "64-bit guest is not allowed on 32-bit host.\n");
+                               vcpu, "64-bit guest is not allowed on 32-bit host.\n");
             } else if ((state->_efer & IA32_EFER_LME) && vtlb_active(vcpu)) {
                 hax_panic_vcpu(vcpu, "64-bit guest is not allowed on core 2 "
                                "machine.\n");
             } else {

                 vmwrite_efer(vcpu);
             }
+
             break;
         }
         case IA32_STAR:
```

raphaelning: Yes, that's about right. But the devil is in the detail...

1. The first `if` condition may fail to detect any 1 in bits 32..63 of `val`. This is because `IA32_EFER_xxx` constants are `unsigned int`, not `uint64_t`. You may want to convert the result of the bitwise OR to `uint64_t` before the NOT operation.
1. nit: The `else` before the second `if` condition is unnecessary.
1. The third `if` condition checks CR0.PG = 0 rather than CR0.PG = 1. Also, the log message should say "enable **or disable** Long Mode".
1. The `else` before the third `if` condition is simply wrong, because if the guest indeed changes the LMA bit, we'll end up skipping the PG/LME check. IMHO, the 3 checks we want to add are independent of each other, so they shouldn't be connected with `else`.
1. nit: Use `hax_error()` for the first and third checks. Use `hax_warning()` for the second check.

In addition, I'd appreciate it if you could make an effort to adapt your code to the style of the existing code:
1. Wrap lines longer than 80 characters.
1. `){` => `) {`.
1. Avoid unnecessary blank lines (e.g. two consecutive ones).
1. Avoid unnecessary spaces inside `()`: `( expression )` => `(expression)`.

PR #190, opened 2019-02-23
Changed major numbers per tech-kern@ discussion

polprog:
Changed major numbers to match new src/sys/conf change by @krytarowski for NetBSD

http://mail-index.netbsd.org/source-changes/2019/02/23/msg103709.html


krytarowski:
Looks good! You need to update your patch to include signoff.

krytarowski:
Please specify in the commit description that this is specific to NetBSD.

PR #191, opened 2019-02-23
 Fixed accesses to AH/CH/DH/BH registers in the emulator

### LINKED ISSUE #189: Access the wrong 8-bit register when using MMIO
**Describe the Bug**

Summary: 
When accessing the 8-bit register with instructions related to MMIO, HAXM access the wrong register.
In the current implementation of HAXM, register numbers are allocated in the following order.
AL, CL, DL, BL, SP, BP, SI, DI
However, when using the 8-bit register, the following order is correct[1].
AL, CL, DL, BL, AH, CH, DH, BH

[1] IA-32 Intel(R) Architecture Software Developer’s Manual Volume 2: Instruction Set Reference
B.1.1. Reg Field (reg)

**Host Environment**

* HAXM version: 7.4.1

* Host OS version: Windows 7 Professional SP1 

* Host CPU model: Intel Core i7-2600K

* Host RAM size: 16 GB

**To Reproduce**

For example, please run following instruction.

mov ch, [MEM_MMIO]
8a 2e xx xx

Here, MEM_MMIO is the address of memory mapped IO.
In this case, the value will be set to BP (Low) in HAXM.

**Expected Behavior**

Expectation: 
The value should be set in CH.

**Reproducibility**

100% (Any case when using MMIO address)

**Additional context**

I don't know if this fix is appropriate, I got the expected results by modifying vcpu_read_gpr and vcpu_write_gpr in vcpu.c as follows:

```
static uint64_t vcpu_read_gpr(void *obj, uint32_t reg_index, uint32_t size)
{
	struct vcpu_t *vcpu = obj;
	uint64_t value = 0;
	if (reg_index >= 16) {
		hax_panic_vcpu(vcpu, "vcpu_read_gpr: Invalid register index\n");
		return 0;
	}
	if (size == 1) {
		memcpy(&value, (UINT8*)(&vcpu->state->_regs[reg_index & 0x3]) + (reg_index >> 2), size);
		return value;
	}
	memcpy(&value, &vcpu->state->_regs[reg_index], size);
	return value;
}
static void vcpu_write_gpr(void *obj, uint32_t reg_index, uint64_t value,
	uint32_t size)
{
	struct vcpu_t *vcpu = obj;
	if (reg_index >= 16) {
		hax_panic_vcpu(vcpu, "vcpu_write_gpr: Invalid register index\n");
		return;
	}
	if (size == 1) {
		memcpy((UINT8*)(&vcpu->state->_regs[reg_index & 0x3]) + (reg_index >> 2), &value, size);
		return;
	}
	memcpy(&vcpu->state->_regs[reg_index], &value, size);
}
```

AlexAltea:
**This patch depends on #185!**

Implemented by adding a `op->reg.shift` field that indicates the how many bytes does the full-width register need to be right-shifted in order to access the desired value.

This patch fixes #189.

### DIFF
--- a/core/emulate.c
+++ b/core/emulate.c
@@ -310,28 +310,45 @@ static const struct em_opcode_t opcode_table_0F3A[256] = {
     X16(N), X16(N), X16(N), X16(N),
 };
 
-static uint64_t gpr_read(struct em_context_t *ctxt,
-                         unsigned index, size_t size)
+static uint64_t gpr_read_shifted(struct em_context_t *ctxt,
+                                 unsigned index, size_t size, size_t shift)
 {
     uint64_t value = 0;
+    uint8_t *value_ptr;
 
     if (!(ctxt->gpr_cache_r & (1 << index))) {
         ctxt->gpr_cache[index] = ctxt->ops->read_gpr(ctxt->vcpu, index);
         ctxt->gpr_cache_r |= (1 << index);
     }
-    memcpy(&value, &ctxt->gpr_cache[index], size);
+    value_ptr = (uint8_t*)&ctxt->gpr_cache[index];
+    memcpy(&value, &value_ptr[shift], size);
     return value;
 }
 
-static void gpr_write(struct em_context_t *ctxt,
-                      unsigned index, size_t size, uint64_t value)
+static void gpr_write_shifted(struct em_context_t *ctxt,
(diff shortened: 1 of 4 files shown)


AlexAltea:
I've merged this patch into pull request #185. Sorry for the confusion.

PR #197, opened 2019-03-25
Support stand-alone memory mappings

raphaelning:
Android Emulator may dynamically create and destroy temporary
memory mappings at guest runtime for certain rendering tasks via
hax_user_backed_ram_map() and hax_user_backed_ram_unmap()
($AOSP/external/qemu/target/i386/hax-mem.c), e.g.:

 hax_user_backed_ram_map() <1>
 1.1) ADD_RAMBLOCK <1>:     HVA 0x14e070000..0x16e070000
 1.2) SET_RAM2 (map) <1>:   GPA 0x7dffff000..0x7fffff000 =>
                            HVA 0x14e070000..0x16e070000
 hax_user_backed_ram_unmap() <1>
 1.3) SET_RAM2 (unmap) <1>: GPA 0x7dffff000..0x7fffff000
 hax_user_backed_ram_map() <2>
 2.1) ADD_RAMBLOCK <2>:     HVA 0x14de70000..0x16de70000

The second ADD_RAMBLOCK call fails, because its HVA range overlaps
with that of the first ADD_RAMBLOCK call.

The problem is that the "map" step creates a RAM block, but the
"unmap" step doesn't destroy it. Instead of adding a DEL_RAMBLOCK
ioctl, simply exempt the caller from calling ADD_RAMBLOCK in the
first place:

 - Introduce a new hax_memslot flag for "stand-alone" mappings,
   along with a new capability flag for this API change.
 - Remove the ADD_RAMBLOCK call from hax_user_backed_ram_map().
   Instead, call SET_RAM2 with the new flag. (This will be done on
   the Android Emulator side.)
 - Internally, SET_RAM2 creates a stand-alone RAM block for each
   stand-alone mapping.
 - When the stand-alone mapping is unmapped, the reference count
   of the corresponding stand-alone RAM block will hit 0, which
   allows SET_RAM2 to destroy this temporary RAM block.

+ Replace HAX_RAM_INFO_xxx with HAX_MEMSLOT_xxx in code that is
  not directly in touch with user space.

### DIFF
--- a/core/hax.c
+++ b/core/hax.c
@@ -340,6 +340,7 @@ int hax_get_capability(void *buf, int bufLeng, int *outLength)
         cap->winfo |= HAX_CAP_64BIT_RAMBLOCK;
 #ifdef CONFIG_HAX_EPT2
         cap->winfo |= HAX_CAP_64BIT_SETRAM;
+        cap->winfo |= HAX_CAP_IMPLICIT_RAMBLOCK;
 #endif
         cap->winfo |= HAX_CAP_TUNNEL_PAGE;
         cap->winfo |= HAX_CAP_RAM_PROTECTION;
--- a/core/include/memory.h
+++ b/core/include/memory.h
@@ -55,6 +55,8 @@ typedef struct hax_ramblock {
     uint8_t *chunks_bitmap;
     // Reference count of this object
     int ref_count;
+    // Whether this RAM block is associated with a stand-alone mapping
+    bool is_standalone;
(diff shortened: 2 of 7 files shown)


acrnsi:

krytarowski:
Is this going to help to https://github.com/intel/haxm/issues/171 ?

raphaelning:
> Is this going to help to #171 ?

Sorry, no. This is to fix an issue specific to Android Emulator.

Issue #164, opened 2019-02-05
Permit MMIO exits to bypass the emulation.


cracyc:
It appears to be impossible for a program to handle a mmio exit in userspace rather than round tripping to kernel every time.  If the registers are reloaded after HAX_EXIT_FAST_MMIO they will likely then be corrupted in em_emulate_insn.  I also tried to use HAX_EPT_PERM_NONE but it seems to not support small memory regions.

raphaelning:
I guess what you want is something like the macOS Hypervisor.framework API, which allows user space to access most VMCS registers, so the QEMU HVF accelerator (which runs in user space) can fetch the MMIO instruction, emulate it, and then resume the guest from the next instruction.

The HAXM API provides a higher-level abstraction by design, which is why `HAX_EXIT_FAST_MMIO` isn't suitable for your purpose (it assumes most of the instruction emulation work is done in the kernel).

`HAX_EXIT_PAGEFAULT` (assuming the MMIO region can be protected with `HAX_RAM_PERM_NONE`) might be a better alternative. However, right now the GPA protection mechanism is limited to non-MMIO regions, and even if it worked on MMIO regions, you would still need to make additional "round trips" to the kernel:

1. One for reading the vCPU register state to get the current instruction pointer (RIP), before you can fetch the MMIO instruction.
1. One for writing the vCPU register state to make RIP point to the next instruction, after you emulate the MMIO instruction.

With the HAXM API, you can't read/write the guest RIP alone, but have to sync a larger set of vCPU registers (again that's by design), so these round trips will be very costly.

In fact, if the instruction only accesses one MMIO address (which is the most common case), I think the `HAX_EXIT_FAST_MMIO` approach already guarantees the minimum number of kernel/user context switches (kernel => user => kernel), with pretty low overhead.

cracyc:
To be clear I'm trying to emulate VGA which requires an MMIO region 64K in size.  HAX_EXIT_FAST_MMIO requires a round trip for every vram read and write and there'll be thousands while a frame is drawn then none for a long period.  Try a vga program that uses unchained mode such as doom in qemu with haxm and you'll see the performance is horrid, far slower than emulation so I was trying to do it like dosemu which exit kvm at the first vram access and emulates until it appears the program is finshed.  As you say HAX_EXIT_PAGEFAULT would likely be the best solution but it appears that it has a 2MB granularity which is far to large for the 64k vga vram window and will put the entire vm address space within 2MB when running in real mode.

[...]

PR #199, opened 2019-04-18
Exception interface for supervisor

leecher1337:
When implementing HAXM into NTVDMx64, I experienced the need to tell HAXM some exit NMI vectors that I needed to handle (for instance, handle an invalid exception to be able to handle BOPs by client code).
Therefore I would consider it useful to have such an interface.


krytarowski:
Sorry, but a lot of whitespace issues (tabs vs spaces).

core/memslot.c
```diff
@@ -364,8 +364,12 @@ static inline void memslot_delete(hax_memslot *dest)
 
 static inline void memslot_move(hax_memslot *dest, hax_memslot *src)
 {
+    hax_list_node entry;
+
     ramblock_deref(dest->block);
+    entry = dest->entry;
     memslot_init(dest, src);
+    dest->entry = entry;
```
wcwang: Many thanks for @leecher1337 shooting the root cause! For performance reasons, it is recommended to get rid of memslot_init() invocation to avoid assigning dest->entry 3 times within this function. I drafted the commit message for your reference. The following code should be equivalent for your reference. Thanks again.

Fix a bug of memslot_move()

Invoking memslot_init() in memslot_move() will result in losing the
linked list information of the original memory slot node. Reimplement
memslot_move() to resolve some exceptional issues during mapping memory
slots.

static inline void memslot_move(hax_memslot *dest, hax_memslot *src)
{
    ramblock_deref(dest->block);
    src->entry = dest->entry;
    *dest = *src;
    ramblock_ref(dest->block);
}

core/memslot.c
```diff
@@ -362,14 +362,12 @@ static inline void memslot_delete(hax_memslot *dest)
...
-{
-    hax_list_node entry;
-
-    ramblock_deref(dest->block);
-    entry = dest->entry;
-    memslot_init(dest, src);
-    dest->entry = entry;
+static inline void memslot_move(hax_memslot *dest, hax_memslot *src)
+{
+	ramblock_deref(dest->block);
```
wcwang: According to the current coding style, it is recommended to use 4 spaces as indent.

core/cpu.c
```diff
@@ -222,6 +222,7 @@ bool vcpu_is_panic(struct vcpu_t *vcpu)
     if (vcpu->panicked) {
         hax_error("vcpu has panicked, id:%d\n", vcpu->vcpu_id);
         hax_panic_log(vcpu);
+		htun->_exit_reason = vmx(vcpu, exit_reason).basic_reason;
```
VelocityRa: use spaces
```suggestion
        htun->_exit_reason = vmx(vcpu, exit_reason).basic_reason;
```

leecher1337: Sorry, hopefully this was the last tab vs. space issue, I did a grep on \t now and didn't find any more areas

wcwang:
@leecher1337, we are preparing a new release and this pull request #199 needs more time to review from my colleagues. Do you agree to submit a separate pull request for memslot.c to fix the memslot_move() bug? We would like to merge this patch first in this release. The below title and content of commit message are for your reference. And remember to add 'Signed-off-by' signature in commit message. Thanks.

---
memslot: Fix a bug of memslot_move()

Invoking memslot_init() in memslot_move() will result in losing the
linked list information of the original memory slot node. Reimplement
memslot_move() to resolve some exceptional issues during mapping memory
slots.

coxuintel:
@leecher1337 I just found your amazing project https://github.com/leecher1337/ntvdmx64 . Can I assume that you are adding VT support using HAXM? Besides the README in the project, is there any additional step I should take to test you HAXM implementation? I'd like to test NTVDMx64 and check the HAXM issue you mentioned.

core/vcpu.c
```diff
@@ -4187,6 +4204,13 @@ void vcpu_debug(struct vcpu_t *vcpu, struct hax_debug_t *debug)
     vcpu_update_exception_bitmap(vcpu);
 };
 
+void vcpu_setexcbmp(struct vcpu_t *vcpu, uint32_t excbmp)
+{
+    vcpu->user_excbmp = excbmp;
+	hax_log(HAX_LOGE, "set user_excbmp = %08X", vcpu->user_excbmp);
```
VelocityRa: tab -> spaces

leecher1337:
@coxuintel 
You are right,it's exactly the project I'm using HAXM for.
I hopefully will find time to release a patch-set in the next days,so that HAXM support can be tried out, so watch for commits.
Be aware that there are currently some issues preventing satisfactory operation:
* Currently, only realmode works with HAXM, DPMI will crash (investigating this will be tough, I need to find a way to hook interrupts for instance,which is required for DPMI, some musings about it can be found in the HAXM support sourcecode files already)
* #164 is still a show-stopper. While I get decent speed-ups in one-page Textmode applications and DOS applications using streaming I/O compared to the CCPU, everything involving a lot of video output will suffer from a huge performance degradation. Original NTVDM was able to use the VGA card directly to prevent performance problems with applications doing graphics with the XPDM Display driver model. However as of Windows 8, XPDM was discontinued and replaced wirth WDDM which doesn't support VGA at all (Original Windows 32bit NTVDM suffers from the same problem since Win8). So NTVDMx64 can only do graphics by using the console screen buffer (yes, it still exists on Windows, it just has a bug that my loader fixes, see https://github.com/microsoft/terminal/issues/246 ). FASTMMIO is far too slow for it and the method I choose as a workaround doesn't support VGA planar reads. I think that this is an architectural problem that cannot be solved, as qemu etc. have the same issues.
* I currently didn't check in all the modifications that I made to HAXM so that this pull request won't get polluted by the other changes. I may need to add a new branch for that.

coxuintel:
For the rendering performance issue, XPDM or WDDM may not be the root cause from my point of view. In Vista or Win7, WDDM is the default graphics driver model. Unless you force install a XPDM driver onto the graphics adapter, the GPU works in WDDM model. So it may be due to the removed support of video BIOS INT call or something else, but not the XPDM or WDDM difference. I'm thinking, if it's possible to "translate" the real mode drawing to simply memory write, then use Direct2D to blt the content expected place so that you can leverage GPU HW acceleration.

leecher1337:
> For the rendering performance issue, XPDM or WDDM may not be the root cause from my point of view. In Vista or Win7, WDDM is the default graphics driver model. Unless you force install a XPDM driver onto the graphics adapter, the GPU works in WDDM model. So it may be due to the removed support of video BIOS INT call or something else, but not the XPDM or WDDM difference. I'm thinking, if it's possible to "translate" the real mode drawing to simply memory write, then use Direct2D or Direct3D to blit the content to expected place so that you can leverage GPU HW acceleration.
The removal of the ability to pass through the MMIO area to the real video adapter (possibly, because all rendering is now done via the GPU und therefore switch to VGA mode isn't supported) removed the drawing possibilities of NTVDM.
As you can clearly see, you cannot use graphics on normal 32bit NTVDM anymore. Try to do ALT+RETURN key combination and you will get an error message.
See https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/content/video/nf-video-videoportsettrappedemulatorports for more details. As long as there is no possibility to directly access the VGA card, I don'T see any change for usable video output in V86/VT-x.

> btw. Have you checked how DOS run on QEMU+KVM?
Yes, it's completely unusable in this regard as soon as QEMU uses HAXM. Without HAXM with the emulated CPU, it's OK. 
dosemu switches to an emulated CPU to circumvent the problem and then switches back to hardware after a certain numer of instructions got executed without any video access.

leecher1337:
@coxuintel  HAXM-build of NTVDMx64 is now available. Just check doc\haxm.txt for details.
Basically you have to use bld-haxm.cmd instead of bld.cmd and mkrelease-haxm.cmd instead of mkrelease.cmd to build it. Looking forward for your feendback.

leecher1337:
@coxuintel Compilation process has now been simplified, maybe this facilitates testing for you?

mirh:
Needs rebase.

leecher1337:
I'll leave that as an excercise to the maintainer, as it is a lot of work (we have conflicting files now) and I have no idea when maintainer will apply the pull request (maybe it takes another year?), so it's not worth the effort.

PR #214, opened 2019-06-20
vcpu: Update EFER state for set/get_regs

VelocityRa:
HAXM didn't return the correct EFER value (tested on the same guest code against WHPX).
Without this you can only get it by reading the MSR directly.

### DIFF
--- a/core/vcpu.c
+++ b/core/vcpu.c
@@ -3878,6 +3878,9 @@ int vcpu_get_regs(struct vcpu_t *vcpu, struct vcpu_state_t *ustate)
     ustate->_dr3 = state->_dr3;
     ustate->_dr6 = state->_dr6;
     ustate->_dr7 = state->_dr7;
+
+    ustate->_efer = state->_efer;
+
     _copy_desc(&state->_cs, &ustate->_cs);
     _copy_desc(&state->_ds, &ustate->_ds);
     _copy_desc(&state->_es, &ustate->_es);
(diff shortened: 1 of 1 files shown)


hyuan3:
Looks good to me. Thanks for the patch.

Reviewed-by: Hang Yuan <hang.yuan@intel.com>

wayne-ma:
OK to verify

coxuintel:
> HAXM didn't return the correct EFER value (tested on the same guest code against WHPX).
> Without this you can only get it by reading the MSR directly.

Is there any specific failure if missing this?

VelocityRa:
@coxuintel The guest I tested this on is really barebones (the entirety of the code is pretty close to this: https://wiki.osdev.org/Talk:Entering_Long_Mode_Directly) so I'm not sure, but it seems like there's no failure guest-side without this, based on some testing.

nevilad:
@coxuintel if missing this snapshot loading may not update EFER. But it doesn't work anyway (see https://github.com/intel/haxm/pull/248).

VelocityRa:
Bump?

wcwang:
Sorry for late response. Do you remember that your test cases can be passed on WHPX but not on HAXM when invoking HAX_VCPU_SET_REGS? Furthermore, in your patch, we noticed that you updated the entire EFER bits on this occasion. But maybe some bits are not allowed to set according to HAXM current implementation. Meanwhile, we will consider to prepare full test on your pull request and then decide further solution. Thanks for your tracking.

PR #204, opened 2019-05-25
Windows10 guest support

nevilad:
Added more supported features to cpuid and PAT MSR support. This enables to install and run a windows10 x86 guest. Windows10 x64 guests are not yet working, since the amount of CR8 reads\writes is very big and the loading process stucks on handling them.
The patch was tested on a windows 7 and windows 10 guests on a windows10x64 host.
There are open questions:
1) cr_pat is not read\written to qemu. Thus PAT MSR is not saved\restored from snapshots. I don't know how to add this support, since hax does not provide it's version to qemu, so qemu would not be able to take different actions for different hax versions - read cr_pat from the new version and don't read from the old. The same for CR8 register.
2) in vcpu.c on line 2020 both the local and the vcpu object versions of entry_ctls are modified. This makes it possible that the check at line 2043 will be false and entry controls are not saved to vmcs. Is this OK?
3) In cpuid filtering there are such checks:
 if (cpu_has_feature(X86_FEATURE_AESNI)) {
        cpuid_1_features_ecx |= FEATURE(AESNI);
    }
Is this really necessary, given that the function returns the intersection of cpuid_1_features_ecx and the value returned by host cpuid:
state->_ecx = (cpuid_1_features_ecx & state->_ecx) | FEATURE(HYPERVISOR);
?

PR #216, opened 2019-06-21
emulate: save/restore host flags in fastop_dispatch

jarveson:
fastop_dispatch isn't saving/restoring host flags before overwriting with guest flags, which can cause some interesting issues. This fixes that by just pushing and poping host flags before/after executing fastop instruction.

### DIFF
--- a/core/emulate_ops.asm
+++ b/core/emulate_ops.asm
@@ -197,12 +197,13 @@ global fastop_dispatch
     ;   src1     - stack2  (saved to non-volatile esi)
     ;   src2     - stack3  (saved to non-volatile edi)
     ;   flags    - stack4  (saved to non-volatile ebp)
-%define stack_arg(index) [esp + 4*04h + 04h + (index*04h)]
+%define stack_arg(index) [esp + 5*04h + 04h + (index*04h)]
 fastop_dispatch:
     push ebx
     push esi
     push edi
     push ebp
+    pushf
     mov ebx, stack_arg(1)
     mov esi, stack_arg(2)
     mov edi, stack_arg(3)
@@ -216,6 +217,7 @@ fastop_dispatch:
     pushf
     pop dword [ebp]
     mov [ebx], reg_dst
+    popf
     pop ebp
     pop edi
     pop esi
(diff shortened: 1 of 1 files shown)


wayne-ma:
OK to verify

coxuintel:
emmm I'm investigating another interesting issue in [PR#204](https://github.com/intel/haxm/pull/204) recently and finally came to the same root cause as described in this PR, that RFLAGS are corrupted after fastop.
I have some comments to the change, and I'll also add Alex to the reviewer.

core/emulate_ops.asm
```diff
@@ -210,11 +210,13 @@ fastop_dispatch:
     mov reg_dst, [ebx]
     mov reg_src1, [esi]
     mov reg_src2, [edi]
+    pushf
```
coxuintel: I'll suggest move PUSHF to beginning, after push ebp. And move POPF to end, before pop ebp. Then change stack_arg(index) by adding another 0x4 offset. No need to change call stack_arg() so that the fastop handler and remaining arguments are still start from index 0.

core/emulate_ops.asm
```diff
@@ -237,11 +239,13 @@ fastop_dispatch:
     mov reg_dst, [rsi]
     mov reg_src1, [r10]
     mov reg_src2, [r11]
+    pushf
```
coxuintel: Ditto. Move this PUSHF to beginning and adjust stack_arg() offset. Move POPF to end.

core/emulate_ops.asm
```diff
@@ -262,11 +266,13 @@ fastop_dispatch:
     mov reg_dst, [r11]
     mov reg_src1, [r8]
     mov reg_src2, [r9]
+    pushf
```
coxuintel: Ditto.

AlexAltea:
Looks good to me! Thanks for the patch @jarveson, I also agree with the new `pushf`/`popf` placement suggested by @coxuintel.

jarveson:
sure, so something like this then?

coxuintel:
@jarveson would you mind update the commit message in your own branch? I'll merge it.

jarveson:
like this? otherwise im not fully sure what you mean

coxuintel:
> like this? otherwise im not fully sure what you mean

Exactly. Unlike PR comments, detailed commit msg in original commit will remains in git repo which helps future development.

PR #217, opened 2019-06-24
Improve HaxmLoader.

coxuintel:
- Stop release service "IntelHaxm" before loading the test service.
  User doesn't have to stop "IntelHaxm" manually before use HaxmLoader.
- Prompt required administrative privileges if not run as administrators.
  If without administrative privileges, user only see access denied but
  isn't clear what should do.

### DIFF
--- a/HaxmLoader/haxm_loader.c
+++ b/HaxmLoader/haxm_loader.c
@@ -32,7 +32,13 @@
 #include <strsafe.h>
 #include <windows.h>
 
-#define SERVICE_NAME L"haxm service for test"
+#define WIDE2(x) L##x
+#define WIDE1(x) WIDE2(x)
+#define WFUNCTION WIDE1(__FUNCTION__)
+
+#define SERVICE_NAME_RELEASE L"IntelHaxm"
+#define SERVICE_NAME L"IntelHaxmTest"
+#define SERVICE_DISPLAY_NAME L"Intel HAXM Test Service"
 
 static void PrintErrorMessage(void)
 {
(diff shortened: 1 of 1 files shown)


HaxmLoader/haxm_loader.c
```diff
@@ -110,9 +156,8 @@ static int DriverInstall(const wchar_t *sysFilePath)
 
     pathLen = GetFullPathNameW(sysFilePath, MAX_PATH, driverLocation, NULL);
     if (!pathLen || pathLen >= MAX_PATH) {
-        printf("%s(): Failed to get absolute path:\r\n",
-                __FUNCTION__);
-        wprintf(L"  sysFilePath='%s'\r\n", sysFilePath);
+        wprintf(L"%s(): Failed to get absolute path:\r\n  sysFilePath=\"%s\"\r\n",
```
wcwang: Can wchar_t string be split into multiple lines as this line exceeds 80 columns?

HaxmLoader/haxm_loader.c
```diff
@@ -32,8 +32,32 @@
 #include <strsafe.h>
 #include <windows.h>
 
+#define WIDE2(x) L##x
+#define WIDE1(x) WIDE2(x)
+#define WFUNCTION WIDE1(__FUNCTION__)
+
+#define SERVICE_NAME_RELEASE L"IntelHaxm"
 #define SERVICE_NAME L"haxm service for test"
```
wcwang: Do you think it is appropriate to rename the SERVICE_NAME to "IntelHaxm_t" or "HaxmTest" as it is inconvenient for command line parameter using the long string with spaces?

coxuintel: emm, good point. I'll change that too.

HaxmLoader/haxm_loader.c
```diff
@@ -32,8 +32,32 @@
...
 
+BOOL IsTokenElevated() {
+    BOOL bElevated = FALSE;
+    HANDLE hToken = NULL;
+
+    if (OpenProcessToken(GetCurrentProcess(), TOKEN_QUERY, &hToken)) {
```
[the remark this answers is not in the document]
coxuintel: Personally I don't prefer early returns a lot. But I'll follow the style around.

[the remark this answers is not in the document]
coxuintel: Checking hToken after successful OpenProcessToken seems redundant, A process HANDLE can never be NULL if it's obtained successfully.

HaxmLoader/haxm_loader.c
```diff
@@ -66,6 +90,28 @@ static int DriverUninstall(void)
         printf("%s(): OpenSCManager failed.\r\n", __FUNCTION__);
         goto cleanup;
     }
+    // Stop installed service
+    hService = OpenServiceW(hSCManager, SERVICE_NAME_RELEASE, SERVICE_ALL_ACCESS);
+    if (hService) {
+        SERVICE_STATUS_PROCESS ssp = { 0 };
+        DWORD ssp_size = 0;
+
+        if (QueryServiceStatusEx(hService, SC_STATUS_PROCESS_INFO, (LPBYTE)(&ssp), sizeof(ssp), &ssp_size)) {
```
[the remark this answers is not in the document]
coxuintel: The parentheses around LPBYTE is necessary, otherwise LPBYTE(&ssp) is treated as a function and cause C2063 compiling error.

HaxmLoader/haxm_loader.c
coxuintel: Sure, will do that.

PR #225, opened 2019-08-07
Fix incorrect VMX_ENTRY_CONTROLS loading.

coxuintel:
There exists some issue when programming VMX_ENTRY_CONTROLS and GUEST_EFER:
- ENTRY_CONTROL_LONG_MODE_GUEST has no chance to clear.
- vmwrite_efer() called by vmwrite_cr() will check vmx field before
vmx field gets updated, and itself will also program VMX_ENTRY_CONTROLS,
which may cause incorrect VMX_ENTRY_CONTROLS programmed.
- vmx filed entry_ctls_base and entry_ctls are loaded from vmcs in:
vcpu_create()->vcpu_prepare()->fill_common_vmcs()->load_vmcs_common().
When vmwrite_cr() and vmwrite_efer() write a dirty VMX_ENTRY_CONTROLS,
it only updates vmx field entry_ctls, but it compares with value of
entry_ctls_base for dirty check, which may cause reduntant or miss
programming to VMX_ENTRY_CONTROLS.

Below changes are made:
- Add the missing clear of ENTRY_CONTROL_LONG_MODE_GUEST.
- Move vmwrite_efer() to last of vmwrite_cr() after VMX_ENTRY_CONTROLS
  programmed.
- Remove unnecessary entry_ctls_base in vmx field.

### DIFF
--- a/core/vcpu.c
+++ b/core/vcpu.c
@@ -1900,7 +1900,7 @@ static void vmwrite_cr(struct vcpu_t *vcpu)
     struct vcpu_state_t *state = vcpu->state;
     struct per_cpu_data *cpu = current_cpu_data();
 
-    uint32_t entry_ctls = vmx(vcpu, entry_ctls_base);
+    uint32_t entry_ctls = vmx(vcpu, entry_ctls);
     uint32_t pcpu_ctls = vmx(vcpu, pcpu_ctls_base);
     uint32_t scpu_ctls = vmx(vcpu, scpu_ctls_base);
     uint32_t exc_bitmap = vmx(vcpu, exc_bitmap_base);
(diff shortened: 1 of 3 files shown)


core/vcpu.c
```diff
@@ -2015,18 +2015,11 @@ static void vmwrite_cr(struct vcpu_t *vcpu)
     }
 
     if ((state->_cr4 & CR4_PAE) && (state->_cr0 & CR0_PG) &&
-        (state->_cr0 & CR0_PE)) {
+        (state->_cr0 & CR0_PE))
```
wcwang: According to the coding style, conditional or loop statements with complex conditions or statements may be more readable with curly braces. Some projects require that an if must always have an accompanying brace.

coxuintel: OK, I'll keep the curly braces.

PR #231, opened 2019-08-13
set PCLMULQDQ in feature informantion of CPUID

hyuan3:
CTS 'test_SSLSocket_getHandshankeSession_duringHandshake' relys on
PCLMULQDQ feature. Enable it in CPUID emulation if host can support.

The instructions of this feature may have MMIO operand. Need to add
their emulation if there is real case to use MMIO operand.

Signed-off-by: Hang Yuan <hang.yuan@intel.com>


wayne-ma:
The Jenkins build failed due to there is a compiling failure on Mac 10.12 which is brought by /haxm/platforms/darwin/hax_wrapper.cpp . Colin will fix it and then we will re-trigger this build .

wayne-ma:
Ok to verify

PR #232, opened 2019-08-14
Fix macOS build failure on SDK 10.12 and host crash.

coxuintel:
Macro panic() is expand to postfix with panic line number, when passing
a va_list to panic(), the macro can't expand correctly.
Since panic() is now wrapped with hax_panic(), the line number always
points inside hax_panic() which isn't necessary.
Use panic_plain instead.

Before hax logger refine, hax_log_level() incorrectly use printf() for
va_list. However it should use vprintf(). The new hax_log() inherits
from hax_log_level() but doens't fix printf()->vprintf().


wayne-ma:
Ok to verify

PR #233, opened 2019-08-14
Fix incorrect interruptibility_state before vmx entry.

coxuintel:
According to SDM 26.3.1.5 Checks on Guest Non-Register State, Bit 0
(blocking by STI) must be 0 if the IF flag (bit 9) is 0 in the RFLAGS
field.
There is an issue during snapshot loading, that IF and
interruptibility_state don't pass the checks, which will result in
VM-entry failure due to invalid guest state.
This WA correct the bit so that vmx entry check could pass. The normal
interruptibility_state update is done when advancing the IP.
In future, proper approach is expected to replace the WA.

### DIFF
--- a/core/intr_exc.c
+++ b/core/intr_exc.c
@@ -121,12 +121,15 @@ uint hax_intr_is_blocked(struct vcpu_t *vcpu)
 {
     struct vcpu_state_t *state = vcpu->state;
     uint32_t intr_status;
+    uint32_t intr_blocking = 0;
 
     if (!(state->_eflags & EFLAGS_IF))
         return 1;
 
+    intr_blocking |= GUEST_INTRSTAT_STI_BLOCKING;
+    intr_blocking |= GUEST_INTRSTAT_SS_BLOCKING;
     intr_status = vmx(vcpu, interruptibility_state).raw;
-    if (intr_status & 3)
+    if (intr_status & intr_blocking)
         return 1;
     return 0;
 }
--- a/core/vcpu.c
+++ b/core/vcpu.c
@@ -1078,7 +1078,27 @@ static void load_dirty_vmcs_fields(struct vcpu_t *vcpu)
         vcpu->rflags_dirty = 0;
     }
 
-    // interruptibility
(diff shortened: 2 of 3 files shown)


core/vcpu.c
```diff
@@ -1079,6 +1079,24 @@ static void load_dirty_vmcs_fields(struct vcpu_t *vcpu)
...
+    //   don't pass the checks as mentioned in SDM 26.3.1.5.
+    // In in-order execution, interruptibility_state is updated when advancing
+    //   the IP. However when a snapshot is loaded, EFLAGS are restored but
+    //   guest non-register state not restored.
+    //   TODO: Find better approach instead of letting the check pass.
+    if (!(state->_rflags & EFLAGS_IF)) {
+        if (vmx(vcpu, interruptibility_state).raw &
+            GUEST_INTRSTAT_STI_BLOCKING) {
+            vmx(vcpu, interruptibility_state).raw &=
+                ~GUEST_INTRSTAT_STI_BLOCKING;
```
wcwang: Use 8-space indent for long line break, while the 4-space indent is for normal line wrap.

core/intr_exc.c
```diff
@@ -121,12 +121,15 @@ uint hax_intr_is_blocked(struct vcpu_t *vcpu)
 {
     struct vcpu_state_t *state = vcpu->state;
     uint32_t intr_status;
+    uint32_t intr_blocking = 0;
 
     if (!(state->_eflags & EFLAGS_IF))
         return 1;
 
+    intr_blocking |= GUEST_INTRSTAT_STI_BLOCKING;
+    intr_blocking |= GUEST_INTRSTAT_SS_BLOCKING;
```
wcwang: intr_blocking = GUEST_INTRSTAT_STI_BLOCKING | GUEST_INTRSTAT_SS_BLOCKING;

coxuintel: I code in this way to prevent line exceeding 80 chars. And possibly more bits will be OR-ed.

PR #239, opened 2019-09-24
Add VM pause support

hyuan3:
Check if Qemu wants to pause VM. If yes, don't enter guest. This is
helpful in snapshot scenario that Qemu can tell HAXM to pause after
it finishes last guest IO/MMIO instruction.

Signed-off-by: Hang Yuan <hang.yuan@intel.com>

### DIFF
--- a/core/vcpu.c
+++ b/core/vcpu.c
@@ -1711,6 +1711,14 @@ int vcpu_execute(struct vcpu_t *vcpu)
             goto out;
         }
     }
+
+    // Check if Qemu pauses VM
+    if (htun->_exit_reason == HAX_EXIT_PAUSED) {
+        htun->_exit_status = HAX_EXIT_PAUSED;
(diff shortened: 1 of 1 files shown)


core/vcpu.c
```diff
@@ -1711,7 +1711,15 @@ int vcpu_execute(struct vcpu_t *vcpu)
...
-    err = cpu_vmx_execute(vcpu, htun);
+
+    // Check if Qemu pauses VM
+    if (htun->_exit_reason == HAX_EXIT_PAUSED) {
+        htun->_exit_status = HAX_EXIT_PAUSED;
+        hax_log(HAX_LOGD, "vcpu paused\n");
+        goto out;
+    }
+
+	err = cpu_vmx_execute(vcpu, htun);
```
krytarowski: whitespace issue.

PR #243, opened 2019-11-01
mac: Update macOS SDK version

wcwang:
Change default macOS SDK version to 10.12 according to current macOS
supported version.

Signed-off-by: Wenchao Wang <wenchao.wang@intel.com>


coxuintel:
We'll need update CI script.

PR #245, opened 2019-11-14
CPUID handling code refactoring

nevilad:
CPUID handling code refactoring, originally part of windows 10 support PR.

### DIFF
--- a/core/vcpu.c
+++ b/core/vcpu.c
@@ -2556,7 +2556,7 @@ static void handle_cpuid_virtual(struct vcpu_t *vcpu, uint32_t a, uint32_t c)
     uint32_t hw_model;
     uint8_t physical_address_size;
 
-    static uint32_t cpu_features_1 =
+    static uint32_t cpuid_1_features_edx =
             // pat is disabled!
             FEATURE(FPU)        |
             FEATURE(VME)        |
@@ -2581,30 +2581,24 @@ static void handle_cpuid_virtual(struct vcpu_t *vcpu, uint32_t a, uint32_t c)
             FEATURE(PSE)        |
             FEATURE(HTT);
 
-    static uint32_t cpu_features_2 =
+    static uint32_t cpuid_1_features_ecx =
             FEATURE(SSE3)       |
             FEATURE(SSSE3)      |
             FEATURE(SSE41)      |
             FEATURE(SSE42)      |
             FEATURE(CMPXCHG16B) |
             FEATURE(MOVBE)      |
+            FEATURE(AESNI)      |
+            FEATURE(PCLMULQDQ)  |
             FEATURE(POPCNT);
 
-    uint32_t cpu_features_ext =
+    static uint32_t cpuid_8000_0001_features_edx =
             FEATURE(NX)         |
             FEATURE(SYSCALL)    |
+            FEATURE(RDTSCP)     |
             FEATURE(EM64T);
 
-    // Conditional features
-    if (cpu_has_feature(X86_FEATURE_AESNI)) {
-        cpu_features_2 |= FEATURE(AESNI);
-    }
-    if (cpu_has_feature(X86_FEATURE_PCLMULQDQ)) {
-        cpu_features_2 |= FEATURE(PCLMULQDQ);
-    }
-    if (cpu_has_feature(X86_FEATURE_RDTSCP)) {
-        cpu_features_ext |= FEATURE(RDTSCP);
-    }
+    static uint32_t cpuid_8000_0001_features_ecx = 0;
 
     switch (a) {
         case 0: {                       // Maximum Basic Information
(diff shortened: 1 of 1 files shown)


core/vcpu.c
```diff
@@ -2727,7 +2745,8 @@ static void handle_cpuid_virtual(struct vcpu_t *vcpu, uint32_t a, uint32_t c)
             state->_eax = state->_ebx = state->_ecx = 0;
```
coxuintel: As @PatrickvL mentioned, this line will cause succeeding state->_ecx always cleared.

core/vcpu.c
```diff
@@ -2581,30 +2581,27 @@ static void handle_cpuid_virtual(struct vcpu_t *vcpu, uint32_t a, uint32_t c)
...
+            FEATURE(AESNI)      |
+            FEATURE(PCLMULQDQ)  |
             FEATURE(POPCNT);
 
-    uint32_t cpu_features_ext =
+    static uint32_t cpuid_8000_0001_features_edx =
             FEATURE(NX)         |
             FEATURE(SYSCALL)    |
+            FEATURE(RDTSCP)     |
+            FEATURE(PDPE1GB)    |
```
coxuintel: Seems like current implementation doesn't add PDPE1GB support yet. Is there any specific case that we can verify the correctness of this one?

core/vcpu.c
```diff
@@ -2581,30 +2581,27 @@ static void handle_cpuid_virtual(struct vcpu_t *vcpu, uint32_t a, uint32_t c)
...
-    }
-    if (cpu_has_feature(X86_FEATURE_PCLMULQDQ)) {
-        cpu_features_2 |= FEATURE(PCLMULQDQ);
-    }
-    if (cpu_has_feature(X86_FEATURE_RDTSCP)) {
-        cpu_features_ext |= FEATURE(RDTSCP);
-    }
+    static uint32_t cpuid_8000_0001_features_ecx =
+            FEATURE(LAHF)       |
+            FEATURE(PREFETCHW);
```
coxuintel: Ditto for LAHF and PREFETCHW. We need verify them.

core/vcpu.c
```diff
@@ -2581,30 +2581,27 @@ static void handle_cpuid_virtual(struct vcpu_t *vcpu, uint32_t a, uint32_t c)
...
-    // Conditional features
-    if (cpu_has_feature(X86_FEATURE_AESNI)) {
-        cpu_features_2 |= FEATURE(AESNI);
-    }
-    if (cpu_has_feature(X86_FEATURE_PCLMULQDQ)) {
-        cpu_features_2 |= FEATURE(PCLMULQDQ);
-    }
-    if (cpu_has_feature(X86_FEATURE_RDTSCP)) {
-        cpu_features_ext |= FEATURE(RDTSCP);
-    }
```
coxuintel: The original code here is to verify host feature before report to guest. Although modern CPU support them, we may need some logic to verify them before report to guest. On one hand we can filter out host unsupported feature since haxm behaves CPU features as subset of the host CPU, on the other in future we can disable certain features reported to guest for specific purpose like customize CPU feature.

So add a function here to verify 80000001 ecx/edx before the switch case.

core/vcpu.c
```diff
@@ -2622,10 +2619,31 @@ static void handle_cpuid_virtual(struct vcpu_t *vcpu, uint32_t a, uint32_t c)
...
+            if (0xF != cpuid_eax.familyID)
+                hw_family = cpuid_eax.familyID;
+            else
+                hw_family = cpuid_eax.familyID + (cpuid_eax.extFamilyID << 4);
+
+            if (0x6 == cpuid_eax.familyID || 0xF == cpuid_eax.familyID)
+                hw_model = (cpuid_eax.extModelID << 4) + cpuid_eax.model;
+            else
+                hw_model = cpuid_eax.model;
+
```
coxuintel: The refactoring makes it much clean now and match the pseudo code of SDM better.

core/vcpu.c
nevilad: So far I understand, haxm code supports 4-level paging, so I added this feature bit. I have no tests for this feature.

core/vcpu.c
nevilad: These bits are set by the processor when it supports these instructions:
Bit 00: LAHF/SAHF available in 64-bit mode.*
Bit 08: PREFETCHW.
Both don't need virtualization support (not mentioned in intel virtualization docs). When the processor does not support them, they will be not in state->_ecx\edx and after and operation  these bits will be cleared.
I agree, all additions of feature bits are not refactoring, but adding new features, which I added to support win10.

core/vcpu.c
nevilad: handle_cpuid_virtual returns the intersection of supported by the real processor features (state->_eax etc) and supported by the hypervisor. 
cpu_has_feature calls are redundant, since not supported features are not in state->_eax etc.
Hax doesn't support cpu features setting from qemu, so these lists are hardcoded. When such support will be added, hax will get these feature lists from qemu.

core/vcpu.c
nevilad: Yes, ecx zeroing should be removed.

core/vcpu.c
coxuintel: OK then, please move the adding of PDPE1GB, LAHF and PREFETCHW to a separate patch, so that we can bi-sect in case we found any issues during the full test, which generally took long time.

core/vcpu.c
coxuintel: See above. I think we can keep cpuid_8000_0001_features_ecx as 0 for now, so that you can add them in next patch.

core/vcpu.c
coxuintel: Actually we have plan on the list to add the support of setting cpu feature from qemu, so I'm thinking adding such validation for a CPUID handling refactoring patch is suitable.

nevilad: The current list of supported hax cpuid features is nearby of nehalem processor, which is very old. Getting feature list from qemu has any meaning when you support the newest CPUs and allow users to run hax with older CPU features. To support newer processors a lot of features should be added to hax, not just returning feature bits for cpuid. Some of them can slow down it's execution by generating new vm-exits (example - xsave support). When I tried to support Win10, I tried to add more features and experienced this.
Even if you add them, why do you want to check them in cpuid handling code, and not in ioctl? I think when the user wants to run a specific feature set and the underlying CPU does not support some of them, it would be correct to check this in ioctl handling code and report an error.
Even if you want to check them in cpuid code, why should it be done with a specific function - handle_cpuid_virtual already has the supported by the CPU features in it's state (eax\ebx\ecx\edx) and a simple intersection with the supported (by example state->_ecx = cpuid_8000_0001_features_ecx & state->_ecx) does the same work as checking the features and adding them to the list (example - cpuid_8000_0001_features_ecx).

coxuintel: Maybe I made this complicated a little bit early.

Yes currently all CPUID handled by handle_cpuid_virtual() will first return from host asm_cpuid() then do the filtering, so check host has the feature is redundant.

The case I'm describing is that, if due to some reason QEMU wants to test some feature that hasn't been added to haxm, via ioctl, like recent PCLMULQDQ. Current logic will clear them. What I'm expecting is some function which can validate those features, it doesn't have to be in handle_cpuid_virtual(), could be near iotcl as you said. 
You can leave this to us, we'll make change together when we add the interface.

PR #248, opened 2019-12-09
Snapshot issues

nevilad:
PDPTE values are not restored after snapshot loading. I added PDPTE values reading from guest memory and write these to vmcs after all EFER writes and control register writes from qemu, since these can change memory mode to use PDPTES.
Open questions:
1) Is it necessary to set the check_pae_pdpt after memory load from qemu? It is theoretically possible not changing CRs and MSRs from qemu, but change memory at location pointed by CR3. Should we assume this scenario?

### DIFF
--- a/core/cpu.c
+++ b/core/cpu.c
@@ -367,6 +367,11 @@ int cpu_vmx_execute(struct vcpu_t *vcpu, struct hax_tunnel *htun)
             hax_panic_log(vcpu);
             return 0;
         }
+        if( vcpu->check_pae_pdpt ) {
+          vcpu->check_pae_pdpt = 0;
+          if( !vcpu_check_pae_pdpte(vcpu) )
+            return 0;
+        }
         vcpu_handle_vmcs_pending(vcpu);
         vcpu_inject_intr(vcpu, htun);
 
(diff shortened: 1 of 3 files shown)


nevilad:
Windows 7 x86 guest needs EFER MSR, it sets Execute Disable Bit Enable: IA32_EFER.NXE. Current qemu version does not read\write it from haxm. When this MSR will not be saved\restored the guest OS will not work correctly. So I created this patch for qemu.
[0001-Windows-7-x86-guest-needs-EFER-MSR.patch.txt](https://github.com/intel/haxm/files/3939255/0001-Windows-7-x86-guest-needs-EFER-MSR.patch.txt)

There is a bug in qemu at least on Windows host which does not save the correct registers (https://bugs.launchpad.net/qemu/+bug/1855617). Do they work on other hosts, has somebody used snapshots with haxm on other hosts?

nevilad:
These MSRs are not saved to/from qemu, but they values may be saved during MSR write exits and returned in MSR read exits:
IA32_MTRRCAP
MTRRFIX64K_00000
IA32_MTRR_DEF_TYPE
IA32_CPUID_FEATURE_MASK
IA32_PMC0
IA32_PMC1
IA32_PMC2
IA32_PMC3
IA32_PERFEVTSEL0
IA32_PERFEVTSEL1
IA32_PERFEVTSEL2
IA32_PERFEVTSEL3

PR #250, opened 2019-12-10
Move selector validity checking to vcpu_set_regs

nevilad:
Comments to cs.ar and tr.ar validity checking stating that these can be wrong set by qemu. But the checks were done before every vmentry with reading the values from vmcs and that is bad for performance. I moved the checks to vcpu_set_regs since qemu sets them through this function. I did not found any other places that write these to vmcs.


nevilad:
Please don't apply this patch, I think I found an example where the TR.ar is corrupted by vmx. Need time to debug this.

PR #251, opened 2019-12-10
fpu and MSR dumping in vcpu_state_dump

nevilad:
Added fpu and MSR dumping to vcpu_state_dump since these are part ofthe guest state (these are switched from host to guest values before each vmentry and back after each vmexit).

### DIFF
--- a/core/vcpu.c
+++ b/core/vcpu.c
@@ -4244,7 +4244,12 @@ static void vcpu_dump(struct vcpu_t *vcpu, uint32_t mask, const char *caption)
 
 static void vcpu_state_dump(struct vcpu_t *vcpu)
 {
-    hax_log(HAX_LOGD,
+    int i;
+    struct gstate *gstate = &vcpu->gstate;
+    bool em64t_support = cpu_has_feature(X86_FEATURE_EM64T);
+    struct fx_layout *gfx = (struct fx_layout *)hax_page_va(gstate->gfxpage);
+
+    hax_log(HAX_LOGW,
             "RIP: %08llx  RSP: %08llx  RFLAGS: %08llx\n"
             "RAX: %08llx  RBX: %08llx  RCX: %08llx  RDX: %08llx\n"
             "RSI: %08llx  RDI: %08llx  RBP: %08llx\n"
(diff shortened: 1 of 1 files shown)


nevilad:
What is with this PR? Are there any questions?

core/vcpu.c
```diff
@@ -4311,6 +4316,66 @@ static void vcpu_state_dump(struct vcpu_t *vcpu)
             vcpu->state->_idt.base,
             vcpu->state->_idt.limit,
             (uint32_t)vcpu->state->_efer);
+
+    // Dump FPU state
+    hax_log(HAX_LOGW,
+            "FX layout:\n"
+            "fcw:    %08x  fsw:    %08x  ftw: %08x  res1: %08x  fop: %08x\n"
```
wcwang: 1. Uppercase each state as vcpu state format.
2. Pay attention to the space usage. Use one space after the colon and use 2 spaces between states.

core/vcpu.c
```diff
@@ -4311,6 +4316,66 @@ static void vcpu_state_dump(struct vcpu_t *vcpu)
...
+            "fpu_ip: %08llx  fpu_dp: %08llx\n"
+            "mxcsr:  %08x  mxcsr_mask: %08x\n",
```
wcwang: As both states are uint32_t, should they use 'lx' instead of 'x'?
MXCSR: %08lx  MXCSR_MASK: %08lx

core/vcpu.c
```diff
@@ -4311,6 +4316,66 @@ static void vcpu_state_dump(struct vcpu_t *vcpu)
...
+            gfx->fcw, gfx->fsw, gfx->ftw, gfx->res1, gfx->fop, gfx->fpu_ip,
+            gfx->fpu_dp, gfx->mxcsr, gfx->mxcsr_mask);
+
+    for (i = 0; i < 16; i++) {
+        hax_log(HAX_LOGW, "st_mm[%d] %02x%02x%02x%02x%02x%02x%02x%02x\n", i,
```
wcwang: Do you mind to add the colon after the array name and add a space every 4 bytes for separation?

core/vcpu.c
```diff
@@ -4311,6 +4316,66 @@ static void vcpu_state_dump(struct vcpu_t *vcpu)
...
+                gfx->st_mm[0][i], gfx->st_mm[1][i], gfx->st_mm[2][i],
```
wcwang: Could you explain why you output the array by column? According to the 2-dimensional array in memory layout, do you think it is more reasonable as below format?

    for (i = 0; i < 8; i++) {
        hax_log(HAX_LOGW, "st_mm[%d]: %02x%02x%02x%02x %02x%02x%02x%02x "
                "%02x%02x%02x%02x %02x%02x%02x%02x\n", i, gfx->st_mm[i][0],
                gfx->st_mm[i][1], gfx->st_mm[i][2], gfx->st_mm[i][3],
                gfx->st_mm[i][4], gfx->st_mm[i][5], gfx->st_mm[i][6],
                gfx->st_mm[i][7], gfx->st_mm[i][8], gfx->st_mm[i][9],
                gfx->st_mm[i][10], gfx->st_mm[i][11], gfx->st_mm[i][12],
                gfx->st_mm[i][13], gfx->st_mm[i][14], gfx->st_mm[i][15]);
    }

core/vcpu.c
```diff
@@ -4311,6 +4316,66 @@ static void vcpu_state_dump(struct vcpu_t *vcpu)
...
+
+    for (i = 0; i < 16; i++) {
+        hax_log(HAX_LOGW, "mmx_2[%d] %02x%02x%02x%02x%02x%02x%02x%02x\n", i,
+                gfx->mmx_2[0][i], gfx->mmx_2[1][i], gfx->mmx_2[2][i],
+                gfx->mmx_2[3][i], gfx->mmx_2[4][i], gfx->mmx_2[5][i],
+                gfx->mmx_2[6][i], gfx->mmx_2[7][i]);
+    }
+
+    for (i = 0; i < 96; i += 8) {
+        hax_log(HAX_LOGW,"pad %02x%02x%02x%02x%02x%02x%02x%02x\n",
```
wcwang: Add the colon and spaces for format.

    hax_log(HAX_LOGW, "pad: %02x%02x%02x%02x %02x%02x%02x%02x\n",

core/vcpu.c
```diff
@@ -4311,6 +4316,66 @@ static void vcpu_state_dump(struct vcpu_t *vcpu)
...
+    if (cpu_has_feature(X86_FEATURE_RDTSCP)) {
+        hax_log(HAX_LOGW, "MSR IA32_TSC_AUX:%08llx\n", gstate->tsc_aux);
+    }
+
+    if (!hax->apm_version)
+        return;
+
+    for (i = 0; i < (int)hax->apm_general_count; i++) {
+        uint32_t msr = (uint32_t)(IA32_PMC0 + i);
+        hax_log(HAX_LOGW, "MSR %08x:%08llx\n", msr, gstate->apm_pmc_msrs[i]);
```
wcwang: Should 0x08lx be for uint32_t?

core/vcpu.c
nevilad: 1. OK
2. I use 2 spaces between states. Field names are of different length, so I use different number  of spaces after the colon, so the output of different lines will be aligned by field names start and values, and that makes the output easier to understand. Are you sure to use one space? See how the output looks like, I think it is better.

core/vcpu.c
nevilad: I thought under Windows 'x' is formatting a 32 bit value. Is this not true, or different under other platforms?

core/vcpu.c
nevilad: Like this:
st_mm [%d] %02x%02x%02x%02x %02x%02x%02x%02x\n
?

core/vcpu.c
nevilad: I thought st_mm are registers of 8 byte size. Can you give me Intel docs link where these are described?

core/vcpu.c
nevilad: OK

core/vcpu.c
wcwang: Great, using column alignment is better. I agree with you.

core/vcpu.c
wcwang: You are right. I referred to the table from https://www.cs.uic.edu/~jbell/CourseNotes/C_Programming/DataTypesSummary.pdf and considered 'lx' should be more exact. It is up to you.

core/vcpu.c
wcwang: Correct, 4-byte separation is just for reading convenience.

core/vcpu.c
wcwang: According to the definition in SDM Vol. 2A 3.1.1.3, st_mm are 8-byte width. I am not sure why the definition of the array here is 8 by 16. But from the existing code, see vcpu_get_fpu() in core/vcpu.c, the nested loop is raws first and then columns. Anyway, I think the 16 bytes within each raw is memory consecutive. If you are sure your implementation is reasonable, I will also accept the approach as I cannot find any exact document link to confirm that. Of course, to make the format consistent with others is okay.

SDM definition:

ST(i) — The ith element from the top of the FPU register stack (i ← 0 through 7).
mm — An MMX register. The 64-bit MMX registers are: MM0 through MM7.

nevilad: Table 3-47 in sdm-vol-2abcd - these are saved by fxrstor as 10 bytes width, 6 bytes reserved. Rewritten the code for this and xmms.

core/vcpu.c
```diff
@@ -4311,6 +4316,80 @@ static void vcpu_state_dump(struct vcpu_t *vcpu)
...
+                gfx->mmx_2[i][3], gfx->mmx_2[i][4], gfx->mmx_2[i][5],
+                gfx->mmx_2[i][6], gfx->mmx_2[i][7], gfx->mmx_2[i][8],
+                gfx->mmx_2[i][9], gfx->mmx_2[i][10], gfx->mmx_2[i][11],
+                gfx->mmx_2[i][12], gfx->mmx_2[i][13], gfx->mmx_2[i][14],
+                gfx->mmx_2[i][15]);
+
+    }
+
+    for (i = 0; i < 96; i += 8) {
+        hax_log(HAX_LOGW,"pad: %02x%02x%02x%02x %02x%02x%02x%02x\n",
```
wcwang: Please add a space after the first comma.

nevilad: Done.

wayne-ma:
ok to verify

wayne-ma:
Ok to verify

Issue #254, opened 2019-12-21
Cannot Install Manually or through Sdk Manager


Phred7:
Whenever I attempt to install haxm I get an error. Ive looked through many forums and q&a sites saying how to fix problems but I still cannot get it to install.

Ive tried multiple times through android studio and always get the same error (attached)
![Haxm](https://user-images.githubusercontent.com/10716448/71314836-738cbf00-240d-11ea-8737-b18bfd62382a.JPG)

Alternatively, I have downloaded the zip file from git hub but it doesn't contain any sort of installer like the documentation suggests.

Furthermore, I cannot see Hyper-V in the programs and features section of the control panel, my computer has virtualization enabled and Im running windows 10 on a i7-8750H

Please let me know what I am doing wrong

wayne-ma:
@Phred7 ,could you try to download and install the release version from below link?
https://github.com/intel/haxm/releases/download/v7.5.4/haxm-windows_v7_5_4.zip

Phred7:
Thank you, I thought I downloaded the installer before but I'm not sure what I did before because that worked flawlessly!

wayne-ma:
Hi @Phred7 , so per my understanding ,you could install the HAXM v7.5.4 successfully now ,right? If so, I will close this issue here.

Issue #252, opened 2019-12-17
BSOD on Host Win10x64 Drv 7.5.4


FreddMadison:
Random and rear reproduce. Locally built 7.5.4 .pdb used. Again. Since there no shared release pdb

    HAXM version: Release driver DriverVer=7.5.4.0 x64
    Host OS version: Win10 x64
    Host OS architecture: x64
    Guest OS version: Win7 x86

BUGCHECK_P1: c0000005

BUGCHECK_P2: fffff8057d903597

BUGCHECK_P3: ffffa3051f105c60

BUGCHECK_P4: 0

EXCEPTION_CODE: (NTSTATUS) 0x12dcb470 - <Unable to get error code text>

FAULTING_IP: 
IntelHaxm!decode_op_modrm_rm+537 [c:\git\haxm_7.5.4\core\emulate.c @ 822]
fffff805`7d903597 48837f1800      cmp     qword ptr [rdi+18h],0

CONTEXT:  ffffa3051f105c60 -- (.cxr 0xffffa3051f105c60)
rax=0000000000000001 rbx=ffffb88dc6361da0 rcx=ffffb88dc6361da0
rdx=000000000000001d rsi=ffffb88dda2f8920 rdi=0000000000000000
rip=fffff8057d903597 rsp=ffffa3051f106650 rbp=0000000000000038
 r8=fcffffffffffffff  r9=ffffb88dda2f8000 r10=ffffb88dda2f8910
r11=000000000000001c r12=0000000000000000 r13=0000000000000000
r14=ffffb88dddf09260 r15=0000000000000001
iopl=0         nv up ei ng nz na po nc
cs=0010  ss=0018  ds=002b  es=002b  fs=0053  gs=002b             efl=00050286
IntelHaxm!decode_op_modrm_rm+0x537:
fffff805`7d903597 48837f1800      cmp     qword ptr [rdi+18h],0 ds:002b:00000000`00000018=????????????????
Resetting default scope

BUGCHECK_STR:  12DCB470

CPU_COUNT: 6

CPU_MHZ: af8

CPU_VENDOR:  GenuineIntel

CPU_FAMILY: 6

CPU_MODEL: 9e

CPU_STEPPING: a

CPU_MICROCODE: 0,0,0,0 (F,M,S,R)  SIG: 96'00000000 (cache) 0'00000000 (init)

BLACKBOXBSD: 1 (!blackboxbsd)

BLACKBOXNTFS: 1 (!blackboxntfs)

[...]

PR #258, opened 2019-12-30
Separate old EPT implementation with CONFIG_HAX_EPT2.

coxuintel:
Some fail cases in hax_create_vm() may call old EPT implementation,
which are not expected.
Separate old/new EPT routines with CONFIG_HAX_EPT2 so that only one
approarch is enabled at same time.

### DIFF
--- a/core/ept.c
+++ b/core/ept.c
@@ -240,6 +240,7 @@ static eptp_t ept_construct_eptp(hax_paddr_t addr)
     return eptp;
 }
 
+#ifndef CONFIG_HAX_EPT2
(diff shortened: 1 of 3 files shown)


krytarowski:
Please remove the old implementation entirely, it's not supported by at least NetBSD (Linux is likely broken too).

coxuintel:
> Please remove the old implementation entirely, it's not supported by at least NetBSD (Linux is likely broken too).

Yes, we plan to do so very soon. This PR is for tracking #252 and #254.

nevilad:
Vtlb is too not used anymore, would be good to remove it too.

PR #277, opened 2020-03-20
Enable setting CPUID feature for guest VCPUs

wcwang:
Add a new IOCTL HAX_VCPU_IOCTL_SET_CPUID to enable setting CPUID features for guest vCPUs. Instead of initializing guest CPUID once by the static features set, this feature will support dynamically setting guest CPUID features during the runtime of VM. The implementation refers to Intel SDM Vol. 2A 3.2 CPUID.

* Implements corresponding interfaces and data structures in CPUID module
* Enables Windows and macOS platforms support
* Adds the capability flag HAX_CAP_CPUID
* Adds API documentation for HAX_VCPU_IOCTL_SET_CPUID


platforms/windows/hax_entry.c
```diff
@@ -431,6 +431,18 @@ NTSTATUS HaxVcpuControl(PDEVICE_OBJECT DeviceObject,
             vcpu_debug(cvcpu, (struct hax_debug_t*)inBuf);
             break;
         }
+        case HAX_VCPU_IOCTL_SET_CPUID: {
+            struct hax_cpuid *cpuid = (struct hax_cpuid *)inBuf;
+            if (inBufLength < sizeof(struct hax_cpuid) +
+                cpuid->total * sizeof(struct hax_cpuid_entry)) {
```
nevilad: This will be a crash, if inBufLength is zero.

core/include/vcpu.h
```diff
@@ -234,6 +234,14 @@ struct vcpu_t {
     struct em_context_t emulate_ctxt;
     struct vcpu_post_mmio post_mmio;
     struct mmio_fetch_cache mmio_fetch;
+
+    // Guest CPUID feature set
+    // * The CPUID feature set is always same for each vCPU. A CPUID instruction
+    //   executed on any core will get the same result.
+    // * All vCPUs share the unique memory, which is actually allocated by the
+    //   first vCPU created by VM. If any vCPU sets features in this field, all
+    //   vCPUs will change accordingly.
```
nevilad: Why save this data in vCPU if all of them share same memory? Why not in vm?

core/vcpu.c
```diff
@@ -2536,50 +2548,16 @@ static void handle_cpuid_virtual(struct vcpu_t *vcpu, uint32_t a, uint32_t c)
...
-            FEATURE(EM64T);
-
-    static uint32_t cpuid_8000_0001_features_ecx = 0;
+    // To fully support CPUID instructions (opcode = 0F A2) by software, it is
+    // recommended to add opcode_table_0FA2[] in core/emulate.c to emulate
+    // (Refer to Intel SDM Vol. 2A 3.2 CPUID).
+    cpuid_get_guest_features(vcpu->guest_cpuid, &cpuid_1_features_ecx,
+                             &cpuid_1_features_edx,
+                             &cpuid_8000_0001_features_ecx,
+                             &cpuid_8000_0001_features_edx);
```
nevilad: SET_CPUID interface allows to set all output registers for any cpuid codes. But the only used data is ecx and edx for code 1 and 80000001. Why?

core/cpuid.c
```diff
@@ -114,3 +131,234 @@ bool cpuid_host_has_feature_uncached(uint32_t feature_key)
...
+    cpuid->feature_1_ecx &= cache.host_supported.feature_1_ecx &
+                            cache.hax_supported.feature_1_ecx;
+    cpuid->feature_1_edx &= cache.host_supported.feature_1_edx &
+                            cache.hax_supported.feature_1_edx;
+
+    // Set fixed supported features
+    cpuid_set_fixed_features(cpuid);
+
+    if (entry->ecx != cpuid->feature_1_ecx ||
+        entry->edx != cpuid->feature_1_edx) {
```
nevilad: There are 3 types of action with the actual supported features and the ones provided in ioctl:
1) Use intersection and ignore mismatch. This is currently used.
2) Use intersection and return error in case of mismatch. The user wanted the provided features and we can't support these.
3) Use provided features, no intersection with supported. It is users responsibility when the provided features are not supported by the cpu. This is the case of KVM, but it supports a lot of features, so the problems will be in case when the CPU does not support them.
Would be good add this as flags to the interface, at least in future.

core/cpuid.c
```diff
@@ -114,3 +131,234 @@ bool cpuid_host_has_feature_uncached(uint32_t feature_key)
...
+    entry = find_cpuid_entry(cpuid_info, kFunction, 0);
+    if (entry == NULL)
+        return;
+
+    hax_log(HAX_LOGI, "%s: function: %08lx, index: %lu, flags: %08lx\n",
+            __func__, entry->function, entry->index, entry->flags);
+    hax_log(HAX_LOGI, "%s: eax: %08lx, ebx: %08lx, ecx: %08lx, edx: %08lx\n",
+            __func__, entry->eax, entry->ebx, entry->ecx, entry->edx);
+
+    cpuid->feature_8000_0001_edx = entry->edx;
```
nevilad: ecx is not merged. CUrrently supported by haxm bits are zero but this will change in future.

core/cpuid.c
```diff
@@ -114,3 +131,234 @@ bool cpuid_host_has_feature_uncached(uint32_t feature_key)
...
+
+static void cpuid_set_fixed_features(cpuid_t *cpuid)
+{
+    const uint32_t kFixedFeatures =
+        FEATURE(MCE)  |
+        FEATURE(APIC) |
+        FEATURE(MTRR) |
+        FEATURE(PAT);
+
+    cpuid->feature_1_edx |= kFixedFeatures;
```
nevilad: Silently adding flags to the ones provided by the user without even log this. I think when the user want to clear these bits this should be a error.

nevilad:
Seems like you need a way to set only ecx/edx for cpuid codes 1 and 80000001. How Android Studio will use this feature?
Qemu will set all registers for all supported by the virtual cpu cpuid codes. This feature is not usable for this purpose.

core/cpuid.c
wcwang: Thanks for your suggestion. We will consider these cases in future.

core/cpuid.c
wcwang: You are right. This version is based on the current implementation and this feature will be easy to extend in future.

core/cpuid.c
wcwang: Actually these flags were referred from your commits. As for logs, it will always output the unchanged bits as warning in the caller functions, such as cpuid_set_0000_0001().

core/include/vcpu.h
wcwang: Yes, saving in VM has the same effect. But we considered that this IOCTL should be belonging to VCPU as it is CPUID.

core/vcpu.c
wcwang: Good question. Other function code can be added one by one in future. We did not open the setting channels for all codes as HAXM still has not supported for all features in current implementation.

platforms/windows/hax_entry.c
wcwang: Good catch. I will update the code immediately. Thanks for your review.

core/vcpu.c
nevilad: I think https://github.com/intel/haxm/pull/271 should be called set_cpuid and this feature set_cpuid_feature_flags. How Android Studio will use this feature? Who adds ioctl calls there?

wcwang:
> Seems like you need a way to set only ecx/edx for cpuid codes 1 and 80000001. How Android Studio will use this feature?

This PR is only the first version for CPUID feature, as HAXM is open source, you are welcome to suggest the interfaces or implementations if there is any requirement. Android Emulator will use this feature as below user space patch [987306](https://android-review.googlesource.com/c/platform/external/qemu/+/987306/).

> Who adds ioctl calls there?

We will continue to provide the user space code to add IOCTL calls for Android Emulator.

wayne-ma:
Ok to verify

nevilad:
The Android patch is called "Enable whitelisted CPU options by default if host supports them" and mentions the need for CPUID_EXT_AES, CPUID_EXT_PCLMULQDQ and CPUID_FXSR features. By default hax returns intersection of supported by CPU and hax features. This PR removes feature bits if they are not supported by CPU and hax. So using this PR it is only possible to remove already supported features. Is this OK?
So far I understand these bits are cpuid_1_features_edx FEATURE(FXSR) and cpuid_1_features_ecx FEATURE(AESNI)/FEATURE(PCLMULQDQ). So these are already supported, no need to set something from usermode.

platforms/windows/hax_entry.c
```diff
@@ -431,6 +431,18 @@ NTSTATUS HaxVcpuControl(PDEVICE_OBJECT DeviceObject,
             vcpu_debug(cvcpu, (struct hax_debug_t*)inBuf);
             break;
         }
+        case HAX_VCPU_IOCTL_SET_CPUID: {
+            struct hax_cpuid *cpuid = (struct hax_cpuid *)inBuf;
+            if (inBufLength == 0 || inBufLength < sizeof(struct hax_cpuid) +
+                cpuid->total * sizeof(struct hax_cpuid_entry)) {
```
nevilad: This is better:
if (inBufLength < sizeof(struct hax_cpuid) || inBufLength < sizeof(struct hax_cpuid) +
                cpuid->total * sizeof(struct hax_cpuid_entry)) {
total is first member with 4 bytes size. So any inBufLength less than 4 bytes will crash.

wcwang: Thanks for your point. Actually, below expressions should be more exact as sizeof(struct hax_cpuid) is 8. Anyway, I will adopt your suggestion.

`if (inBufLength < sizeof(cpuid->total) || ...)` or `if (inBufLength < sizeof(uint32_t) || ...)`

wcwang:
> So using this PR it is only possible to remove already supported features. Is this OK?

I totally agree with your opinion. In the earlier design of this PR, we would have set the features out of current supported scope. But we found that if user set any value, which feature has not been implemented by HAXM, and then it would result in regression. Just as the feature FEATURE(TSC_DEADLINE), I once asked you in my review comments for PR #271. We will attempt to extend the supported feature set in future after more verification. In future, probably we will separate the default feature set from the supported feature set. Then user can add or remove features by calling this IOCTL. Currently, it is OK to only remove supported features.

> So far I understand these bits are cpuid_1_features_edx FEATURE(FXSR) and cpuid_1_features_ecx FEATURE(AESNI)/FEATURE(PCLMULQDQ). So these are already supported, no need to set something from usermode.

Yes, you are right. Actually, this feature requirement came before the FEATURE(PCLMULQDQ) was added. We intended to enable such features by calling this interface. Due to aforementioned reasons, we screened the unsupported features out in this version. We will consider improving the current framework to meet more requirements. Any of your further input will be important for us and your contribution will be welcome. Thanks.

krytarowski:
NetBSD support is missing. I can test it.. but is there a qemu support?

nevilad:
This feature is made for Android Studio. Qemu support will not be added. I update my PR and then it can be tested.

krytarowski:
OK. I'm going to submit a patch soon.

krytarowski:
This PR also broke building on NetBSD.

krytarowski:
https://github.com/intel/haxm/pull/281

wcwang:
@krytarowski Thanks for your reply. It supports QEMU and I have shared the user space patches for QEMU [here](https://github.com/intel/haxm/files/4385647/qemu_v4.2.0-haxm_cpuid.zip). You input will be welcome.

core/cpuid.c
wcwang: We will update the implementation of CPUID and include a hash table for feature set, which will add ecx or other register settings. It will not change the HAX_VCPU_IOCTL_SET_CPUID interface. The patch will be ready once the current release completes.

nevilad: A hash table for searching entries by find_cpuid_entry? There are currently only two codes with such flags, a vector is sufficient for this task.

wcwang: A hash table for storing in hax_cpuid_t for future extensions although a vector is sufficient currently.

PR #281, opened 2020-03-25
Fix NetBSD build and add support for HAX_VCPU_IOCTL_SET_CPUID

krytarowski:


platforms/netbsd/hax_entry_vcpu.c
```diff
@@ -58,6 +58,40 @@ struct cdevsw hax_vcpu_cdevsw = {
     .d_flag = D_OTHER | D_MPSAFE
 };
 
+#define load_user_data(dest, src, body_len, body_max, arg_t, body_t)          \
```
nevilad: Why not create a function? My ioctl will use it too.

krytarowski: A function is better.. but I want to avoid divergence with other kernels.

core/cpuid.c
```diff
@@ -35,10 +35,10 @@
 #define CPUID_CACHE_SIZE 6
 
 typedef struct cpuid_cache_t {
-    uint32_t data[CPUID_CACHE_SIZE];  // Host cached features
-    cpuid_t  host_supported;          // Physical CPU supported features
-    cpuid_t  hax_supported;           // Hypervisor supported features
-    bool     initialized;
+    uint32_t     data[CPUID_CACHE_SIZE];  // Host cached features
+    hax_cpuid_t  host_supported;          // Physical CPU supported features
```
nevilad: Is cpuid_t  present in NetBSD headers and renaming is to resolve intersection issues?

platforms/netbsd/hax_entry_vcpu.c
nevilad: How does a function lead to divergence?

core/cpuid.c
krytarowski: yes

platforms/netbsd/hax_entry_vcpu.c
krytarowski: Linux + Darwin use macro here.

wcwang: > Why not create a function?

This pair of function-like macros are designed for all interfaces with variable-length arguments. As there are types, struct members in the macro parameters, it is hard to define a real function to handle these cases. But I agree to bring the definition to a common place, such as `hax.h`, instead of duplicating the codes here.

> Linux + Darwin use macro here.

As for the minor difference between NetBSD/Darwin and Linux, should we merge them into one macro by platform switch?

core/cpuid.c
krytarowski: There is no option to rename this. `_t` is a reserved namespace for POSIX/implementation and third party code must not reuse it for internal purposes.

krytarowski: There are also several other namespace conflicts in the past, such as `vaddr_t`. They were patched accordingly.

platforms/netbsd/hax_entry_vcpu.c
krytarowski: It's better to keep the macro in private files of kernels as there are OS-specific differences.

core/cpuid.c
wcwang: Got it. That makes sense. Could you help to update the patch fe1524c to ensure the code line length not exceed column 80? After replacing string, some lines exceed column 80 and need to format. Then how about reordering the patch sequence as fe1524c first and 9417d38 last? Thanks.

krytarowski: I will do it.

krytarowski:
Done.

wcwang:
Thanks for your update. Your pull request is ready to merge.

wcwang:
@krytarowski, HAXM has introduced the new API HAX_VCPU_IOCTL_GET_CPUID, would you implement it for NetBSD before the next release? Thanks.

PR #287, opened 2020-04-08
Add IOCTL to support haxm max vcpu value query

### LINKED ISSUE #195: Support more than 16 vCPUs
Currently, HAXM supports [max 16 vCPUs](https://github.com/intel/haxm/blob/7cb307c4e7ff066b51a503c76f72d7dfefc78760/core/include/config.h#L69) per VM. 

Considering recent multi-core CPUs (e.g., Xeon scalable processor), I think the current limit  is small. KVM supports [max 288 vCPUs](https://github.com/torvalds/linux/blob/a5ed1e96cafde5ba48638f486bfca0685dc6ddc9/arch/x86/include/asm/kvm_host.h#L41).

It seems on NetBSD, [`HAX_MAX_VMS * HAX_MAX_VCPUS` pseudo devices are created at the module initialization](https://github.com/intel/haxm/blob/7cb307c4e7ff066b51a503c76f72d7dfefc78760/platforms/netbsd/hax_entry.c#L299),  but on other platforms, a pseudo device is created when creating a vCPU (i.e., in the `hax_vcpu_create_host()`). So I think except for NetBSD, increasing  `HAX_MAX_VCPUS` has no side effects.

[QEMU uses the `HAX_MAX_VCPUS`](https://github.com/qemu/qemu/blob/d97a39d903fe33c45be83ac6943a2f82a3649a11/target/i386/hax-i386.h#L43). [This is my patch](https://github.com/mmisono/qemu/commit/048c131a769b3d3143045e0b7c45edc04e1c8245) for the QEMU to allocate the `hax_vcpu_state` as much as necessary (I tested this on the Linux host / Linux guest). I'm not sure other software using HAXM.
Is there anything to consider?

bowen218:
This commit tried to add ioctl HAX_IOCTL_CAP_MAX_VCPU that will return
the HAXM max vcpu support. Also, the switch case for hax device and vm
device add return -ENOSYS in the default case. Moreover, the max HAXM
supported vcpu value is updated from 16 to 64. Previously, HAXM doesn't
support IOCTL that would return the max vcpu value.

This issue is resolved by adding a new IOCTL HAX_IOCTL_CAP_MAX_VCPU that
will simply return HAX_MAX_VCPUS. (IOCTL naming credit to KVM)

This commit results in if QEMU calls this IOCTL, it will get the max
vcpu value HAXM driver supported and can then compare it with the QEMU
max value and smp value to determine whether the smp value is valid.

Signed-off-by: WangBowen <bowen.wang@intel.com>


krytarowski:
The NetBSD part looks good, however I would keep `HAX_MAX_VCPUS` at level 16 for this OS. It's not urgent to switch it to a higher number and I would like to write a proper support for emulated devfs (haxmvfs) that will create dynamically the `/dev` nodes.

bowen218:
> 
> 
> The NetBSD part looks good, however I would keep `HAX_MAX_VCPUS` at level 16 for this OS. It's not urgent to switch it to a higher number and I would like to write a proper support for emulated devfs (haxmvfs) that will create dynamically the `/dev` nodes.

Thanks for you review! I have updated the code, please take a look~

krytarowski:
Looks good to me, maybe we could update the comment `// TODO: Handle 64 VMs` to reflect the new reality.

bowen218:
> 
> 
> Looks good to me, maybe we could update the comment `// TODO: Handle 64 VMs` to reflect the new reality.

Do you mean to add a comment // TODO: Handle 64 VCPUS ?

krytarowski:
Yes. I would rephrase the existing comment to: `// TODO: Handle 64 VMs and 64 VCPUs.`

bowen218:
> 
> 
> Yes. I would rephrase the existing comment to: `// TODO: Handle 64 VMs and 64 VCPUs.`

Thanks for your suggestion! I have updated the code~

platforms/windows/hax_entry.c
```diff
@@ -714,6 +715,16 @@ NTSTATUS HaxDeviceControl(PDEVICE_OBJECT DeviceObject,
             infret = sizeof(uint32_t);
             ret = STATUS_SUCCESS;
             break;
+        case HAX_IOCTL_CAP_MAX_VCPU:
+            if (outBufLength < sizeof(uint32_t)) {
+                Irp->IoStatus.Information = 0;
```
nevilad, line 720: Irp->IoStatus.Information is changed after done:
done:
    Irp->IoStatus.Status = ret;
    Irp->IoStatus.Information = infret;
infret is zero after init, so this line (Irp->IoStatus.Information = 0;) is not needed.

bowen218: Thanks for your suggestion! I have updated the code~

coxuintel (approved):
Looks good to me.

nevilad:
How this feature will be used, is there any patch for qemu?

bowen218:
> How this feature will be used, is there any patch for qemu?

Basically, this feature will be used for QEMU to obtain the maximum vcpu HAXM supported when creating the vm to check whether the requested vcpu number is valid. Also, as [PR#255](https://github.com/intel/haxm/pull/255) supports more host cpus, we proceed to support more vcpus. (deals with [Issue#195](https://github.com/intel/haxm/issues/195).  And there will be a qemu patch (already in review process).

PR #288, opened 2020-04-09
MacOS support

nevilad:
Added MacOS support. Please don't merge this PR yet, since it should be modified according last changes it haxm (cpuid feature setting addition).
Tested it on windows host with image from https://github.com/intel/haxm/issues/149. It runs but ends in a damaged disk error window. Same with qemu+tcg, so this looks like installation iso issue, at least on my host.
MacOS can read/write different MSRs in different runs, when some MSR is not emulated, the guest crashes. To debug this, I've added logging of accesses to these MSRs. When you encounter crashes, look to the log.
MacOS writes data to UEFI vars flash. When these are written wrong (due to unexpected halt etc.), you may encounter display issues. The first sign of these is that the apple in not in the center of the monitor. In such a case restore UEFI vars flash to it's original file.
This qemu patch is needed to support UEFI in qemu for hax:
[0001-Add-support-for-ROMD-regions-for-hax.patch.txt](https://github.com/intel/haxm/files/4455872/0001-Add-support-for-ROMD-regions-for-hax.patch.txt)


core/vcpu.c
```diff
@@ -448,7 +448,7 @@ struct vcpu_t *vcpu_create(struct vm_t *vm, void *vm_host, int vcpu_id)
             sizeof(struct vcpu_state_t), HAX_MEM_NONPAGE);
     if (!vcpu->state)
         goto fail_4;
-    memset(vcpu->state, 0, sizeof(struct vcpu_state_t));
+    memset(vcpu->state, 0, sizeof(struct vcpu_state_t)); //cr8 = 0
```
krytarowski, line 451: This comment is unclear (without looking at the whole commit).

nevilad: I will remove it in the final version. This is the place where cr8 is initialized. I used it to analyze the cr8 support feature.

wcwang:
@nevilad, could you help to rebase the commit onto the latest release version so that it can be tested on macOS? Thanks.

nevilad:
This PR is not ready! I'm waiting for my set_cpuid feature, so the user can set Penryn cpu. Now it's hardcoded in this line:
state->_eax = 0x10673; //Penryn
Another thing is IA32_PERF_STATUS. I think it would be better to return as bus ratio part of the MSR the value from the underlying hardware, if it is new and contains this data in this MSR.

When you wan't check cr8 impact on booting empty multicore guests on MacOS, it is OK to test on this code, without latest changes from master.

PR #328, opened 2020-10-03
Fixed releases url in README

brahambence:
The previous url opened a view with only the single latest release, which might be "CheckTool" or "HAXM", depending on the current release schedule. 

The new url brings up a reverse-chronological view of the latest releases, so the same information is available at the same place, but scrolling down will give you more recent releases. On the previous url, you couldn't find downloads for HAXM right now, only CheckTool.

Signed-off-by: Abraham Bence <brahambence@gmail.com>


wcwang:
Thanks for your fixing. Could you update your patch with below aspects and I will merge it directly. Thanks a lot.

* Remove the trailing slash of the URL
* Keep the commit message line not exceeding column 72
* Uppercase 'url' to 'URL' in the commit message

wcwang:
Thanks for your commit. This issue has been resolved by PR #348.

PR #347, opened 2021-01-26
Fix some vulnerability issues of loading DRs and MSRs

wcwang:
Fix some vulnerability issues of loading DRs and MSRs during VM entries and exits.

- Reset DR7 before setting DR0 when loading guest DRs
- Automatically load guest and host MSRs by using MSR list
- Add more MSRs to pass-through and fix some MSR array initialization issues
- Check if instruction offset is valid before accessing

Signed-off-by: Wenchao Wang <wenchao.wang@intel.com>

### DIFF
--- a/core/emulate.c
+++ b/core/emulate.c
@@ -586,28 +586,48 @@ static void register_add(struct em_context_t *ctxt,
 
 static uint8_t insn_fetch_u8(struct em_context_t *ctxt)
 {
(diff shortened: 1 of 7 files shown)


wayne-ma:
Ok to verify

wayne-ma:
ok to verify

hyuan3 (approved):

PR #386, opened 2021-06-02
Remove legacy VTLB engine

wcwang:
As the new memory virtualization engine EPT2 (Extended Page Table) has been stable, the VTLB codes are deprecated.

Change summary:

* Removes the VTLB implementation and header files, and update the project files for 4 platforms;
* Removes the related variables in other modules and redundant memory allocation for VTLB structure;
* Adds the MMIO module to manage MMIO related functions.


core/include/mmio.h
```diff
@@ -0,0 +1,95 @@
+/*
```
hyuan3, line 1: If mmio.h and mmio.c are not related with VTLB code clean, better submit in another commit.

wcwang: Thanks for your careful review. These MMIO functions had already existed in VTLB module before. All functions will be invoked by vCPU module currently, so they cannot be omitted because of build reason. All changes this time are atomic. For more details, see the comments at line 727 in previous [vtlb.c](https://github.com/intel/haxm/blob/c06eb4ecc80549fbf9ddd5602260ef0e53dd8401/core/vtlb.c#L727).

hyuan3 (approved):

PR #438, opened 2022-01-27
netbsd: define min and max macros

ryoon:
Fix `modload haxm.mod` runtime error by missing min() reference.


wcwang:
Thanks for your commitment. The macro min() is only used in cpuid.c and memslot.c, both of which include "types.h", where the macro min() is defined. Could you check the cause and paste the NetBSD build errors for further analysis? Thanks.

ryoon:
Hi,

I am sorry. I had rebased my older patch to the latest tree without checking core/include/types.h change.
I withdraw this pull request.

Thank you very much.

